PR #6, opened 2020-05-07
add .gitignore

markusicu:
- copied from ICU
- ICU-specific file paths removed
- *.war added
- I assume that git ignores dot-files by default anyway so that I need not add .DS_Store

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,81 @@
+# Wildcards
+*.8
+*.a
+*.ao
+*.d
+*.dSYM
+*.db
+*.dll
+*.dylib
+*.exp
+*.lib
+*.lnk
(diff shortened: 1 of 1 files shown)


srl295:
> * I assume that git ignores dot-files by default anyway so that I need not add .DS_Store

that would be nice, but I don't think it does. Probably worth adding it.

markusicu:
> > * I assume that git ignores dot-files by default anyway so that I need not add .DS_Store
> 
> that would be nice, but I don't think it does. Probably worth adding it.

Oops, did the merge before you sent this... Let's see if Mac people see those files in their git status. (We don't have those in the ICU .gitignore.)

PR #12, opened 2020-07-22
Update U+1FA84 MAGIC WAND entry

seanstewart-unicode:
[L2/14-173R](https://www.unicode.org/L2/L2014/14173r-emoji-skin-tone.pdf) was associated with the "magic wand" entry and appears to be unintentional.
This document discusses ways to enhance the emoji standard with additional mechanisms for representation and inclusion of users.

[L2/18-218](https://www.unicode.org/L2/L2018/18218-magic-wand-emoji.pdf) makes no mention of skintone and neither document references each other.

I know it's old, but flagging anyway. Feel free to reject!


macchiati (approved):

nedley (changes requested):
This character has fallen off the list of proposals, and the link on https://unicode-org.github.io/emoji/emoji/charts-13.1/emoji-proposals.html appears to be correct.

macchiati:
So you are saying that we need to drop this PR?

nedley:
> So you are saying that we need to drop this PR?
> […](#)

Yes, unless there is some other place where this problem still occurs.

macchiati:
Sounds good. (I just noticed that it was still open and awaiting me, and
wanted to not be a roadblock (this time))

Mark

PR #9, opened 2020-05-08
Dockerize

srl295:
- rebased on current!
- add action to build docker image (does not deploy anywhere)
- [x] TODO: need to support `$PORT`

PR #13, opened 2020-07-23
update tomcat version to 9.0.37

srl295:
Update to #9


mihnita:
BTW, my approval might not mean much.
I think my Github account is worth nothing for the Unicode (and Google) groups :-)

srl295:
@macchiati @sffc markusicu mihai has okayed this, can i get a rubber stamp? You can see that the build passes now.

macchiati:
At this point, let's wait until a couple hours after the push to
production, just on the off chance that it destabilizes something. That
should be before noon tomorrow, since Tom is on ET.

Mark

srl295:
We can, but this isn’t CLDR :)

macchiati:
whoops!
Mark

macchiati:
Approved.

Mark

PR #17, opened 2020-09-04
CLDR-14145 JSP: automatically push to gcr.io on tag

srl295:
whenever a new release (tag) is created, push the JSPs to gcr.io so it can be deployed.
(CLDR bug number just for tracking purposes)

### DIFF
--- a/.github/workflows/push-jsp-on-tag.yml
+++ b/.github/workflows/push-jsp-on-tag.yml
@@ -7,6 +7,7 @@ jobs:
   build-and-push-to-gcr:
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v2
       - name: Get the version
         id: get_tag_name
         run: echo ::set-output name=GIT_TAG_NAME::${GITHUB_REF/refs\/tags\//}
--- a/.github/workflows/push-jsp-on-tag.yml
+++ b/.github/workflows/push-jsp-on-tag.yml
@@ -19,3 +19,4 @@ jobs:
           image_name: unicode-jsps
           image_tag: ${{ steps.get_tag_name.outputs.GIT_TAG_NAME}}
           dockerfile: ./UnicodeJsps/Dockerfile
+          context: ./UnicodeJsps/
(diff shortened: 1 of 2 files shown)


sffc:
This works.  Alternative, you could also set it up how I have it in https://github.com/unicode-org/jira-github-pr-check.  In that project, I have GitHub synced with a repo on [Cloud Source Repositories](https://cloud.google.com/source-repositories), and then I have a [Google Cloud Build Trigger](https://cloud.google.com/cloud-build/docs/automating-builds/create-manage-triggers) that runs a job whenever the repository is pushed to.  But, what you did here is fine, too.

echeran (approved):
This looks really cool.  The GH Actions triggering on a new tag seems like an easy way to do things.  Clever saving of info in one step as a "variable" value to use in later steps.

IIUC, this workflow is executing a Dockerfile to build up a Docker image with a Jetty webserver that's running the compiled webapp from JSPs as source, and then it pushes the image to GCR (GCP's Docker image registry site thing).  If that's the case, and if this webapp isn't trying to doing anything stateful that requires hitting a backend datastore, then maybe the deployment of the webapp can be done more directly by just using GCP App Engine's support of Java (which apparently runs Jetty for you)?  Would that would help make things easier in any way?  I like living in a world that pretends it doesn't remember J2EE fan favorites.  In addition to specific language APIs, App Engine supports deploying webapps in the form of Docker images, but I'm guessing that's how [the app](https://unicode-jsps-5ocgitonaa-uc.a.run.app/UnicodeJsps/) is being deployed currently?

srl295:
@echeran that's exactly what it's doing. I"ve worked w/ docker, have not worked with any other runtime for GCR… do you want to make an alternative PR ? Or maybe we can move forward on this one and iterate?

echeran:
Oh, definitely move forward with this PR.  I was just thinking out loud about theoretical future ways to improve, in case that's something worth considering.  I have to confess that I haven't used GAE, even though it's technically been around even before Heroku.  Docker is definitely the more precise & configurable & industry-supported way of doing things.

srl295:
@echeran @sffc  can I get a review

sffc (approved):

PR #21, opened 2020-09-08
add license

srl295:


README.md
```diff
@@ -13,3 +13,16 @@ https://sites.google.com/site/unicodetools/home
...
+
+
+### Licenses
+
+- Data and software is governed by the [Unicode Terms of Use](http://www.unicode.org/copyright.html)
+a copy of which is included as [LICENSE](./LICENSE).
+
+### Copyright
+
+Copyright &copy; 1991-2020 Unicode, Inc.
```
markusicu, line 25: Should we use the same lines here as in ICU source files?
```
© 1991 and later: Unicode, Inc. and others.
License & terms of use: http://www.unicode.org/copyright.html
```

srl295: Consistency here is good. Updated to that.

macchiati (approved):

PR #25, opened 2020-09-12
Fix charts & data for emoji 13.1

macchiati:
Accumulated fixes for problems in the charts. Sorry that I didn't separate them into separate PRs.

Most of the changes are 
* updating the version to 13.1
* updates to the internal data files to fix resulting generated files
* retiring or working around some awful hacks to deal with the structure of the skintones with multiperson emoji
* a couple of XUnicodeSet.java chunks of code used in testing.
* some restructuring of code while debugging problems, with some tests for particular cases (just for debugging)
* fixing some directories, since the structure changed when moving to github
* a few updates to the header text for candidate charts
* restucturing ProposalData.SINGLETON because of order-of-static-initialization problems.


unicodetools/org/unicode/tools/emoji/Emoji.java
```diff
@@ -62,6 +62,7 @@
     /**
      * Constants for versions
      */
+    public static final VersionInfo VERSION14 = VersionInfo.getInstance(13,1);
```
pedberg-icu, line 65: Was this going to get fixed? (especially since VERSION_BETA below refers to it)

pedberg-icu (commented):
Emoji.java still seems to have
```
VersionInfo VERSION14 = VersionInfo.getInstance(13,1);
```
Otherwise seems OK

macchiati:
Will fix, thanks

macchiati:
Just fixed the version, and the proposal data.

pedberg-icu (approved):

PR #32, opened 2020-12-02
ICU-21243 Migrates desuffixucd.py to Python 3.

gnrunge:
https://unicode-org.atlassian.net/browse/ICU-21243


markusicu (approved):
lgtm tnx

looks plausible, haven't tried it yet

gnrunge:
I did a basic test. Created a file with the name of the sample file in the
commentin the script, pointed the script to the respective directory, ran
with Python 3, and the script did what it's supposed to do
without complaints.

PR #33, opened 2020-12-09
Fixed Setting, testing UnicodeSet parsing

macchiati:
CLDR dropped some settings that needed to be restored to get unicode tools to work. (Note: there is now a ticket to clean up both CLDR and UnicodeTools settings).

Added some test files for UnicodeSet.

### DIFF
--- a/unicodetools/org/unicode/utilities/UnicodeSetParser.java
+++ b/unicodetools/org/unicode/utilities/UnicodeSetParser.java
@@ -1,0 +1,881 @@
+package org.unicode.utilities;
+
+import java.io.IOException;
+import java.text.ParsePosition;
+import java.util.Set;
+import java.util.TreeSet;
+
+import com.ibm.icu.impl.PatternProps;
+import com.ibm.icu.impl.RuleCharacterIterator;
+import com.ibm.icu.impl.StringRange;
+import com.ibm.icu.impl.Utility;
+import com.ibm.icu.lang.CharSequences;
+import com.ibm.icu.text.SymbolTable;
+import com.ibm.icu.text.UTF16;
+import com.ibm.icu.text.UnicodeMatcher;
+import com.ibm.icu.text.UnicodeSet;
+import com.ibm.icu.util.ICUUncheckedIOException;
+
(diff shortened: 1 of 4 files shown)


unicodetools/org/unicode/utilities/UnicodeSetParser.java
```diff
@@ -0,0 +1,881 @@
...
+import com.ibm.icu.impl.StringRange;
+import com.ibm.icu.impl.Utility;
+import com.ibm.icu.lang.CharSequences;
+import com.ibm.icu.text.SymbolTable;
+import com.ibm.icu.text.UTF16;
+import com.ibm.icu.text.UnicodeMatcher;
+import com.ibm.icu.text.UnicodeSet;
+import com.ibm.icu.util.ICUUncheckedIOException;
+
+public class UnicodeSetParser {
```
markusicu, line 19: What is the purpose of this?
Is it a copy of the parser inside ICU's UnicodeSet?
If so, why make a copy?
What does it do differently?

macchiati:
It is a feasibility experiment, moving the parsing code out of UnicodeSet

markusicu:
> It is a feasibility experiment, moving the parsing code out of UnicodeSet

I had no doubt that that can be done.
If it's a duplicate, it will get out of sync, and then we have two copies to maintain...

macchiati:
It doesn't exist anywhere else; I wanted to see how easy it was, and if so,
propose it to the ICU group, and remove it from here.

Mark

markusicu:
> It doesn't exist anywhere else; I wanted to see how easy it was, and if so, propose it to the ICU group, and remove it from here.

We are not good about cleaning up cruft. I would prefer not to add a big chunk of duplicate code.

It should be possible to pull out the parser, and you showed that in fact it is possible. Good.
But why check it into another repo? Who should use it?

macchiati:
I don't trust having only copies of code on my local machine, unless it
takes zero effort.

I also want to experiment with adding some of the other syntax we are
missing from UTS#18.

Mark

markusicu (approved):

PR #35, opened 2020-12-22
Fix 2 errors in Confusables exposed by change in protectCollections

macchiati:
This might help


Manishearth (approved):
This works, thanks! It's able to do a partial run now, there are further errors, but I can debug them myself. What I mostly needed to know was whether my setup was broken or whether these things were actually bugs.

Issue #18, opened 2020-09-08
Favicon should be changed to the one used for Unicode (tech site)


rickmcgowan:
The blue-and-red "J:" favicon here: https://util.unicode.org/UnicodeJsps/ seems like one that came with something installed there. Users of our tools might be re-assured if this was the same as the favicon used on our tech site here:
http://www.unicode.org/main.html

That red icon can be found here:  http://www.unicode.org/favicon.ico
Or if you're logged into the server it is: /home/httpd/htdocs/favicon.ico

rickmcgowan:
Root of https://util.unicode.org/ is 404. The main page is probably the Utility page:
https://util.unicode.org/UnicodeJsps/
But iif you remove the trailing path component the leftover root is 404 not found.

PR #37, opened 2021-01-11
Add more tests for regex BNF

macchiati:
Add more code to test the UTS #18 EBNF

### DIFF
--- a/unicodetools/org/unicode/tools/emoji/ProposalData.java
+++ b/unicodetools/org/unicode/tools/emoji/ProposalData.java
@@ -21,25 +21,32 @@ import org.unicode.cldr.util.EmojiConstants;
 import org.unicode.cldr.util.RegexUtilities;
 import org.unicode.cldr.util.TransliteratorUtilities;
 import org.unicode.cldr.util.With;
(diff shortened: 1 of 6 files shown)


macchiati:
Adds some files that do parsing at a higher level (not as optimized as UnicodeSet) to make it easier to make changes. Use it to test proposed UTS #18 EBNF updates.

markusicu:
Is this still relevant? Should I assign this to myself for review?

I think I looked at this some time ago and had some concerns about duplicating code from ICU class UnicodeSet... or was that a different discussion?

markusicu:
@macchiati reminder of an old project...

markusicu:
Hi @macchiati it looks like there are two sets of changes here:
- Some changes in emoji-proposal-handling code, adding some extra checks?
- An additional UnicodeSet parser related to the TC39 ECMAScript regex proposal?

Are you intending to take one or both of these into the main branch eventually?

@nedley FYI something to grab from the emoji code changes here?

macchiati:
This was during investigations of regex sets of strings. I'll close it.

PR #52, opened 2021-04-19
Update ReadMe files for Emoji 14.0

nedley:
Not sure why the two files, but both now say “DRAFT” and reference 14.0.


macchiati (approved):
Looks good. There are 2 readmes because the files will end up in 2 different directories.

PR #60, opened 2021-05-26
finish Unicode 14 baseline: add missing files

markusicu:
Copy files verbatim from Unicode 13 / emoji 13.1 where there is no new file yet for Unicode 14.
Finishes the baseline for Unicode 14 work, and when we get data updates for 14 we can see the changes in the file diffs.

(https://github.com/unicode-org/unicodetools/blob/main/docs/inputdata.md)


markusicu:
FYI -- I just copied the files and used `git add`. There is not `git cp` apparently. Hoping that git is smart enough to figure out that these are exact copies of existing files, and to store them by reference.

macchiati (approved):

Manishearth (approved):

PR #68, opened 2021-06-01
fewer env vars, require repo dirs

markusicu:
Groups related path variables into nested subclasses. Makes relationships clearer. Also defers fetching an environment variable until needed, which allows us to fail with an exception when a required path is not specified.

Renames path variables and environment variables to be clearer. *Everyone will need to adjust their environment variables after this change.*

Reduces the number of path environment variables: 1 per repo + 1 for output

### DIFF
--- a/unicodetools/org/unicode/text/utility/Settings.java
+++ b/unicodetools/org/unicode/text/utility/Settings.java
@@ -5,24 +5,13 @@ import org.unicode.cldr.util.CldrUtility;
 
 public class Settings {
 
-    // TODO Many of these settings are crufty and need revision. https://unicode-org.atlassian.net/browse/CLDR-14335 is for fixing these and CLDR.
-    
-    public static final boolean BUILD_FOR_COMPARE = org.unicode.cldr.util.CldrUtility.getProperty("BUILD_FOR_COMPARE", "false").startsWith("t");
+    // TODO Many of these settings are crufty and need revision.
+    // https://unicode-org.atlassian.net/browse/CLDR-14335 "Rationalize CLDR constants"
+    // is for fixing these and CLDR.
 
-    public static final String SVN_DIRECTORY = CldrUtility.getPath(CldrUtility.getProperty("SVN_DIR", CLDRPaths.BASE_DIRECTORY + "/../"), null);
-
-    public static final String AUX_DIRECTORY = CldrUtility.getPath(CldrUtility.getProperty("CLDR_TMP_DIR",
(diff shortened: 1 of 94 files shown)


Manishearth (approved):

unicodetools/org/unicode/text/utility/Settings.java
```diff
@@ -5,60 +5,108 @@
...
-    public static final String SVN_WORKSPACE_DIRECTORY = Utility.fixFileName(
-	    CldrUtility.getProperty("SVN_WORKSPACE", SVN_DIRECTORY)) + "/";
-    public static final String OTHER_WORKSPACE_DIRECTORY = Utility.fixFileName(
-	    CldrUtility.getProperty("OTHER_WORKSPACE", CLDRPaths.LOCAL_DIRECTORY)) + "/";
-
-    public static final String BASE_DIRECTORY = Utility.fixFileName(
-	    CldrUtility.getProperty("BASE_DIRECTORY", SVN_DIRECTORY + "../")) + "/";
+    // TODO: Why do we sometimes use CldrUtility.getPath() which normalizes paths via java.nio.file.Paths
+    // and sometimes Utility.fixFileName() which normalizes paths via java.io.File?
+    // Are they equivalent for our purposes?
```
macchiati, line 14: I suspect they are the same; just accidental.

unicodetools/org/unicode/text/utility/Settings.java
```diff
@@ -5,60 +5,108 @@
...
+         * The root of the unicodetools repo.
+         * Contains the UnicodeJsps and unicodetools folders etc.
+         */
+        public static final String UNICODETOOLS_REPO_DIR =
+                getRequiredPathAndFix("UNICODETOOLS_REPO_DIR");
+        public static final String UNICODETOOLS_DIR = UNICODETOOLS_REPO_DIR + "unicodetools/";
+        public static final String UNICODEJSPS_DIR = UNICODETOOLS_REPO_DIR + "UnicodeJsps/";
+        public static final String DATA_DIR = UNICODETOOLS_DIR + "data/";
+        public static final String UCD_DIR = DATA_DIR + "ucd/";
+        // TODO: IDN_DIR is used, but there is no .../data/IDN/ folder. Should this be .../data/idna/ ?
```
macchiati, line 77: These would take some research.

macchiati (approved):
Much better than it was, thanks! We should do some other changes in concert with CLDR.

PR #76, opened 2021-06-04
refresh Unicode 14 with d17 data files posted 20210604

markusicu:
Refresh source data files verbatim from https://www.unicode.org/Public/


srl295 (approved):
✔️ rslgtm

PR #83, opened 2021-06-10
RecommendedSetGenerator use latestVersion not hardcoded

markusicu:
This tool is used for Unicode version updates in ICU. (Shane wrote it.)


Manishearth (approved):

echeran (approved):
Everything LGTM. I see that the relevant instructions for updating the version in `Settings.java` already exist in https://github.com/unicode-org/unicodetools/blob/main/docs/index.md

sffc (approved):
Apparently I wrote the tool, because my name is in it, but I don't remember what it does very well. :innocent:

PR #84, opened 2021-06-15
Link to charts/index.html from pages in future/

nedley:
Rick noted that the link in the top bar of the pages in future/ go to the wrong page. As far as I can tell, this particular link is the only one that escapes that directory and needs to link to index.html in the sibling directory.


markusicu (approved):
lgtm

PR #86, opened 2021-06-18
fix unicode.org/unicode/ URLs

markusicu:
These relied on old redirects.


rickmcgowan (approved):
These all look correct to me. Thanks for taking care of them.

PR #92, opened 2021-07-10
Unicode 14 data files as of 20210709

markusicu:


macchiati (approved):

srl295 (approved):
don't see any issues… so, i'll say RSLGTM

Issue #97, opened 2021-07-16
Fix Extended_Pictographic in UnicodeProperty subclasses


macchiati:
This is both in the JSPs and in the regular unicodetools.

There is a special hack for performance that works for most properties. Only one character in the unassigned range is tested for the property in some of the internals if the property is marked as hasUniformUnassigned() = true.

If the value is false, then every unassigned character is tested. It is the responsibility of each UnicodeProperty subclass to mark any exceptions. For example, 

`    protected ICUProperty(String propName, int propEnum) {
      setName(propName);
      this.propEnum = propEnum;
      setType(internalGetPropertyType(propEnum));
      if (propEnum == UProperty.DEFAULT_IGNORABLE_CODE_POINT
              || propEnum == UProperty.BIDI_CLASS
              || propEnum == UProperty.BLOCK
              || propEnum == UProperty.EAST_ASIAN_WIDTH
              || propEnum == UProperty.LINE_BREAK
              || propEnum == UProperty.NONCHARACTER_CODE_POINT
              || propEnum == UProperty.PATTERN_SYNTAX
              || propEnum == UProperty.PATTERN_WHITE_SPACE
              || propEnum == UProperty.CHANGES_WHEN_CASEFOLDED
              || propEnum == UProperty.EMOJI
              || propEnum == UProperty.EMOJI_MODIFIER
              || propEnum == UProperty.EMOJI_MODIFIER_BASE
              || propEnum == UProperty.EMOJI_PRESENTATION
              ) {
        setUniformUnassigned(false);
      }
    }
`

New properties like Extended_Pictographic need to be added to lists like this. The way to fix it is to search for setUniformUnassigned(false) and check that the lists are right.

NOTE: This performance hack was from a long time ago, and may not be needed anymore, but we should analyze the impact before removing.

srl295:
what would be best is to have a test that verifies that the heuristic is correct… test _all_ unassigned codepoints.

macchiati:
I agree. Could be pretty simple. 

1. Create a subclass of UnicodeProperty that wraps any other, but resets setUniformUnassigned(true).
2. Walk through all the properties that have hasUniformUnassigned() = true, and create a wrapped class.
3. For each such case, walk through all Unicode code points to verify that both instances have identical values.

nedley:
Is this issue distinct from #53?

macchiati:
53 is a duplicate (and has less information)

Mark

srl295:
[...]

PR #105, opened 2021-07-20
Fix Extended_Pictographic in UnicodeProperty for JSP

srl295:
For #97

This is the only property-specific site that calls setUniformUnassigned that I could find so far, and it's in the JSPs.  Should there be a site within the main tools that has `NONCHARACTER_CODE_POINT` etc? 

I'll get this posted somewhere and then work on adding a test.


macchiati (approved):

macchiati:
CLDR has the same code (but outdated). I don't think it matters there,
since we want to end up removing UnicodeProperty from CLDR.

Mark

markusicu (approved):
lgtm

macchiati:
Great! Could you please let me know when it is in a testing version?

Mark

srl295:
@macchiati On top of U13?  I am building a tag now…

PR #109, opened 2021-08-11
Unicode 14 data files as of 20210811

markusicu:
verbatim from https://www.unicode.org/Public/


Manishearth (approved):

srl295 (approved):
spot checked and rslgtm

PR #114, opened 2021-08-17
Unicode 14 line break rule changes

markusicu:
- remove the redundant first of three LB27 rules
- add a future-proofing LB30b rule for code points reserved for emoji (unassigned ExtPict)


Manishearth (approved):

macchiati (approved):
looks great. If you don't want to have $ExtPictUnassigned cause extra values, I *think* you can drop line 108, and make 30.22 be [\p{Extended_Pictographic}&\p{Cn}] × $EM

markusicu:
> looks great. If you don't want to have $ExtPictUnassigned cause extra values, I _think_ you can drop line 108, and make 30.22 be [\p{Extended_Pictographic}&\p{Cn}] × $EM

Thanks! It would be nice if we could define variables without exploding the combinatorics. Will consider next time.

PR #115, opened 2021-08-17
Unicode 14: temp hack in IdnaTestV2 generator, and URL fix

markusicu:
For action item 164-A52 In IdnaTestV2.txt, fix the URL .../tr46/proposed.html#Processing to point to the latest version of UTS 46
(also removed some trailing spaces)


srl295 (approved):
ouch.

Issue #129, opened 2021-09-23
update bidi util to Unicode 14/UBA 14


markusicu:
Ken writes:

I've updated, rebuilt, tested, and deployed the C version of the bidiref code for 14.0. See:

https://www.unicode.org/Public/PROGRAMS/BidiReferenceC/14.0.0/

I've checked that this update runs o.k. and passes both the BidiTest.txt and BidiCharacterTest.txt sets of conformance test cases with no complaints.

In principle, the next step would be the update of:

https://util.unicode.org/UnicodeJsps/bidic.jsp

but I don't have any experience with the jsp's. Currently that jsp is using the 13.0 version of bidiref1.c, although the demonstration program itself hasn't yet been updated from UBA 11.0.

So if you run the jsp currently, you see the bidiref output in the "Analysis" section identified as "Initialized bidiref 13.0.0 library for UBA version 11.0".

The next steps, when somebody who knows how to do this has the time, presumably would be:

Grab and build the 14.0.0 version of bidiref1.c. (See directions in the top level ReadMe.txt in the 14.0.0 directory.) Deploy that and 14.0.0 versions of UnicodeData.txt and BidiBrackets.txt to wherever this jsp is expecting this executable to be running.

Then in the Java code itself, update the UBA Version pulldown widget to add 12.0, 13.0, and 14.0 and to default to 14.0.

In the static template text at the top of the bidic.jsp page, update the "Versions 6.2 through 11.0" to "Versions 6.2 through 14.0", with the 14.0 link going to https://www.unicode.org/reports/tr9/tr9-44.html and then the 11.0 link for the C Reference Implementation, also to "14.0" and going to https://www.unicode.org/Public/PROGRAMS/BidiReferenceC/14.0.0/

I don't know if there is anything else that is version-specific in the Java code -- there shouldn't be, but someone should check.

srl295:
as to C, you should just have to update the version at https://github.com/unicode-org/unicodetools/blob/HEAD/UnicodeJsps/Dockerfile#L3

I'm not sure about the UBA version pulldown.

srl295:
- [X] C code
- [ ] C pulldown
- [ ] J code/pulldown

srl295:
J says:

> This Java implementation supports UBA rules and character properties as of Unicode 6.2.

so it looks like there's nothing to do for Java

markusicu:
[...]

PR #135, opened 2021-10-11
bump: BidiReferenceC@4.0.0

srl295:
- just update CVERSION=14.0.0
 - does not update any of the Java/JSP content

Refs: https://github.com/unicode-org/unicodetools/issues/129


srl295:
as discussed earlier

PR #136, opened 2021-10-11
Update docs for setup of Maven + {CLI,Eclipse}

### LINKED ISSUE #133: doc: improve docs around maven
comments from https://github.com/unicode-org/unicodetools/pull/91

echeran:
Fixes #133 

I tried to refactor the instructions into shared setup + CLI-specific/Eclipse-specific.  

We should verify and expand the list of the common tasks that are invoked from unicodetools.  (And then make sure that they are reflected somewhere in the Github Actions CI jobs.)

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -27,13 +27,36 @@ can be used to:
 
 ## Instructions
 
-1.  Set up Eclipse and CLDR, according to instructions:
+### General Setup for Maven
+
+1.  Configure Maven settings (including Github tokens) according to these instructions:
     <http://cldr.unicode.org/development/maven>
+2.  Get your github account authorized for <https://github.com/unicode-org/unicodetools>,
+    create a fork under your account, and create a local clone.
+3.  Clone both the [Unicode Tools](https://github.com/unicode-org/unicodetools) and [CLDR](https://github.com/unicode-org/cldr) repositories:
+```
+git clone https://github.com/unicode-org/unicodetools.git
+git clone https://github.com/unicode-org/cldr.git
+```
+4.  In the root of the `unicodetools` local working copy, create the `Generated` folder 
+    1. (Eclipse users can do this graphically by following the corresponding step in the Eclipse section below)
+    1. At the command-line: `cd <unitools>; mkdir -p Generated/BIN`
+5. The tools and commands are parameterized such that important configuration values are set as Java System properties at runtime as JVM arguments. Here is a list of property names and example values:
+| Property                | Example Value                                     |
+|-------------------------|---------------------------------------------------|
+| CLDR_DIR                | /usr/local/google/home/mscherer/cldr/uni/src      |
+| IMAGES_REPO_DIR         | /usr/local/google/home/mscherer/images/mine/src   |
+| UNICODETOOLS_REPO_DIR   | /usr/local/google/home/mscherer/unitools/mine/src |
+| UNICODETOOLS_OUTPUT_DIR | /usr/local/google/home/mscherer/unitools/mine     |
+| UVERSION                | 14.0.0                                            |
+
+
+### Eclipse-specific Additional Setup
+
+1. Follow the Eclipse-specific settings at <http://cldr.unicode.org/development/maven>
     1.  Edit the cldr-code project’s Build Path:
         Under “Order and Export”, set the check mark next to “Maven Dependencies”
         so that CLDR makes its dependencies available to the Unicode Tools project.
-2.  Get your github account authorized for <https://github.com/unicode-org/unicodetools>,
-    create a fork under your account, and create a local clone.
 3.  Import the unicodetools project into Eclipse. (Using Maven: General > Existing Projects into Workspace)
 4.  Also create the project **and directory** Generated. Various results are
     deposited there. You need the directory, but the Eclipse project is optional.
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -44,36 +68,54 @@ can be used to:
...
+Eclipse users:
+- Use the VM argument `-ea` (enable assertions) in your Preferences or in your Run/Debug configurations
 
+### Commands for Unicode Tools Tasks
 
-# MakeUnicodeFiles
-mvn exec:java -Dexec.mainClass="org.unicode.text.UCD.Main" -Dexec.args="version 14.0.0 build MakeUnicodeFiles" -pl unicodetools  -DCLDR_DIR=$(cd ../cldr;pwd) -DUNICODETOOLS_OUTPUT_DIR=$(cd ..;pwd) -DUNICODETOOLS_REPO_DIR=$(pwd) -DUVERSION=14.0.0
-# build and test
-mvn package -DCLDR_DIR=$(cd ../cldr;pwd) -DUNICODETOOLS_OUTPUT_DIR=$(cd ..;pwd) -DUNICODETOOLS_REPO_DIR=$(pwd) -DUVERSION=14.0.0
+Common tasks for Unicode Tools are listed below with example CLI commands with example argument values that they need:
```
echeran, line 107: What are all the basic unicodetools tasks and corresponding CLI command examples that should be listed here?

srl295 (dismissed):
LGTM definite improvement

echeran:
@srl295 Do you know how [this line in the CI](https://github.com/unicode-org/unicodetools/blob/main/.github/workflows/build-jsp.yml#L42) actually succeeds (given the previous line creating a real directory in the same place)?

srl295:
> @srl295 Do you know how [this line in the CI](https://github.com/unicode-org/unicodetools/blob/main/.github/workflows/build-jsp.yml#L42) actually succeeds (given the previous line creating a real directory in the same place)?

```
ln -s $(pwd)/unicodetools/data/security ./output/Generated/ &&
```

Yes. This ends up with `output/Generated/security` as a symlink to _repo_`/unicodetools/data/security` so that the in-repo security files appear as if they had been generated whereas they are actually already in the repo.

I don't know why this is necessary. It seems to satisfy prerequisites for some tests or processes.

docs/index.md
```diff
@@ -27,13 +27,37 @@ can be used to:
...
+    create a fork under your account, and create a local clone.
+3.  Clone both the [Unicode Tools](https://github.com/unicode-org/unicodetools) and [CLDR](https://github.com/unicode-org/cldr) repositories:
+```
+git clone https://github.com/unicode-org/unicodetools.git
+git clone https://github.com/unicode-org/cldr.git
+```
+4.  In the root of the `unicodetools` local working copy, create the `Generated` folder 
+    1. (Eclipse users can do this graphically by following the corresponding step in the Eclipse section below)
+    1. At the command-line: `cd <unitools>; mkdir -p Generated/BIN`
+5. The tools and commands are parameterized such that important configuration values are set as Java System properties at runtime as JVM arguments. Here is a list of property names and example values:
```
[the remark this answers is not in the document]
echeran: I updated the command for the table to indicate that these path values are for an out-of-source build workspace setup.

docs/index.md
```diff
@@ -27,13 +27,37 @@ can be used to:
...
+1.  Configure Maven settings (including Github tokens) according to these instructions:
     <http://cldr.unicode.org/development/maven>
+2.  Get your github account authorized for <https://github.com/unicode-org/unicodetools>,
+    create a fork under your account, and create a local clone.
+3.  Clone both the [Unicode Tools](https://github.com/unicode-org/unicodetools) and [CLDR](https://github.com/unicode-org/cldr) repositories:
+```
+git clone https://github.com/unicode-org/unicodetools.git
+git clone https://github.com/unicode-org/cldr.git
+```
+4.  In the root of the `unicodetools` local working copy, create the `Generated` folder
```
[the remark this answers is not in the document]
echeran: I added notes for in-source and out-of-source build setups. I tested out both sets of setup & task commands, and they run without errors locally.

FYI @srl295 I also deduced that the `mvn ... compile` step is somehow performing some sort of pre-requisite initialization step.  I'm not sure how it is the case that the `mvn ... compile` command has to be run first before the other commands should be run (including a `mvn ... package` command), since `compile` is a precursor to `package` in the Maven lifecycle progression whenever `package` is run. My naive suspicion is that there are some settings for <repositories> in that `.github/worklfows/mvn-settings.xml` that could (should?) be moved up to the parent `pom.xml`. We don't need to solve it here, though, so just FYI & food for thought.

docs/index.md
```diff
@@ -27,53 +27,130 @@ can be used to:
...
+2.  Get your github account authorized for <https://github.com/unicode-org/unicodetools>,
+    create a fork under your account, and create a local clone.
+
+#### In-source / Out-of-source build setup
+
+Some of the tasks within the Unicode Tools generate output files that can also be input files into other steps.
+For this purpose, we create a folder named `Generated` to store these files.
+This folder can be a subfolder inside the local working copy root (called an "In-source build" workspace layout), or this folder can be outside (ex: a sibling folder) the local working-copy root (called an "Out-of-source build" workspace layout). Both workspace styles are described below.
+
+Out-of-source builds keep a separation between source files of the repository and their generated output files, which are no tracked in the repository. Out-of-source builds allow developers to maintain a clean view of changes to tracked source files, without mixing generated output files. (Out-of-source builds are also useful for C++ repositories in which multiple configurations can be invoked to generate independent sets of makefiles that result in corresponding different output compiled binary files.
```
markusicu, line 43:
```suggestion
Out-of-source builds keep a separation between source files of the repository and their generated output files, which are not tracked in the repository. Out-of-source builds allow developers to maintain a clean view of changes to tracked source files, without mixing generated output files. (Out-of-source builds are also useful for C++ repositories in which multiple configurations can be invoked to generate independent sets of makefiles that result in corresponding different output compiled binary files.)
```

docs/index.md
```diff
@@ -27,53 +27,130 @@ can be used to:
...
+mkdir -p cldr/mine/src
+```
+2. Clone the repositories' contents into their respective source directories:
+```
+git clone https://github.com/unicode-org/unicodetools.git unicodetools/mine/src
+git clone https://github.com/unicode-org/cldr.git cldr/mine/src
+```
+3. Create the `Generated` folder structure as a sibling to the local working copy root:
+```
+mkdir -p unicodetools/mine/output/Generated/BIN
```
markusicu, line 59: The extra "output" folder level seems redundant with "Generated". I suggest removing the "output" layer.

docs/index.md
```diff
@@ -27,53 +27,130 @@ can be used to:
...
+ln -s ../../src/unicodetools/data/security unicodetools/mine/output/Generated/security
```
markusicu, line 60: I think that you & Steven have discussed this line in the context of CI, but in my running tools to generate security files I have so far not created such a symlink, and I am pretty sure that we used to not have any instructions for such a thing. What problem does this symlink solve?

docs/index.md
echeran: The [code expects](https://github.com/unicode-org/unicodetools/blob/main/unicodetools/src/main/java/org/unicode/text/utility/Settings.java#L151-L154) `Generated` to be a subdirectory of the "root of where we write output files". There is [at least one instance](https://github.com/unicode-org/unicodetools/blob/main/unicodetools/src/main/java/org/unicode/tools/Ids.java#L1267) where a different subdirectory within this root output folder is used. Also, getting rid of the output directory structure would make the UNICODETOOLS_OUTPUT_DIR system property value effectively be the parent of the UNICODETOOLS_REPO_DIR.

Do you want to document the meaning of the Java system properties, either here in this PR (ex: adding a 3rd column to the table below) or in the future?

docs/index.md
echeran: If I remove the symlink, the first listed command succeeds (Make Unicode Files), but the second command fails (Build and Test) because it can't find the TestSecurity classes and a related file:

```
[ERROR] Tests run: 8, Failures: 0, Errors: 8, Skipped: 0, Time elapsed: 0.554 s <<< FAILURE! - in org.unicode.test.TestSecurity
[ERROR] org.unicode.test.TestSecurity.TestCldrConsistency  Time elapsed: 0 s  <<< ERROR!
java.lang.ExceptionInInitializerError
        at org.unicode.test.TestSecurity.<clinit>(TestSecurity.java:52)
Caused by: java.io.FileNotFoundException: /usr/local/google/home/elango/oss/tmp/unicodetools-docs-testing-2/unicodetools/output/Generated/security/14.0.0/IdentifierType.txt (No such file or directory)
        at org.unicode.test.TestSecurity.<clinit>(TestSecurity.java:52)

[ERROR] org.unicode.test.TestSecurity.TestIdempotence  Time elapsed: 0.001 s  <<< ERROR!
java.lang.NoClassDefFoundError: Could not initialize class org.unicode.test.TestSecurity

[ERROR] org.unicode.test.TestSecurity.TestSpacing  Time elapsed: 0 s  <<< ERROR!
java.lang.NoClassDefFoundError: Could not initialize class org.unicode.test.TestSecurity

[ERROR] org.unicode.test.TestSecurity.TestXidMod  Time elapsed: 0 s  <<< ERROR!
java.lang.NoClassDefFoundError: Could not initialize class org.unicode.test.TestSecurity

[ERROR] org.unicode.test.TestSecurity.TestConfusables  Time elapsed: 0 s  <<< ERROR!
java.lang.NoClassDefFoundError: Could not initialize class org.unicode.test.TestSecurity

[ERROR] org.unicode.test.TestSecurity.TestScriptDetection  Time elapsed: 0 s  <<< ERROR!
java.lang.NoClassDefFoundError: Could not initialize class org.unicode.test.TestSecurity

[ERROR] org.unicode.test.TestSecurity.TestWholeScripts  Time elapsed: 0 s  <<< ERROR!
java.lang.NoClassDefFoundError: Could not initialize class org.unicode.test.TestSecurity

[ERROR] org.unicode.test.TestSecurity.TestWholeScriptData  Time elapsed: 0 s  <<< ERROR!
java.lang.NoClassDefFoundError: Could not initialize class org.unicode.test.TestSecurity
```

docs/index.md
markusicu: > The [code expects](https://github.com/unicode-org/unicodetools/blob/main/unicodetools/src/main/java/org/unicode/text/utility/Settings.java#L151-L154) `Generated` to be a subdirectory of the "root of where we write output files". There is [at least one instance](https://github.com/unicode-org/unicodetools/blob/main/unicodetools/src/main/java/org/unicode/tools/Ids.java#L1267) where a different subdirectory within this root output folder is used.

Yes, I rewrote Settings.java earlier this year to look like this. I intended, documented, and used this with Generated as a sibling of the repo root (my "src"), which means that the UNICODETOOLS_OUTPUT_DIR is the parent folder of where "src" is.

> Also, getting rid of the output directory structure would make the UNICODETOOLS_OUTPUT_DIR system property value effectively be the parent of the UNICODETOOLS_REPO_DIR.

Yes, exactly. As intended.
I am not sure if anything writes anywhere outside of the Generated child of the UNICODETOOLS_OUTPUT_DIR. Either everything or almost everything works inside GEN_DIR.
We would only get a problem if we wrote into a folder directly in UNICODETOOLS_OUTPUT_DIR whose name happened to collide with that of the repo root folder ("src" in my case).

> Do you want to document the meaning of the Java system properties, either here in this PR (ex: adding a 3rd column to the table below) or in the future?

Too much detail. We have comments in Settings.java for that, and sometimes it just boils down to look at which code uses which path.

docs/index.md
markusicu: > If I remove the symlink, the first listed command succeeds (Make Unicode Files), but the second command fails (Build and Test) because it can't find the TestSecurity classes and a related file:

That seems wrong; or rather, it seems like an artifact of trying to run TestSecurity as a unit test in isolation. I have run that test successfully, and without a symlink, according to the relevant instructions: The point is that we normally want to run TestSecurity after GenerateConfusables which writes files into .../Generated/security. In other words, we write new files and then see if they work. That's why TestSecurity wants to look at the output files.

See https://github.com/unicode-org/unicodetools/blob/main/docs/security.md

If we want to make this work as a unit test that looks at the input data without running GenerateConfusables, then TestSecurity should get a parameter for looking at input data vs. output data.

Putting an unconditional instruction for this symlink into the general setup will make GenerateConfusables malfunction.

More generally, remember that we used to not routinely run unit tests. What tests people ran they ran ad hoc after running other tools. As we move to real, continous unit tests, we need to keep in mind what the test code was intended to do, and adapt it to what we need.

docs/index.md
markusicu: PS: There is another way of collapsing "output" and "Generated": We could change GEN_DIR to be the same as UNICODETOOLS_OUTPUT_DIR rather than a child folder. Then UNICODETOOLS_OUTPUT_DIR could be a sibling to "src" without concerns about a child folder accidentally colliding with the repo root. We would lose the special name "Generated" but that's probably ok; we would need to change other docs pages for that.

That would require a code change. I would first document what we have now, and if we agree to collapse the two levels, then make another PR that changes code & docs.

docs/index.md
srl295: @markusicu 
> an artifact of trying to run TestSecurity as a unit test in isolation

Could TestSecurity be changed to read the `src/unicodetools/data/security` files as input instead, unless invoked by `GenerateConfusables` ? 

`security.md` could be helpful, but we kind of need the overall process description. I'm still not sure what all of the input and output terms are. I guess I could try going through each of the .md files and trying to perform their steps.

srl295 (commented):
not going to approve while the 'security' part is still being sorted out, but generally lgtm

docs/index.md
markusicu: > Could TestSecurity be changed to read the `src/unicodetools/data/security` files as input instead, unless invoked by `GenerateConfusables` ?

Sure, just a matter of programming :-)
As I said above, this could be a new parameter. Or I suppose it could automatically try the input path if there is nothing at the output path.

Let's please remove the symlink line from the developer setup instructions, get this PR in, and then fiddle with TestSecurity.

> `security.md` could be helpful, but we kind of need the overall process description. I'm still not sure what all of the input and output terms are. I guess I could try going through each of the .md files and trying to perform their steps.

We need to fix the setup instructions (this PR), and we will do a survey of all of the "tools", but it makes sense to have specific docs pages per tool or group of tools.

srl295:
> Let's please remove the symlink line from the developer setup instructions, get this PR in, and then fiddle with TestSecurity.

Maybe Remove it with a TODO warning that tests will fail. 

Sounds like a good path forward.

markusicu:
> > Let's please remove the symlink line from the developer setup instructions, get this PR in, and then fiddle with TestSecurity.
> 
> Maybe Remove it with a TODO warning that tests will fail.

Running TestSecurity without first running the generator will make the test code fail, yes.
A TODO is fine.
And of course keep the symlink in the CI bot until TestSecurity has been modified.

markusicu (approved):
lgtm tnx! pse *squash*-and-merge

PR #169, opened 2021-11-29
update bidi util to Unicode 14/UBA 14

### LINKED ISSUE #129: update bidi util to Unicode 14/UBA 14

srl295:
- CVERSION was bumped in 161a4917499f22253dd34ec7336f47de69fe950e (#135)
- Just need to update the jsp text
- Reformatted HTML to make variable parts clearer and easier to update.

Note: Bidi-J is on a fixed version and so was not changed.

Fixes #129

### DIFF
--- a/UnicodeJsps/src/main/webapp/bidic.jsp
+++ b/UnicodeJsps/src/main/webapp/bidic.jsp
@@ -221,8 +221,12 @@ function setUbaInput(str) {
 <%@ include file="subtitle.jsp" %>
 <p><a target="help" href="https://unicode-org.github.io/unicodetools/help/breaks"><b>help</b></a> | <%@ include file="others.jsp" %></p>
 <p>Shows processing of a single paragraph of text by the Unicode Bidirectional Algorithm (UBA),
-    Versions <a target="doc" href="http://www.unicode.org/reports/tr9/tr9-27.html">6.2</a> through <a target="doc" href="http://www.unicode.org/reports/tr9/tr9-39.html">11.0</a>,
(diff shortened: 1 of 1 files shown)


markusicu (approved):
lgtm

PR #173, opened 2021-12-06
Clean up output dir settings

### LINKED ISSUE #150: clean up output dir setting
Settings.java has
```
    public static final class Output {
        /** The root of where we write output files. Most go into a "Generated" sub-folder. */
        public static final String UNICODETOOLS_OUTPUT_DIR =
                getRequiredPathAndFix("UNICODETOOLS_OUTPUT_DIR");
        public static final String GEN_DIR = UNICODETOOLS_OUTPUT_DIR + "Generated/";
        public static final String BIN_DIR = GEN_DIR + "BIN/";
        ...
```
In an "out of source build" setup, you want to put the Generated folder next to the repo source folder. With these existing settings, that means you point the UNICODETOOLS_OUTPUT_DIR to the *parent* of the repo source folder.

This feels a bit weird and unsafe. It could theoretically let tools write over the repo source if there is a name collision.

- [ ] There are very few places that write directly into UNICODETOOLS_OUTPUT_DIR. We should change them to write into GEN_DIR + something instead.
- [ ] And then we should change the environment variable to point one level down to what in this setup is the sibling of the repo source folder -- conventionally called "Generated". For clarity, we could rename the environment variable from UNICODETOOLS_OUTPUT_DIR to UNICODETOOLS_GEN_DIR.

echeran:
Fixes #150 

- [x] There are very few places that write directly into UNICODETOOLS_OUTPUT_DIR. We should change them to write into GEN_DIR + something instead.
- [x] And then we should change the environment variable to point one level down to what in this setup is the sibling of the repo source folder -- conventionally called "Generated". For clarity, we could rename the environment variable from UNICODETOOLS_OUTPUT_DIR to UNICODETOOLS_GEN_DIR.
- [x] Also need to make sure that the instructions in `docs/build.md` for Maven CLI steps is in sync with these changes.


srl295 (approved):
LGTM though I wonder— 

should `build-jsp.yml` be folded in? I think the only thing that isn't in `cli-buildinstructions` is the packaging of the JSPs…

srl295:
> LGTM though I wonder— 
> 
> should `build-jsp.yml` be folded in? I think the only thing that isn't in `cli-buildinstructions` is the packaging of the JSPs…

https://github.com/unicode-org/unicodetools/blob/807a088becd173df0afafc25c2d3c09bf62331e3/.github/workflows/build-jsp.yml#L55-L66

just move these lines into cli-buildinstructions.yml and delete the other?

echeran:
> LGTM though I wonder— 
> 
> should `build-jsp.yml` be folded in? I think the only thing that isn't in `cli-buildinstructions` is the packaging of the JSPs…

@markusicu was asking me something similar. I wanted to take things incrementally. I didn't want to touch any of our working CI jobs -- let them keep working. My main intention just want to make sure that the build.rm instructions are actually correct -- verified by the new CI job -- and be able to make sense of it for myself a little better. 

I think the Build JSPs job should remain separate (whether in the same or a separate workflow file doesn't matter) because that will allow us to show what are the minimal set of steps (and their ordering) required to build the JSPs, whereas the cli-buildinstructions.yml could theoretically grow to include all the useful commands in our processes -- those that we know and those that we don't yet know. So for example, for the job to build JSPs, can we get rid of the "Generate Unicode data" step + "Run unicodetools tests" , or can we only safely get rid of the latter, or neither?

But I'll create a followup ticket just so that the CI workflows are running through their steps consistently (I realized through this PR that build-jsp.yml checks out the `cldr` working copy as a child of the `unicodetools` working copy, so we should run `mv cldr ..` to make it a sibling directory, in order to proceed with the instructions in build.md). Also, at the same time, I think might be able to fold in mvn-settings.xml into the parent pom.

markusicu:
> > should `build-jsp.yml` be folded in? I think the only thing that isn't in `cli-buildinstructions` is the packaging of the JSPs…
> 
> @markusicu was asking me something similar. I wanted to take things incrementally. I didn't want to touch any of our working CI jobs -- let them keep working. My main intention just want to make sure that the build.rm instructions are actually correct -- verified by the new CI job -- and be able to make sense of it for myself a little better.

I am ok with this to get a good warm & fuzzy feeling about the process and instructions.
Longer-term, I suspect that we will want to re-consolidate the workflows. I am skeptical that we will find lasting value of two workflows building & running the same stuff.

> I think the Build JSPs job should remain separate (whether in the same or a separate workflow file doesn't matter) because that will allow us to show what are the minimal set of steps (and their ordering) required to build the JSPs, whereas the cli-buildinstructions.yml could theoretically grow to include all the useful commands in our processes -- those that we know and those that we don't yet know. So for example, for the job to build JSPs, can we get rid of the "Generate Unicode data" step + "Run unicodetools tests" , or can we only safely get rid of the latter, or neither?

If they diverge meaningfully, then it makes sense to have separate workflows, or maybe just separate jobs.

.github/workflows/cli-build-instructions.yml
```diff
@@ -0,0 +1,99 @@
...
+      - name: Cache local Maven repository
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      - name: Set up out-of-source output dir
+        run: |
+          mkdir -p unicodetools/mine/output/Generated/BIN
```
markusicu, line 80:
```suggestion
          mkdir -p unicodetools/mine/Generated/BIN
```
Let's drop the "output" layer in the out-of-source version.

docs/build.md
```diff
@@ -72,32 +72,19 @@ git clone https://github.com/unicode-org/cldr.git
...
-| IMAGES_REPO_DIR         | /usr/local/google/home/mscherer/images/mine/src      |
-| UNICODETOOLS_REPO_DIR   | /usr/local/google/home/mscherer/unitools/mine/src    |
-| UNICODETOOLS_OUTPUT_DIR | /usr/local/google/home/mscherer/unitools/mine        |
-| UVERSION                | 14.0.0                                               |
+| Property                | Example Value                                                  |
+|-------------------------|----------------------------------------------------------------|
+| CLDR_DIR                | /usr/local/google/home/mscherer/cldr/mine/src                  |
+| IMAGES_REPO_DIR         | /usr/local/google/home/mscherer/images/mine/src                |
+| UNICODETOOLS_REPO_DIR   | /usr/local/google/home/mscherer/unitools/mine/src              |
+| UNICODETOOLS_GEN_DIR    | /usr/local/google/home/mscherer/unitools/mine/output/Generated |
```
markusicu, line 86:
```suggestion
| UNICODETOOLS_GEN_DIR    | /usr/local/google/home/mscherer/unitools/mine/Generated |
```
Let's drop the "output" layer in the out-of-source version.

pom.xml
```diff
@@ -102,7 +102,7 @@
 							<java.awt.headless>true</java.awt.headless>
 							<UNICODETOOLS_REPO_DIR>${project.basedir}/..</UNICODETOOLS_REPO_DIR>
 							<!-- sibling of unicodetools-->
-							<UNICODETOOLS_OUTPUT_DIR>${project.basedir}/../output</UNICODETOOLS_OUTPUT_DIR>
+							<UNICODETOOLS_GEN_DIR>${project.basedir}/../output/Generated</UNICODETOOLS_GEN_DIR>
```
markusicu, line 105:
```suggestion
							<UNICODETOOLS_GEN_DIR>${project.basedir}/../Generated</UNICODETOOLS_GEN_DIR>
```

unicodetools/src/main/java/org/unicode/text/utility/Settings.java
```diff
@@ -258,23 +258,24 @@ public String versionToString(VersionInfo version) {
...
         /** The root of where we write output files. Most go into a "Generated" sub-folder. */
-        public static final String UNICODETOOLS_OUTPUT_DIR =
-                getRequiredPathAndFix("UNICODETOOLS_OUTPUT_DIR");
-        public static final String GEN_DIR = UNICODETOOLS_OUTPUT_DIR + "Generated/";
+        public static final String GEN_DIR =
+                getRequiredPathAndFix("UNICODETOOLS_GEN_DIR");
         public static final String BIN_DIR = GEN_DIR + "BIN/";
         public static final String GEN_UCD_DIR = GEN_DIR + "UCD/";
         public static final String GEN_UCA_DIR = GEN_DIR + "UCA/";
+        public static final String DATA_DIR = GEN_DIR + "DATA/";
```
markusicu, line 266: This could be a little confusing, because there is also a Settings.UnicodeTools.DATA_DIR.

It looks like all of these tools currently also write into a subfolder of whatever..._DIR they start from -- so why not just remove this line here, and change the Output.DATA_DIR in your current version to simply Output.GEN_DIR?

echeran: Done.

pom.xml
echeran: Done.

docs/build.md
echeran: Done.  I think my previous attempt at doing this at the end of a previous PR was incomplete / inconsistent somehow and caused the CLI instructions to not actually work. Anyways, the new CI workflow is still passing in the PR.

.github/workflows/cli-build-instructions.yml
echeran: Done.

srl295 (approved):
Changes LGTM

srl295:
> > > should `build-jsp.yml` be folded in? I think the only thing that isn't in `cli-buildinstructions` is the packaging of the JSPs…
> > 
> > 
> > @markusicu was asking me something similar. I wanted to take things incrementally. I didn't want to touch any of our working CI jobs -- let them keep working. My main intention just want to make sure that the build.rm instructions are actually correct -- verified by the new CI job -- and be able to make sense of it for myself a little better.

Concrete proposal:
1. you finish and merge this PR
2. I'll make a followon afterwards, and delete the other workflow.

> I am ok with this to get a good warm & fuzzy feeling about the process and instructions. Longer-term, I suspect that we will want to re-consolidate the workflows. I am skeptical that we will find lasting value of two workflows building & running the same stuff.


@echeran your workflow is *already building* the JSPs here: <https://github.com/unicode-org/unicodetools/runs/4452984960?check_suite_focus=true#step:10:770>

All that's remaining is to verify that the docker build works.

> > I think the Build JSPs job should remain separate (whether in the same or a separate workflow file doesn't matter) because that will allow us to show what are the minimal set of steps (and their ordering) required to build the JSPs, whereas the cli-buildinstructions.yml could theoretically grow to include all the useful commands in our processes -- those that we know and those that we don't yet know. So for example, for the job to build JSPs, can we get rid of the "Generate Unicode data" step + "Run unicodetools tests" , or can we only safely get rid of the latter, or neither?
> 
> If they diverge meaningfully, then it makes sense to have separate workflows, or maybe just separate jobs.

Those two steps were to get the *unicodetools* build to complete. They weren't specific to the JSP.

markusicu (approved):
lgtm tnx
please remember to *squash and merge*

PR #180, opened 2021-12-09
Uni 15: new block Arabic_Extended_C with default Bidi_Class=AL

markusicu:
for action item 169-A8: In DerivedBidiClass.txt, change the default Bidi_Class value for 10EC0..10EFF from R to AL, for Unicode version 15.0.


Ken-Whistler (approved):
Ok.

PR #184, opened 2021-12-14
Simple code for checking bidi issues in source code.

macchiati:
Just for reference


markusicu:
@macchiati & @eggrobin say that we can close this.

PR #185, opened 2021-12-14
jsp: fix C bidi pulldown

srl295:
https://github.com/unicode-org/unicodetools/issues/129 (but was missed when that was closed)

### DIFF
--- a/UnicodeJsps/src/main/java/org/unicode/jsp/UBAVersion.java
+++ b/UnicodeJsps/src/main/java/org/unicode/jsp/UBAVersion.java
@@ -1,0 +1,45 @@
+package org.unicode.jsp;
+
+import java.util.EnumSet;
+
+import org.unicode.props.UcdPropertyValues.Age_Values;
+
+/**
+ * A class to encapsulate the available C UBA versions.
+ */
+public class UBAVersion {
+    private static EnumSet<Age_Values> C_UBA_AGES = EnumSet.of(
+            Age_Values.V6_2,
+            Age_Values.V6_3,
+            Age_Values.V7_0,
+            Age_Values.V8_0,
+            Age_Values.V9_0,
+            Age_Values.V10_0,
+            Age_Values.V11_0,
+            Age_Values.V12_0,
+            Age_Values.V13_0,
+            Age_Values.V14_0
+    /* Current version is always last */
+    );
+
+    public static EnumSet<Age_Values> getVersions() {
+        return C_UBA_AGES;
+    }
+
+    /**
+     * As a select, such as "140"
+     * @return
+     */
+    public static final String toSelect(Age_Values age) {
+        return age.getShortName().replaceAll("[\\.]", "");
+    }
+
+    /**
+     * Current version
+     *
+     * @return
+     */
+    public static Age_Values getCurrent() {
+        return C_UBA_AGES.toArray(new Age_Values[C_UBA_AGES.size()])[C_UBA_AGES.size() - 1];
+    }
+}
--- a/UnicodeJsps/src/main/webapp/bidic.jsp
+++ b/UnicodeJsps/src/main/webapp/bidic.jsp
@@ -1,7 +1,9 @@
 <html>
 
 <head>
-<%@ include file="header.jsp" %>
+<%@ include file="header.jsp"  %>
+<%@ page import="org.unicode.jsp.UBAVersion" %>
+<%@ page import="org.unicode.props.UcdPropertyValues.Age_Values" %>
 <title>Unicode Utilities: BIDI (UBA) C Reference</title>
 <script type="text/javascript">
 function detectIE() {
@@ -126,7 +128,7 @@ function setUbaInput(str) {
     UtfParameters utfParams = new UtfParameters(queryStr);
 
     String ubaPara = utfParams.getParameter("b", "2");
-    String ubaVersion = utfParams.getParameter("u", "110");
+    String ubaVersion = utfParams.getParameter("u", UBAVersion.toSelect(UBAVersion.getCurrent()));
     String ubaDetail = utfParams.getParameter("d", "2");
     boolean ubaShowVacuous = !"off".equals(utfParams.getParameter("y", "off"));
     String valInputCharSeq = utfParams.getParameter("s", "\u0645\u0627\u0631\u0652\u0643 \u2066\u0031\u2013\u0033%\u2069 mark (\u0366v.2)\u0368!");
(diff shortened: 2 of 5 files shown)


UnicodeJsps/src/main/java/org/unicode/jsp/UBAVersion.java
```diff
@@ -0,0 +1,72 @@
+package org.unicode.jsp;
+
+import org.unicode.props.UcdPropertyValues.Age_Values;
+
+/**
+ * A class to encapsulate the available C UBA versions.
```
markusicu, line 6: This looks ok, but why not define a Java enum?
You could still have custom functions for where the defaults are inconvenient.

UnicodeJsps/src/main/java/org/unicode/jsp/UBAVersion.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+public class UBAVersion {
+    static Age_Values[] C_UBA_AGES = {
```
markusicu, line 9: Please use `private` unless you really need & want the default package visibility.

UnicodeJsps/src/main/webapp/bidic.jsp
```diff
@@ -295,13 +296,13 @@ function setUbaInput(str) {
...
                 <select name="u" size="1">
-                    <option value="62" <%= (ubaVersion.equals("62") ? "selected" : "") %>>6.2</option>
-                    <option value="63" <%= (ubaVersion.equals("63") ? "selected" : "") %>>6.3</option>
-                    <option value="70" <%= (ubaVersion.equals("70") ? "selected" : "") %>>7.0</option>
-                    <option value="80" <%= (ubaVersion.equals("80") ? "selected" : "") %>>8.0</option>
-                    <option value="90" <%= (ubaVersion.equals("90") ? "selected" : "") %>>9.0</option>
-                    <option value="100" <%= (ubaVersion.equals("100") ? "selected" : "") %>>10.0</option>
-                    <option value="110" <%= (ubaVersion.equals("110") ? "selected" : "") %>>11.0</option>
+                    <% for (final UBAVersion v : UBAVersion.getVersions()) {
+                        //final UBAVersion v = UBAVersion.getVersions()[i];
```
markusicu, line 300: delete

UnicodeJsps/src/main/java/org/unicode/jsp/UBAVersion.java
srl295: I thought to make it a custom object because it references the UcdPropertyValues.Age_Values — but it could not.

UnicodeJsps/src/main/webapp/bidic.jsp
srl295: done

UnicodeJsps/src/main/java/org/unicode/jsp/UBAVersion.java
srl295: done

UnicodeJsps/src/main/java/org/unicode/jsp/UBAVersion.java
srl295: I guess I could have simply `V11_0, V12_3` etc. and then change `_` -> `.` where needed.

srl295: @markusicu Is it a valid assumption that there will be a org.unicode.props.UcdPropertyValues.Age_Values for each UBA version?

macchiati (commented):
I didn't add it; I think this was introduced because of the UBA-c page.

We should discuss, because I don't think it is of any particular value to have that pulldown or separate version.

srl295:
> I didn't add it; I think this was introduced because of the UBA-c page.
> 
> 
> 
> We should discuss, because I don't think it is of any particular value to have that pulldown or separate version.

It lets you try separate versions of the uba.

macchiati:
True, but why is that important? It hasn't changed in years.

UnicodeJsps/src/main/java/org/unicode/jsp/UBAVersion.java
Ken-Whistler: It is valid that there will be a Unicode Age corresponding to each UBA version, but not vice versa -- at least not from the point of view of the bidic code, since the reference code only covers versions starting from 6.2. (And it might conceivably skip versions -- I've just been adding each version to the reference code as I update it.)

Ken-Whistler:
@macchiati Being able to try different versions is a feature of the bidi reference (C) code implementation, and is aimed at people deep into the bidi maintenance. However, for the public jsp implementation, I agree it might be simpler to just dispense with all the choices, but make sure that the UBA version used for the call to the C code always matches the latest supported.

UnicodeJsps/src/main/java/org/unicode/jsp/UBAVersion.java
echeran: Hmm, this code is actually really close to being able to convert to a [Java enum](https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html) (I think you could get rid of a little code if you do because enums also maintain an ordering of defined values via an array than you get via `.values()`). 

FWIW, `Age_Values` is already [defined as an enum](https://github.com/unicode-org/unicodetools/blob/main/unicodetools/src/main/java/org/unicode/props/UcdPropertyValues.java#L34).

markusicu: Right. This is just for bidic.jsp, which needs
- getCurrent() (=get latest)
- iterate over all values
- for each value: get the Age_Values.getShortName()
- for each value: same, but without dots

This could be a new enum type with an Age_Values field, constructing each UBAVersion with an Age_Values.

Or, this could simply be an EnumSet<Age_Values>.of(... subset of Age_Values ...), pulling the tiny bit of custom logic into bidic.jsp. No real need for a new type.

srl295: If we need this drop down.. can you give a brief snippet of how you expect the enum to look?

markusicu: If, yes. I don't have an opinion on whether we need it :-)

I would just put something like this into bidic.jsp:
```Java
private static final EnumSet<Age_Values> UBA_VERSIONS = EnumSet.of(V6_2, V6_3, ...);
```
And then just iterate over the set, fetch the last one from iteration or from toArray()/size(), call age.getShortName(), and get the three-digit version from an inline expression.

srl295: > If, yes. I don't have an opinion on whether we need it :-)
> 
> I would just put something like this into bidic.jsp:
> 
> ```java
> private static final EnumSet<Age_Values> UBA_VERSIONS = EnumSet.of(V6_2, V6_3, ...);
> ```
> 
> And then just iterate over the set, fetch the last one from iteration or from toArray()/size(), call age.getShortName(), and get the three-digit version from an inline expression.

oh, an enumset of `Age_Values` … not a new enum.  I was struggling to see how defining the identity of the new enum's values would be a benefit.

markusicu: > oh, an enumset of `Age_Values` … not a new enum. I was struggling to see how defining the identity of the new enum's values would be a benefit.

Well, a new enum is a possibility, and is what I initially thought of. What you built is basically a poor man's enum. By turning it into a real enum, you could remove some boilerplate, such as (as Elango wrote) using `.values()`. But there is already an enum, and what you need is a subset of its values, so an EnumSet seems good enough. And if it's used only in one place, we need not define special API.

srl295: I would rather not put more code into the JSP. JSPs are excruciatingly error prone and unmaintainable. But I can put the relevant code into statics on UBAVersion, and see what you think of that.

srl295: OK @echeran @macchiati @markusicu  please take a look… it's now an enumset.

markusicu (approved):
lgtm tnx

PR #203, opened 2022-02-26
UCD 15 files as of 2022-feb-25 + invariant fixes

markusicu:
except files with various regressions and no data changes
- PropList.txt
- Scripts.txt
- ScriptExtensions.txt

## Changes to fix TestUnicodeInvariants failures

The 8 new Egyptian hieroglyph controls 13439..13440 are excluded from Default_Ignorable_Code_Point, just like the older Egyptian hieroglyph controls.

Some of them are effectively opening and closing format controls, as some of the older ones; thus added to special sets in the test data file.

From Ken:
> the rationale is that the format controls in Egyptian function mostly to stack up quadrats. The Egyptologists are used to reading that syntax (at least for simple quadrats) in the Manual de Codage style with colons and asterisks, etc. And they are better off defaulting to showing visible controls, somewhat reminiscent of the IDS displays for CJK ideographs, if displayed on a rendering system that doesn't support the full quadrat formation, rather than having the format controls vanish on display like ZWJ, etc.

----
The test expects
```
11F43..11F4F   # [13]  KAWI DANDA..KAWI PUNCTUATION CLOSING SPIRAL
```
to not be lb=SA because they are punctuation.
Ken wrote:
> Given the precedent of 104A..104B, the Myanmar section signs (historical equivalent to the dandas), I think the simplest fix here is to adjust the Line_Break values for the new Kawi punctuation to lb=BA, instead of lb=SA.
>
> L2/20-284R was silent wrt Line_Break properties. Some of this Kawi punctuation might be lb=BB, but starting with lb=BA for all of it, instead of lb=SA should eliminate the invariant test issue, and then we can go back and ask more about which of this stuff actually would make more sense as lb=BB, or if it is o.k. to just break *after* any of these.

Thus, changing LineBreak.txt with 11F43..11F4F --> lb=BA.


Manishearth (approved):

macchiati (approved):

macchiati (approved):

Manishearth (approved):

Ken-Whistler (approved):
No objections.

PR #207, opened 2022-03-08
Add tab check

josh-hadley:
- add a simple Python script that checks all .java files in the project for tab characters. The script prints the full path of each file found containing tabs and fails if any are found. The script can be run independently of CI.
- minor modifications to the Python workflow to link up the tab check script

### DIFF
--- a/py/tabcheck.py
+++ b/py/tabcheck.py
@@ -30,7 +30,7 @@ def main() -> bool:
         return False
 
     error_count = 0
-    for p in Path(root).rglob("*.java"):
+    for p in root.rglob("*.java"):
         with open(p, 'r') as f:
             s = f.read()
             if '\t' in s:
--- a/py/tabcheck.py
+++ b/py/tabcheck.py
@@ -1,0 +1,46 @@
+#!/usr/bin/python3 -B
+# © 2021 and later: Unicode, Inc. and others.
+# License & terms of use: http://www.unicode.org/copyright.html
+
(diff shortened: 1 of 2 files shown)


py/tabcheck.py
```diff
@@ -0,0 +1,46 @@
...
+                    help="Force exit with status code 0 even if there are errors.")
+
+    args = ap.parse_args()
+    root = Path(args.ROOT_DIR)
+
+    if not root.exists():
+        print(f'{root.absolute()}: directory not found; aborting.')
+        return False
+
+    error_count = 0
```
markusicu, line 32: Why do you count the errors when you only return a boolean?

markusicu (approved):
One minor comment.

Otherwise lgtm -- please **squash-and-merge** when ready.

py/tabcheck.py
josh-hadley: Ah, good point. I was originally going to return the error_count (for a non-zero exit code) but changed my mind later on without modifying that bit. I can change that to "errors_found" and treat it as a boolean if you prefer. I don't think it makes much difference.

markusicu: I would like that. I can quickly re-approve.

josh-hadley: done in https://github.com/unicode-org/unicodetools/pull/207/commits/af7a0e13c7d04b1a16157936971ab2348b1e047a

markusicu (approved):
lgtm
please **squash-and-merge**

PR #205, opened 2022-03-07
Add more tools to CI and docs

echeran:
Contributes to #187

PR #213, opened 2022-03-22
update UCA.Main instructions

markusicu:
including recommended command-line options


echeran (approved):
LGTM

What I have in draft for #205 for UCA.Main runs 2 separate commands, one that uses `writeValidityLog`, the other that uses `ICU`.  The first passes, the second one doesn't and needs a followup issue.  I suppose once that future followup issue is fixed, the 2 commands would be combined according to these updated instructions.

Issue #186, opened 2021-12-16
auto-test documented test programs


markusicu:
In a CI workflow, run unit tests for documented tests.
- Look through the https://github.com/unicode-org/unicodetools/tree/main/docs pages (excluding "help")
- For each test documented there,
  - Add it to the list below
  - If it is not already using JUnit, then convert it to that
  - Add it to CI

List of tests:
- [ ] TestUnicodeInvariants, see https://github.com/unicode-org/unicodetools/blob/main/docs/build.md#invariant-checking
- [ ] TODO: look for more

markusicu:
Josh to find test programs, Steven to JUnit-ify them.

markusicu:
Example: https://github.com/unicode-org/unicodetools/blob/main/docs/idna.md
--> add an entry in the issue description/checklist above for TestIdna.java

Use judgment; some classes that should be unit tests may not be called `*Test*.java`.

markusicu:
Hi, just checking -- any progress? I think "Josh to find test programs" comes first...?

josh-hadley:
Thanks for the reminder -- I have been sidetracked but hoping to make some progress on this and #187 this week

josh-hadley:
@markusicu @srl295 I went through and added what appeared to be tests ("checks" in some cases).

markusicu:
Thanks @josh-hadley ! Over to @srl295 ?
- TestUnicodeInvariants should be shovel-ready, and it's important because it frequently shows us needing to adjust properties.
  - Ideal would be a parameterized test, I think, but even just initially having it fail (boolean) if there is anything inconsistent would be useful.
- TestIdna is hopefully easy
- CheckXmlProperties is definitely a test, not a tool.
  - As a test, it would be useful to JUnit-ify it.
  - Issue: The ucdxml files are usually updated only rarely, maybe only once or twice per release. This means that this test would usually fail, so we can't put it into Continuous Integration.
  - Issue: The docs suggest that the test is failing already, and that one is to ignore those failures. In order to make it pass, it sounds like we would need to improve the XML file parsing.
  - Issue: Ownership of the ucdxml files & spec is up for grabs. Future uncertain.
- CheckProperties looks more like a tool than a test
  - Most of its actions print useful or interesting information, or export properties data in another format (e.g., JSON), etc.
  - "COMPARE" shows the differences between the latest data and the previous Unicode version
  - "ICU" shows the differences between the current data and the ICU data
[...]

Issue #66, opened 2021-05-29
there can be only one UnicodeProperty.java


markusicu:
Background: ucd-dev email thread “[Unicode Tools vs. UnicodeProperty.java](https://groups.google.com/g/ucd-dev/c/H60xj1kTfCU)”

We have two versions of org/unicode/cldr/util/props/UnicodeProperty.java, one in CLDR and one in the Unicode Tools.
Until 1.5 years ago they were identical, but they have been diverging and are now no longer even interface-compatible. The CLDR version is the newer one.

For now, I am adding a note to the setup instructions to move cldr-code above unicodetools in the build path, but of course that's evil and brittle.

We had agreed to move UnicodeProperty from CLDR into the Unicode Tools, but other CLDR code depends on this class, and I am not sure if I have time to get through that.
Maybe I just rename the unused Unicode Tools version at first.

There is also a third version of UnicodeProperty.java, in the UnicodeJsps code, but thankfully it is in a different package. Still, this is confusing and error-prone; we should either merge the classes or give them distinct names.
- https://github.com/unicode-org/cldr/blob/master/tools/cldr-code/src/main/java/org/unicode/cldr/util/props/UnicodeProperty.java
- https://github.com/unicode-org/unicodetools/blob/main/unicodetools/org/unicode/cldr/util/props/UnicodeProperty.java
- https://github.com/unicode-org/unicodetools/blob/main/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeProperty.java

srl295:
- See https://unicode-org.atlassian.net/browse/CLDR-9787 for the CLDR side of duplicate classes between ICU/CLDR (similar problem), and https://unicode-org.atlassian.net/browse/ICU-21757 for what may be a needed part of the solution
- Once #64 is done, we should change UnicodeJsps to be dependent on unicodetools. This will allow the classes to be present only once and shared between those two.

macchiati:
I tried at one point to pull UnicodeProperty out from CLDR, but it is
not simple. There are classes that depend on classes and so on, soit was
too hard to figure out what the tangled set of dependencies were, and where
I could cut the cord to allow me to extract a set of classes and tie off
the loose ends.

Mark

srl295:
Josh's analysis: https://docs.google.com/spreadsheets/d/1sAIFut5QL30EXwMK_ZM4yKzWJge2gatLA0yfl_zxtSY/edit?usp=sharing

srl295:
[...]

PR #214, opened 2022-03-24
move UnicodeProperty so it doesn't shadow CLDRs

srl295:
For: #66 

while working on #186 I found some tools wouldn't even work for me due to the shadowed classes issue discussed in #66

This untangles things a bit… while making other things a little more tangled.

Draft because I'm not sure about all of this, but it does pass tests, and GenerateIdna actually runs again instead of crashing.


unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
```diff
@@ -4,7 +4,7 @@
  * others. All Rights Reserved.                                                *
  *******************************************************************************
  */
-package org.unicode.cldr.util.props;
+package org.unicode.props;
```
markusicu, line 7: Looking at the description of issue #66, we have been using the CLDR version of UnicodeProperty which has apparently changed in significant ways. (I don't remember what diffs I looked at almost a year ago...)

From the issue:
> The CLDR version is the newer one. ...
> For now, I am adding a note to the setup instructions to move cldr-code above unicodetools in the build path, ...

Now that you change the unicodetools code to explicitly use this org.unicode.props.UnicodeProperty, shouldn't this be a copy of the CLDR version that we have been using, rather than the unicodetools/org.unicode.cldr version that was shadowed, in order to keep the behavior intact?

srl295: I'm not sure we've been using the CLDR version consistently. I think it may be hit or miss depending on how it is packaged or loaded. Some code worked, some didn't.

but… that's an interesting tack.

srl295: I assumed the one in unicodetools was 'needed' somehow.

macchiati:
I think the very best approach would be to remove UnicodeProperty from CLDR
if possible. There are however lots of direct and indirect dependencies.
Second best would be to merge anything useful from the unicodetools one
into the CLDR one, and delete the unicodetools one.

markusicu:
Mark:
> I think the very best approach would be to remove UnicodeProperty from CLDR if possible. There are however lots of direct and indirect dependencies. Second best would be to merge anything useful from the unicodetools one into the CLDR one, and delete the unicodetools one.

Yes, all as discussed in the top description of issue #66.
And, as it says there, the unicodetools code has been running with the CLDR repo version which is newer and has more recent changes.

I agree with moving the unicodetools version to a different package, to avoid the shadowing, but I strongly suspect that the version there needs to be mostly a copy of the CLDR version, in order to preserve behavior.

Steven:
> I'm not sure we've been using the CLDR version consistently. I think it may be hit or miss depending on how it is packaged or loaded.

My comment from last year says that the CLDR version has a different signature. Either the Unicode Tools rely on that now, or they don't call anything where the API has changed.

From #66:
> Until 1.5 years ago they were identical, but they have been diverging and are now no longer even interface-compatible. The CLDR version is the newer one.

macchiati (commented):
Looks good to me. A couple of questions.

Were the moved files like BagFormatter just copied, or were there more substantive changes?
Why was "final" removed from a number of declarations in unicodetools/src/main/java/org/unicode/jsp/UnicodeRegex.java? Was that because of merging with another version?

srl295:
> Looks good to me. A couple of questions.
> 
> 
> 
> Were the moved files like BagFormatter just copied, or were there more substantive changes?

The ones from CLDR that are completely new were not changed. I can enumerate. 


> 
> Why was "final" removed from a number of declarations in unicodetools/src/main/java/org/unicode/jsp/UnicodeRegex.java? Was that because of merging with another version?

Merging between the unicodetools and the jsp version.

macchiati (approved):

unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
markusicu: Discussed in live meeting. Try to use/copy the CLDR version.

markusicu: Resolved, I think, after various experiments:
The CLDR code needs to call the CLDR version of UnicodeProperty, but the Unicode Tools code that uses UnicodeProperty directly needs to use its own version.

markusicu (approved):
lgtm tnx!

PR #216, opened 2022-04-05
delete UnicodeProperty.java

srl295:
For: #66


markusicu:
From the build failures, clearly CLDR code needs to use CLDR UnicodeProperty, but otherwise Unicode Tools code needs to use its version of UnicodeProperty. Probably need to drop this PR in favor of PR #214?

markusicu:
I suggest closing this PR.

PR #219, opened 2022-04-08
switch from ICU ObjectMatcher to Java Predicate

markusicu:
and then clean up most warnings in UnicodeProperty

For issue #66: This makes unicodetools UnicodeProperty more like CLDR UnicodeProperty. The CLDR version switched to Java Predicate in https://github.com/unicode-org/cldr/pull/460. The warnings cleanup here goes farther than the earlier one in CLDR.

Best to review one commit at a time.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/text/UCD/GenerateConfusables.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/GenerateConfusables.java
@@ -32,6 +32,7 @@ import java.util.Map.Entry;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
+import java.util.function.Predicate;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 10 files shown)


srl295 (approved):

unicodetools/src/main/java/org/unicode/text/UCD/GenerateConfusables.java
```diff
@@ -2493,4 +2493,20 @@ public String transform(Linkage<String, String> source) {
...
+    /**
+     * Retain matching items
+     * @param <T>
+     * @param <U>
+     * @param c
+     * @param f
+     * @return
+     */
+    static <T, U extends Collection<T>> U retainAll(U c, Predicate<T> f) {
+        for (Iterator<T> it = c.iterator(); it.hasNext();) {
```
echeran, line 2506: fyi: the Java 8 way of [filtering a collection](https://www.baeldung.com/java-stream-filter-lambda#1-filtering-collections) might look like:

```
return c.stream().filter(f).collect(Collectors.toList());
```
Just an FYI, since this is probably focused on the higher-level goal of refactoring code across projects.

echeran (approved):
LGTM

unicodetools/src/main/java/org/unicode/text/UCD/GenerateConfusables.java
markusicu: Thanks. My goal was to reduce the yellow warning markers in Eclipse, not to generally modernize the code. There is a ton of code in this project using very old-fashioned Java mechanisms.

Issue #218, opened 2022-04-06
make Unicode Tools work with CLDR 41


markusicu:
The Unicode Tools do not work with the latest CLDR snapshot ([2022-apr-04](https://github.com/unicode-org/cldr/packages/521051)). I tried to pick it up like this:
```patch
diff --git a/pom.xml b/pom.xml
index d5dfbb3a..d60db5e8 100644
--- a/pom.xml
+++ b/pom.xml
@@ -31,7 +31,7 @@
                <!--
                         For CLDR versions, see https://github.com/orgs/unicode-org/packages?repo_name=cldr
                  -->
-               <cldr.version>0.0.0-SNAPSHOT-e995b07a7e</cldr.version>
+               <cldr.version>0.0.0-SNAPSHOT-1e28c0fbb4</cldr.version>
 
 
                <!-- these two set the JDK version for source and target -->
```

The Tools depend on the Xerces XML parser, which they used to get indirectly via CLDR. CLDR removed that dependency due to a Java 11 conflict: https://github.com/unicode-org/cldr/pull/1859

I tried copying the old Xerces Maven dependencies into the Unicode Tools pom.xml files (top + unicodetools) but then got the errors for the Java 11 conflict. I suspect that the Unicode Tools also need to switch to the built-in XML parser. Maybe we need to look at changes from more of the https://unicode-org.atlassian.net/browse/CLDR-14311 pull requests and make similar changes in the Tools?

This is blocking any kind of CLDR-vs-Unicode-Tools refactoring; in my case, I am attempting a refactoring of CLDR class UnicodeProperty and trying that out with Unicode Tools.

@srl295 could you please take a look?

srl295:
Can you try removing xerces?

markusicu:
> Can you try removing xerces?

I don't know what you mean? Remove it from where?

I have an otherwise unmodified Unicode Tools workspace where I tried to upgrade the CLDR dependency to the latest snapshot. That failed to build, complaining that it couldn't find any org.apache stuff including Xerces. Then I tried to add the Xerces dependency that CLDR used to have to Unicode Tools which used to just inherit it. Then I got the conflict errors.

That's what I did, that's all I know :-}

markusicu:
It looks like this is blocking some work for @echeran now...

srl295:
@markusicu  will take a look

srl295:
> > Can you try removing xerces?
> 
> I don't know what you mean? Remove it from where?

[...]

PR #224, opened 2022-04-22
deps: bump cldr to v41, drop xerces

srl295:
For: #218

### DIFF
--- a/UnicodeJsps/pom.xml
+++ b/UnicodeJsps/pom.xml
@@ -69,18 +69,6 @@
 			<version>29.0-jre</version>
 		</dependency>
 
-		<dependency>
-			<groupId>xerces</groupId>
-			<artifactId>xercesImpl</artifactId>
-			<version>2.12.2</version>
-		</dependency>
-		<dependency>
-			<groupId>xml-apis</groupId>
-			<artifactId>xml-apis</artifactId>
-			<version>1.4.01</version>
-			<scope>compile</scope>
-		</dependency>
-
 		<!-- test -->
 		<dependency>
 			<groupId>org.junit.jupiter</groupId>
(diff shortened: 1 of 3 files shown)


srl295:
sorry for the delay, see if this works.

echeran:
Even though this looks like the right change, it didn't make a difference...I'm trying to get Eclipse to import the project without errors, but it still complains about a module conflict for the `org.xml.sax.*` import statements.  (It says the org.xml.sax package is accessible from more than one module: <unnamed>, java.xml).

That error sounds like an error that you may have encountered in CLDR and solved through some of these changes. If so, was there anything else (ex: any other configs somewhere) to solve this problem?

markusicu (approved):
changes lgtm

I haven't tried to patch these in. Given that these changes are equivalent to those in CLDR, maybe just merge, and we update Eclipse and see?

markusicu:
... and thanks for jumping on this!

markusicu:
Looks like the core problem is that the imports fail with the well-known "The package org.xml.sax is accessible from more than one module: `<unnamed>`, java.xml"

markusicu:
unicodetools > Properties > Java Build Path > Classpath / Maven Dependencies still lists xercesImpl

I can't edit it, but this item points to ~/.m2/repository/xerces/xercesImpl/2.12.0/ on my machine.

I shouldn't have to blow that folder away, right? The Maven project should just stop referring to it?
(I am afraid some other Java project might still need it; ICU?)

srl295:
@markusicu  I'm testing on Eclipse now instead of just maven.

srl295:
@markusicu  I found the problem. This is v41, yes, but the JDK11 and CLDR fixes for eclipse landed in v42.

I'm testing a snapshot of  cldr `v42` now https://github.com/unicode-org/cldr/releases/tag/snapshot%2F2022-04-25

PR #226, opened 2022-04-26
Unicode 15 data apr26

markusicu:
Post UTC 171 data file updates from 2022-apr-26, with code changes and generated files.

These updates take care of UCD actions 171-A058, 171-A099, 171-A100, and almost all of the Unihan-related actions.

Note: Unihan extension C now ends one code point higher, at U+2B739.


Ken-Whistler (approved):
The Kawi changes in LineBreak.txt show some regression, but that will be fixed in the next LineBreak.txt commit. Everything else looks as expected.

markusicu:
> The Kawi changes in LineBreak.txt show some regression,

Ok, so 11F43..11F4F changing from BA to SA apparently slipped in somehow. This is probably also why TestUnicodeInvariants fails for more than one Kawi character.

> but that will be fixed in the next LineBreak.txt commit.

Right.

> Everything else looks as expected.

Thanks! Merging...

PR #181, opened 2021-12-11
fix TestUnicodeInvariants failures

markusicu:
- U+0CF3 properties fixes, see https://www.unicode.org/L2/L2021/21130-script-adhoc-rept.pdf
- Kawi punctuation: lb=BA
- add U+11F41 KAWI SIGN KILLER to the $SAScriptExceptions list
- remove Kawi postbase (gc=Mc, InPC=Right) characters from GCB=SpacingMark
- other routine test case adjustments

PR #229, opened 2022-04-27
include Kawi postbase (gc=Mc, InPC=Right) characters in GCB=SpacingMark

markusicu:
= partial reverse of pull request #181

Fixes UTC action item [171-A69](https://www.unicode.org/cgi-bin/GetL2Ref.pl?171-A69).
After bug reports, and discussion in UTC 171, it was decided for Kawi line breaking to follow Western style for now, and try to default to orthographic syllables in the future, rather than using East Asian style (with the whole script in lb=SA).


Ken-Whistler (approved):
o.k.

Issue #222, opened 2022-04-11
Invalid connection between lb=SA and exclusion from GCB=SpacingMark


NorbertLindenberg:
Lines 569-570 of unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt say “Ken: The postbase gc=Mc, InPC=Right characters (for these scripts with lb=SA) should be excluded from the set GCB=SpacingMark, and instead should fall through to GCB=XX.” Line 573 makes use of this connection by using lb=SA in the calculation of $PostBaseSpacingMarks_All.

I believe the connection between lb=SA and exclusion from GCB=SpacingMark made here is invalid. One is about a specific requirement for line breaking, the other about a desire to position cursors inside clusters (see [L2/11-114](https://www.unicode.org/cgi-bin/GetDocumentLink?L2/11-114), which apparently prompted the list of exclusions from SpacingMark). That both currently apply to subsets of the South East Asian scripts is irrelevant. That they are not the same is evidenced, for example, by Khmer characters being included in SpacingMark while having (and needing) lb=SA. I recommend removing that connection, and instead using an explicit list of scripts whose spacing marks should be excluded.

The comment in line 567 "Post-base spacing combining marks of most SE Asian scripts are explicitly excluded from GCB=SpacingMark" is also wrong. Most of the 26 south east Asian scripts in Unicode 15 are not affected by this.

The initially quoted comment has also been added to unicodetools/src/main/java/org/unicode/text/UCD/ToolUnicodePropertySource.java, and should be corrected there at the same time.

markusicu:
FYI

This is based on several email discussions, mostly between Norbert, KenW, Debbie, and myself, which started with an invariant-test failure involving some Kawi characters.

Ken wrote in 2021-dec:
> In this case I think the invariant test is correct. ...
> The *postbase* gc=Mc, InPC=Right characters (for these scripts with lb=SA) should be excluded from the set GCB=SpacingMark, and instead should fall through to GCB=XX. ...
> For Kawi, that should leave just 11F3E..11F3F as GCB=SpacingMark. Those are the *prebase* vowels. Cf. the treatment for U+11726 AHOM VOWEL SIGN E.

So I changed the data and the spec.
- https://github.com/unicode-org/unicodetools/pull/181
- https://github.com/unicode-org/unicode-reports/pull/30

[...]

Issue #221, opened 2022-04-11
Wrong description of purpose of Line_Break=SA


NorbertLindenberg:
Lines 465-467 of unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt have a comment starting with:

> For the scripts (mostly of SE Asia) which have been identified as normally not using spaces in their typography and which have therefore had their letters all given Line_Break=SA (complex context dependent),

This description of Line_Break=SA is incorrect. The use of spaces isn't indicative of the need for complex context dependent analysis. Such analysis is needed when scripts allow line breaks only at word boundaries, but do not mark word boundaries in any way. Thai uses spaces to separate phrases, but not for words, doesn't mark word boundaries in any other way, but allows line breaks only at word boundaries, so it needs the complex context dependent analysis indicated by Line_Break=SA. Balinese, on the other, traditionally doesn't use spaces, but allows line breaks at the boundaries of orthographic syllables, which are easy to identify.

Issue #220, opened 2022-04-11
Wrong Unicode version number for Kawi


NorbertLindenberg:
Line 491 of unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt prefixes some information on Kawi with "14.0", which is likely meant to be a Unicode version number. Kawi was not in Unicode 14.0; it is being added for Unicode 15.0.

Issue #14, opened 2020-07-23
Emoji chart fixes


macchiati:
Drop the sidebar on the index page

---

Fix the links that go to unicode.org/emoji to stay on the tech site

---

The new chart files still have the same 3 non-fatal errors that have been there for a while:
Line 4, Column 15: "ASYNC" is not a member of a group specified for any attribute
```
<script async src='https://www.googletagmanager.com/gtag/js?id=UA-19876713-1'><…
Error Line 4, Column 78: required attribute "TYPE" not specified
…async src='https://www.googletagmanager.com/gtag/js?id=UA-19876713-1'></script>
Error Line 5, Column 8: required attribute "TYPE" not specified
<script>
    ... Typical values for type are type="text/css" for <style> and type="text/javascript" for <script>.

```

----

The following HTML file listing proposals has a few errors, some cascading:
https://www.unicode.org/emoji/flag-charts-13.0-2020-07-23/emoji-proposals.html

E.g, line 31 is:

`<p><b>Caveats:</b> the data is draft, and needs further improvement: suggestions for corrections and other improvements can be submitted at <a target='emoji-proposal-corrections' href='https://docs.google.com/forms/d/e/1FAIpQLSf5RtIkkBP_eUmRFSZfBGm9W6vj2n31_qUzWcv9FXdtyb81kQ/viewform?usp=pp_url&entry.565091389=Add&entry.1700236179=%F0%9F%91%BD&entry.1390707358=L2/99-999'>Emoji Proposal Corrections</a>.</p><p>While these charts use a particular version of the <a target='emoji-data' href='https://unicode.org/Public/emoji/13.0/'>Unicode Emoji data files</a>, the images and format may be updated at any time. For any production usage, consult those data files.  For information about the contents of each column, such as the <b>CLDR Short Name</b>, click on the column header. For further information, see <a target='text' href='index.html'>Index &amp; Help</a>.</p>
`
The source of the validation complaints is probably un-escaped ampersands in URLs, as in the "&entry" in the above line. (I think these don't usually cause browser failure, the validator just doesn't like them.)

----

Emoji-released file https://www.unicode.org/emoji/flag-charts-13.0-2020-07-23/emoji-released.html has an closing </p>, in part of the template:

  Line 25, Column 109: end tag for element "P" which is not open.

I think you want a closing </p> on line 24, just before the phrase "The characters were based on proposals"...

markusicu:
These things appear to be fixed or moot.

PR #230, opened 2022-04-27
fix invariants of lb=SA and of GCB=SpacingMark exceptions

markusicu:
Fixes issue #220
Fixes issue #221
Fixes issue #222

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -602,7 +602,7 @@ Unicode 15 TODO: See above; commit new input data, run tools, review output, cop
         # Total code points: 12089
         **** END Error Info ****
         ```
-6.  The input file is [unicodetools/org/unicode/text/UCD/UnicodeInvariantTest.txt](https://github.com/unicode-org/unicodetools/blob/main/unicodetools/org/unicode/text/UCD/UnicodeInvariantTest.txt).
+6.  The input file is [unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt](https://github.com/unicode-org/unicodetools/blob/main/unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt).
     1.  Some failures are expected for a new Unicode version, or new RTL blocks, etc. Adjust the input file as necessary.
     1.  For other failures, adjust the character properties.
 
--- a/unicodetools/src/main/java/org/unicode/text/UCD/ToolUnicodePropertySource.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/ToolUnicodePropertySource.java
@@ -810,10 +810,12 @@ extend -and not GCB = Virama
             //            unicodeMap.put(0, "Prepend");
             //            unicodeMap.setErrorOnReset(true);
 
-            // https://www.unicode.org/reports/tr29/#SpacingMark
-            // Ken: The *postbase* gc=Mc, InPC=Right characters (for these scripts with lb=SA)
-            // should be excluded from the set GCB=SpacingMark, and instead should fall through to GCB=XX.
+            // Post-base spacing combining marks (gc=Mc, InPC=Right) of certain SE Asian scripts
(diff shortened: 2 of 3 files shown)


markusicu:
@NorbertLindenberg FYI (GitHub does not let me add you to the list of reviewers)

Ken-Whistler (commented):
Generally this looks good, but I am not competent to comment on the details of the construction of the invariant tests.

markusicu:
@Ken-Whistler most of this pull request is in changing explanatory text. Would you mind approving it based on that, assuming the text looks ok?

NorbertLindenberg:
[ToolUnicodePropertySource.java](https://github.com/unicode-org/unicodetools/pull/230#diff-29e6d97f62bf888d7f8351b49a76b6e27835fe56a911f9702d52e849bc95aabe):

* Line 834 still lists Kawi characters.
* Is it still necessary to list the New Tai Lue characters?
* Is the version test for Ahom in line 836 still needed?

[UnicodeInvariantTest.txt](https://github.com/unicode-org/unicodetools/pull/230#diff-c2f2f053bbaa14f43053e3fb81a8afd67add46a4be1377d3e411f0500b40dabf)

* There's still too much emphasis on spaces. In Asia at least, the presence or absence of spaces is not a good predictor for line-breaking behavior. Tibetan uses spaces, but doesn't break lines at them; it has other characters that provide line breaking opportunities. Grantha nowadays uses spaces, but also breaks at orthographic syllable boundaries. Chinese doesn't use spaces, but allows breaks almost anywhere. Thai uses spaces for phrases, but breaks at word boundaries, which are not marked in any way.
* Grantha is not a south east Asian script.
* It doesn't seem necessary to discuss what scripts do that don't fall into the SA bucket.
* I'd replace lines 466-481 with:

> A few southeast Asian scripts allow line breaks only at word boundaries, but do not mark word boundaries in any way, so that the determination of line break opportunities requires language dependent text analysis. In terms of UAX #14, these scripts need the “South East Asian” style of context analysis for line breaking, and their characters are given lb=SA (Line_Break=SA: complex context dependent) as an escape out of the rule-based algorithm in UAX #14.
>
> For these scripts, test the following invariant:

Ken-Whistler:
Yeah, see what Norbert wrote. Again, I say, I'm not really competent to judge the invariant tests here, or even the description of what they are trying to do. If you need an approval to commit this, ask somebody else.

markusicu:
> ToolUnicodePropertySource.java:
> 
> * Line 834 still lists Kawi characters.

I removed them in a separate PR that is merged already: https://github.com/unicode-org/unicodetools/pull/229

> * Is it still necessary to list the New Tai Lue characters?

Unsure; I already added a note about them.
This code looks like it wants to compute different results for different Unicode versions, and I don't want to change anything that might change results for older versions. It would be tedious to test.
I also don't want to make unrelated changes.

> * Is the version test for Ahom in line 836 still needed?

Presumably yes. The data changed for Unicode 14, so we need the version test in order to generate different data for 13 and before.

> UnicodeInvariantTest.txt
> 
> * There's still too much emphasis on spaces. In Asia at least, the presence or absence of spaces is not a good predictor for line-breaking behavior. Tibetan uses spaces, but doesn't break lines at them; it has other characters that provide line breaking opportunities. Grantha nowadays uses spaces, but also breaks at orthographic syllable boundaries. Chinese doesn't use spaces, but allows breaks almost anywhere. Thai uses spaces for phrases, but breaks at word boundaries, which are not marked in any way.
> * Grantha is not a south east Asian script.
> * It doesn't seem necessary to discuss what scripts do that don't fall into the SA bucket.
> * I'd replace lines 466-481 with: [...]

sgtm, thanks, done.

@NorbertLindenberg PTAL

echeran (approved):
RSLGTM

NorbertLindenberg:
Looks good to me.

Manishearth (approved):

PR #231, opened 2022-04-27
Unihan 15 make variants data more consistent

markusicu:
Unihan_Variants.txt updates from @Tsengtsz.
No changes in generated UCD files.


Ken-Whistler (approved):
Look as intended.

PR #238, opened 2022-05-05
add idna2008derived files

markusicu:
Creates a folder with the same name and relative placement (inside "idna" but *next to* the version-numbered Unicode idna data file folders) as in https://www.unicode.org/Public/idna/ -- see https://www.unicode.org/Public/idna/idna2008derived/


Ken-Whistler (approved):
Looks as intended.

PR #243, opened 2022-05-10
Update security-related data files

echeran:
These files are the result of running GenerateConfusables according to the [instructions](https://github.com/unicode-org/unicodetools/blob/main/docs/security.md#generating).


markusicu (approved):
lgtm tnx

PR #247, opened 2022-05-11
TestUnicodeInvariants: add tests for TestCodeInvariants

srl295:
adds tests for these classes
- TestUnicodeInvariants
- TestCodeInvariants

#186


srl295:
- [ ] fix path issue

```
java.io.FileNotFoundException: /home/runner/work/unicodetools/unicodetools/unicodetools/nullorg/unicode/text/UCD/UnicodeInvariantTest.txt (No such file or directory)
[2784](https://github.com/unicode-org/unicodetools/runs/6393672630?check_suite_focus=true#step:8:2784)
```

also @markusicu in the best case there are still 3 failures. Make it a warning for now?

markusicu:
> * [ ]  fix path issue

The file is there, and the test runs fine in my workspace. Is the test run missing a variable definition for the UNICODETOOLS_REPO_DIR ?

https://github.com/unicode-org/unicodetools/blob/main/unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt

> also @markusicu in the best case there are still 3 failures. Make it a warning for now?

That's what we get for not running it every time...

I just ran it locally and got one test failure. I will send an email.
```
\p{lb=EB} = \p{Emoji_Modifier_Base}
**** START Test Failure 1 ****
## Expected empty, got: 2	[\U0001FAF7\U0001FAF8]
## In	\p{Emoji_Modifier_Base}
## But Not In	\p{lb=EB} 
1FAF7..1FAF8   #  [2] (�..�)  LEFTWARDS PUSHING HAND..RIGHTWARDS PUSHING HAND
**** END Test Failure 1 ****
```

markusicu:
> > also @markusicu in the best case there are still 3 failures. Make it a warning for now?
> 
> That's what we get for not running it every time...
> 
> I just ran it locally and got one test failure. I will send an email.

TestUnicodeInvariants runs cleanly after PR #249. Please rebase.

srl295:
> > * [ ]   fix path issue
> 
> The file is there, and the test runs fine in my workspace. 

mine also. will investigate

srl295:
@markusicu couldn't sort it out. Changed to use classpath loader.

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -113,7 +116,31 @@ enum Expected {empty, not_empty, irrelevant};
 
     private static PrintWriter out;
 
-    public static void testInvariants(String outputFile, boolean doRange) throws IOException {
+    /**
+     * Fetch a reader for our input data.
+     * @param inputFile if null, read DEFAULT_FILE from classpath
```
markusicu, line 121: Why would it be null? Line 80 calls testInvariants() with DEFAULT_FILE, or with a String from a command-line argument, if there is one.

unicodetools/src/test/java/org/unicode/text/UCD/TestTestCodeInvariants.java
```diff
@@ -0,0 +1,19 @@
+package org.unicode.text.UCD;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Test;
+
+public class TestTestCodeInvariants {
```
markusicu, line 7: These two TestTest... classes seem awkward.
I see in the PR description "Does not refactor code into the test area, yet" -- is that the next step for this PR? Moving the regular files into the test path and changing them directly, to JUnit tests, replacing these files?

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
srl295: I pass null from TestTestUnicodeInvariants to mean default

unicodetools/src/test/java/org/unicode/text/UCD/TestTestCodeInvariants.java
srl295: @markusicu it's because there are other functions in the original class.  Is it called from elsewhere, is it run directly with a parameter? If I remove the original class from `src/main/java` then it can't be called as easily.

If this looks right I can merge this and then reduce the test classes in a future PR.

unicodetools/src/test/java/org/unicode/text/UCD/TestCodeInvariants.java
```diff
@@ -41,12 +44,24 @@ public static void main(String[] args) {
         testGcbInDecompositions(true);
     }
```
markusicu, line 45: The main() function should go away in the JUnit test, right?

unicodetools/src/test/java/org/unicode/text/UCD/TestCodeInvariants.java
```diff
@@ -41,12 +44,24 @@ public static void main(String[] args) {
         testGcbInDecompositions(true);
```
markusicu, line 44: I would remove the `showAllNfds` parameter from this function and change its code as if it had been false (or replace `showAllNfds` with `VERBOSE`). It seems unnecessary to do verbose printing for thousands of decompositions that pass the test.

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
markusicu: How about passing TestUnicodeInvariants.DEFAULT_FILE from the TestTest... code?

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -77,12 +78,8 @@ public static void main(String[] args) throws IOException {
...
 
         testInvariants(file, doRange);
-        if (TestCodeInvariants.testScriptExtensions() < 0) {
-            System.out.println("Invariant test for Script_Extensions failed!");
-        }
-        if (TestCodeInvariants.testGcbInDecompositions(false) < 0) {
-            System.out.println("Invariant test for GCB in canonical decompositions failed!");
-        }
+
+        System.out.println("Note: TestCodeInvariants is now a separate JUnit test");
```
markusicu, line 82: optional: seems unnecessary

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -113,7 +110,31 @@ enum Expected {empty, not_empty, irrelevant};
...
-    public static void testInvariants(String outputFile, boolean doRange) throws IOException {
+    /**
+     * Fetch a reader for our input data.
+     * @param inputFile if null, read DEFAULT_FILE from classpath
+     * @return BufferedReader
+     * @throws IOException
+     */
+    private static BufferedReader getInputReader(String inputFile) throws IOException {
+        if (inputFile != null) {
+            return FileUtilities.openUTF8Reader(Settings.SRC_UCD_DIR, inputFile);
```
markusicu, line 121: FYI  This should work in the build bots. For example, https://github.com/unicode-org/unicodetools/blob/main/.github/workflows/build-jsp.yml#L52 does set `-DUNICODETOOLS_REPO_DIR=$(pwd)`, and Settings.SRC_UCD_DIR is relative to that. We read all source data files relative to that path.

unicodetools/src/test/java/org/unicode/text/UCD/TestCodeInvariants.java
srl295: ok

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
srl295: should but does not

unicodetools/src/test/java/org/unicode/text/UCD/TestCodeInvariants.java
srl295: removed

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
srl295: removed

unicodetools/src/test/java/org/unicode/text/UCD/TestCodeInvariants.java
```diff
@@ -115,10 +116,10 @@ public static int testScriptExtensions() {
                 System.out.println("Script Extensions invariant works for version " + age + "\n");
             }
 
-        return testResult;
+        assertEquals(TEST_PASS, testResult, "Invariant test for Script_Extensions failed!");
     }
 
-    public static int testGcbInDecompositions(boolean showAllNfds) {
+    public void testGcbInDecompositions() {
```
markusicu, line 122: Missed one test annotation...
```suggestion
    @Test
    public void testGcbInDecompositions() {
```

markusicu (commented):
Ok except for one more thing...

unicodetools/src/test/java/org/unicode/text/UCD/TestCodeInvariants.java
srl295: done

srl295: I keep dropping that… what do i have against GCB decompositions?!

markusicu (approved):
lgtm tnx -- please remember to *squash-and-merge* as usual

PR #217, opened 2022-04-06
add Eclipse .settings/ to .gitignore

markusicu:
After importing the new Maven project into Eclipse, git status reported that there was a new folder `unicodetools-testutils/.settings/`. AFAIU we don't want to check in IDE-specific settings.

PR #254, opened 2022-05-16
ignore Eclipse files + incremental 4-space indent Java formatter

echeran:
Based on previous discussion about #217, if we start ignoring Eclipse project settings dot files/folders in the codebase, we may also lose consistent source formatting (for those who use Eclipse).

Before progressing, we wanted to see if we could also handle source code style & formatting functionality in an IDE-agnostic way (while still having Eclipse support). This is similar to what we have done using Maven for build functionality.

This PR adds a couple of commits on top of #217 to add a Java source formatter that supports 4-space indents, and it only applies itself to files in a PR branch that have been changed (relative to the latest state of the codebase in the upstream repo).

This PR is a cleaned up version of https://github.com/echeran/unicodetools/pull/5, whose checks pass. Also, a negative test PR https://github.com/echeran/unicodetools/pull/6 adds a commit to its branch which touches a single file to create 2-space indents, and the [checks fail](https://github.com/echeran/unicodetools/runs/6461834592?check_suite_focus=true#step:4:2303).


pom.xml
```diff
@@ -118,6 +120,32 @@
 					<artifactId>maven-failsafe-plugin</artifactId>
 					<version>${maven-surefire-plugin-version}</version>
 				</plugin>
+				<!-- 
+					Spotless info:
+					  - How to preview what `mvn spotless:apply` will do: https://github.com/diffplug/spotless/tree/main/plugin-maven#how-do-i-preview-what-mvn-spotlessapply-will-do
+					  - Apply Spotless to specific files: https://github.com/diffplug/spotless/tree/main/plugin-maven#can-i-apply-spotless-to-specific-files
+				-->
+				<plugin>
+					<groupId>com.diffplug.spotless</groupId>
```
markusicu, line 129: Indentation in this file: Please stay consistent, and don't use tabs. 4 spaces??

pom.xml
```diff
@@ -118,6 +120,32 @@
...
+					  <!-- optional: limit format enforcement to just the files changed by this feature branch -->
+					  <!-- You can explicitly disable ratchet functionality by providing the value 'NONE': -->
+					  <ratchetFrom>origin/main</ratchetFrom>
+					  <!-- define a language-specific format -->
+					  <java>
+						<!-- no need to specify files, inferred automatically, but you can if you want -->
+				  
+						<!-- apply a specific flavor of google-java-format and reflow long strings -->
+						<googleJavaFormat>
+						  <version>1.8</version>
```
markusicu, line 142: Is this the Java version for Unicode Tools (which should be 11, right?), or the Java version for running googleJavaFormat? Or the version of the style checker?

pom.xml
echeran: Done. Fixed the inconsistent indentation.

That being said... since the rest of the file is using tabs, I made one commit to make the indentation consistent with the tab-based indents.  The subsequent commit converts the entire file from tabs to spaces.

It seems that multiple *.xml files in the repo use tabs, not spaces.  Interested in adding a [4-space indent rule for certain files](https://github.com/diffplug/spotless/tree/main/plugin-maven#generic-steps) to the formatter / style checker?

markusicu: > It seems that multiple *.xml files in the repo use tabs, not spaces. Interested in adding a [4-space indent rule for certain files](https://github.com/diffplug/spotless/tree/main/plugin-maven#generic-steps) to the formatter / style checker?

Yes, I think so.

@josh-hadley it looks like the tab checker does not look at XML files?

We probably can't enforce much about .txt files. Or maybe none of them even use indentation.

josh-hadley: Right, tab checker only looks at files ending with `.java`. Easy enough to add XML or others if desired.

srl295: our xml files [pom.xml, build.xml, all cldr data files] have traditionally used tabs.  Is there a reason to change?

srl295: Looks like icu4j's build.xml uses spaces, though

.github/workflows/cli-build-instructions.yml
```diff
@@ -14,12 +14,31 @@ env:
...
+    name: Formatter + Style checker
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Unicode Tools
+        uses: actions/checkout@v2
+      # Should we use "upstream" instead of "origin" in Spotless configuration
+      # for the git remote name for unicode-org/unicodetools?
+      - name: Fetch git branches for unicode-org/unicodetools
+        run: |
+          git fetch origin
```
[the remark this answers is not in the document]
srl295: or is this a different fetch than the PR's fetch?

srl295 (commented):
generally lgtm with comments

pom.xml
markusicu: Tabs are generally trouble because there is not a standard advance width for them. Many tools default to tab stops every 8 columns, but some people set some of their tools to 4, and when files use a mix of tabs and spaces (which often happens) the indentation is all messed up in one editor / code review tool or another. Or indentation with tabs and vertically aligned comments at the ends of lines. That's why we try to avoid tabs.

.github/workflows/cli-build-instructions.yml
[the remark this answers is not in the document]
echeran: Done. Thanks for pointing out the `fetch-depth` config for `actions/checkout`, I'm using that instead.  The reason for having the `git fetch` in the first place is because of the Spotless configuration for the "ratchet" feature that only checks formatting diffs to modified files in the PR (and it needs the reference point `origin/main` to diff and grok what's in the PR).... however, we are leaning towards having a single big bang change globally for formatting, which will remove noise from future PRs by allowing those future PRs to only spotlight the changed lines of interest.  That makes sense, and would obviate the need for the ratchet feature.

echeran:
I've updated this PR to include both: 1) adding the Eclipse project files to `.gitignore`, and 2) a) the formatter configs in CI + b) large initial application of the formatter to all files.  The PR already includes the latest from upstream.

Note: One detail that I encountered is that one of the CI jobs to auto run tools for smoke test purposes runs a tool (GenerateEnums) that generates some source code, and then does a diff, according to documented instructions, to ensure changes are committed.  So in that CI job, the formatter needs to be applied after GenerateEnums regenerates source code but before the `git diff` test occurs.

The main files to look at should be the root `pom.xml` and `.github/workflows/ci-build-instructions.yml` -- these control the changes.  The rest of the changes are derivative due to formatting, but within that set, maybe `unicodetools/src/main/java/org/unicode/props/UcdPropertyValues.java` is interesting as one of the files generated by GenerateEnums.

macchiati (approved):
I haven't looked over *all* the files, but did do some spot checks, and looks good to me.

PR #255, opened 2022-05-18
UCA 15 test data files

markusicu:


Ken-Whistler (approved):
Changes to CollationTest.html look correct. Cannot evaluate the tools code changes. For the output files, I'll check against my implementation after deployment.

PR #257, opened 2022-05-19
ZWJ/ZWNJ not Identifier_Status=Allowed

markusicu:
For
[[171-A128](https://www.unicode.org/cgi-bin/GetL2Ref.pl?171-A128)] Action Item for Mark Davis: In the property files IdentifierStatus.txt and IdentifierType.txt, remove the Joiner_Control characters ZWJ and ZWNJ from Identifier_Type=Inclusion and Identifier_Status=Allowed. For Unicode version 15.0.

removals.txt has overrides for ZWJ & ZWNJ, setting them to Identifier_Type=Inclusion.
By removing them, they default to Identifier_Type=Default_Ignorable.

All data files other than removals.txt are modified by the tool.

There are also changes for U+13440 EGYPTIAN HIEROGLYPH MIRROR HORIZONTALLY (still Exclusion but no longer Not_XID). I assume that they are due to changes in other properties.

The confusables changes are only in comments. I think I have seen some of these flip-flopping but didn't check the history.

Drive-by removal of some obsolete version lines that I noticed.

I regenerated the IDNA (UTS 46) files; they did not show any data changes.

I had to add an exception into TestInvariants for the intentional removals from Identifier_Type=Inclusion.

### DIFF
--- a/unicodetools/data/security/dev/IdentifierStatus.txt
+++ b/unicodetools/data/security/dev/IdentifierStatus.txt
@@ -1,5 +1,5 @@
 ﻿# IdentifierStatus.txt
-# Date: 2022-05-10, 17:59:54 GMT
+# Date: 2022-05-18, 21:51:57 GMT
 # © 2022 Unicode®, Inc.
 # Unicode and the Unicode Logo are registered trademarks of Unicode, Inc. in the U.S. and other countries.
 # For terms of use, see https://www.unicode.org/terms_of_use.html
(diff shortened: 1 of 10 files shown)


eggrobin:
> There are also changes for U+13440 EGYPTIAN HIEROGLYPH MIRROR HORIZONTALLY (still Exclusion but no longer Not_XID). I assume that they are due to changes in other properties.

Yes, specifically it follows from
> [[171-C21](https://www.unicode.org/cgi-bin/GetL2Ref.pl?171-C21)] **Consensus:** Change the General_Category property for U+13440 EGYPTIAN HIEROGLYPH MIRROR HORIZONTALLY from Cf to Mn in Unicode version 15.0 (Reference: section 16 of [L2/22-068](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/22-068))

which has the effect of bringing that character into XID_Continue.

eggrobin:
Speaking of U+13440 EGYPTIAN HIEROGLYPH MIRROR HORIZONTALLY, shouldn’t it be made Default Ignorable? It is not in the range of Egyptian hieroglyph format characters listed in https://www.unicode.org/reports/tr44/proposed.html#Default_Ignorable_Code_Point, but it is a mark with no ink, as mentioned in the SAH report cited by the aforementioned consensus:
>  It was noted that there are other characters with general category Mn with no ink: Variation Selectors, Mongolian FVSes,
Signwriting rotator and filler modifiers, TIFINAGH CONSONANT JOINER, and BRAHMI NUMBER JOINER.

The first two of these are default ignorable (though not the remainder).

I’ll add that to the PAG agenda.

eggrobin (approved):

macchiati (approved):

Issue #187, opened 2021-12-16
auto-run documented tool programs


markusicu:
In a CI workflow, run each documented tool program, like we currently run org.unicode.text.UCD.Main. Each tool should run successfully (not crash, not fail). For this issue, ignore the output.
- Look through the https://github.com/unicode-org/unicodetools/tree/main/docs pages (excluding "help")
- For each tool documented there,
  - Add it to the list below
  - Add it to CI

List of tools:
- [x] AacOrder.java, see https://github.com/unicode-org/unicodetools/blob/main/docs/emoji/aac.md#aacorderjava
- [ ] EmojiFrequency.java, see https://github.com/unicode-org/unicodetools/blob/main/docs/emoji/emoji-frequency.md#processing
  - May need a dummy .tsv file
- [x] org.unicode.text.UCA.Main, see https://github.com/unicode-org/unicodetools/blob/main/docs/uca/index.md#tools--tests
- [x] GenerateIdna.java, see https://github.com/unicode-org/unicodetools/blob/main/docs/idna.md
- [x] GenerateIdnaTest.java, see https://github.com/unicode-org/unicodetools/blob/main/docs/idna.md
- [x] GenerateEnums.java, see https://github.com/unicode-org/unicodetools/blob/main/docs/newunicodeproperties.md#run-generateenumsjava
  - Would be useful to fail if the generated output differs from the repo files.
- [x] GenerateConfusables, see https://github.com/unicode-org/unicodetools/blob/main/docs/security.md#generating
- [x] GenerateUnihanCollators, see https://github.com/unicode-org/unicodetools/blob/main/docs/unihan.md#run-generateunihancollators
- [x] CheckProperties, see https://github.com/unicode-org/unicodetools/blob/main/docs/newunicodeproperties.md#checking-other-properties

Not a tool:
- ToolUnicodePropertySource.java, see https://github.com/unicode-org/unicodetools/blob/main/docs/changing-ucd-properties.md#toolunicodepropertysourcejava

Skip:
- GenerateUnihanCollatorFiles, see https://github.com/unicode-org/unicodetools/blob/main/docs/unihan.md#run-generateunihancollatorfiles
- GenerateEmoji.java, see https://github.com/unicode-org/unicodetools/blob/main/docs/emoji/generate-emoji.md#running-the-tool
  - 2022-mar-22: out of scope because it needs access to another repo, see [discussion below](https://github.com/unicode-org/unicodetools/issues/187#issuecomment-1075566958)

[...]

PR #261, opened 2022-05-24
Add the `ICU` arg to CI test for UCA to also build files for CLDR/ICU

echeran:
Part of #187


markusicu (approved):
lgtm nice :-)

Will see what happens when we add UCD data for more new characters before their sort order data is defined...

PR #272, opened 2022-06-06
Reorder several new emoji based on ESC feedback

nedley:
Corresponding to https://github.com/unicode-org/emoji/pull/26


nedley:
@macchiati I will be out the rest of this week, if there is urgency in conveying this to CLDR could you help out?

macchiati (approved):
Thanks! Once this is in, let me know so we can update the CLDR data. (We'll need a regenerated emoji-test file.)

nedley:
emoji-test.txt and not something like emoji-ordering-rules.txt? If the latter, it has already been committed the emoji repo. If the former, I can rebuild and repost once I’m back from vacation next week.

josh-hadley:
@nedley when you're ready please Squash and merge

macchiati:
It uses the emoji-test file plus the ordering rules. The former gives cldr
the category information, while the latter provides the ordering.

nedley:
> Okay, feel free to take whatever you need.

Ah, I guess you still need a new emoji-text.txt? Will get that in shortly, sorry for my poor reading comprehension.

macchiati:
NP

nedley:
@markusicu One might ask the same about your question. ;)

PR #276, opened 2022-06-28
Update emoji data files to latest ordering

nedley:
Reflect current status after https://github.com/unicode-org/unicodetools/pull/272


nedley:
@macchiati Is this sufficient for CLDR?

nedley:
At this point I am trying to get some traction with Mark only. We have #283 for cleaning this the duplicated files later.

unicodetools/data/emoji/dev/internal/emojiOrdering.txt
```diff
@@ -137,214 +135,14 @@
 👭
 👫
 👬
-💏
```
macchiati, line 140: Something funny is going on with the ordering. 
For the eventual ICU collation rules, normally we need to have **all** combinations of modifiers that are applicable to a human ZWJ sequence. There is one exception: any final modifier can be left off.

Now, the EmojiOrdering class has some logic built in to avoid listing all the sequences, but I suspect that it doesn't handle these cases properly, which is why they are explicitly listed. So their removal makes me nervous.

unicodetools/data/emoji/dev/emoji-test.txt
```diff
@@ -4533,9 +4535,9 @@
 1F505                                                  ; fully-qualified     # 🔅 E1.0 dim button
 1F506                                                  ; fully-qualified     # 🔆 E1.0 bright button
 1F4F6                                                  ; fully-qualified     # 📶 E0.6 antenna bars
+1F6DC                                                  ; fully-qualified     # 🛜 E15.0 wireless
 1F4F3                                                  ; fully-qualified     # 📳 E0.6 vibration mode
 1F4F4                                                  ; fully-qualified     # 📴 E0.6 mobile phone off
-1F6DC                                                  ; fully-qualified     # 🛜 E15.0 wireless
 
 # subgroup: gender
 2640 FE0F                                              ; fully-qualified     # ♀️ E4.0 female sign
```
macchiati, line 4543: All of the changes to the test file look reasonable.

macchiati:
PS. Before the alpha integration with ICU we'll also need
emoji-ordering-rules.txt
(see https://cldr.unicode.org/development/generate-emoji-paths) @Peter
Edberg ***@***.***>

unicodetools/data/emoji/dev/internal/emojiOrdering.txt
nedley: None of these sequences were in https://www.unicode.org/emoji/charts-14.0/emoji-ordering-rules.txt, and they continue to not be in https://www.unicode.org/emoji/charts-15.0/emoji-ordering-rules.txt… since this file being in dev/ is only causing consternation, can it safely be removed?

macchiati:
Fine by me

markusicu:
Re commit [e038f1d](https://github.com/unicode-org/unicodetools/pull/276/commits/e038f1d52834f9fb602ef2c367b314b32d0d101d) -- we were pushing to consolidate on the data file in the data folder, rather than the duplicate in the to-be-jarred folder...

unicodetools/data/emoji/dev/internal/emojiOrdering.txt
markusicu: unicodetools/data/... is the normal place for data files. As noted in #283 that's where we would expect the one file to live.

Ned had said "We have #283 for cleaning up the duplicated files later." -- that would be ok with me. That is, keeping duplicates in this PR (with the same contents) and deduplicating later.

markusicu:
> Fine by me

Mark: When you reply by email, the comments don't thread properly. Please reply to GitHub comments by clicking on their GitHub links and replying in the UI.

nedley:
> Fine by me

r?

nedley:
I would love to get this merged. Any takers?

macchiati (approved):

macchiati:
done. (Sorry, was on my phone earlier, and the Add Review button didn't
show up)

markusicu:
I think this removed the version of emojiOrdering.txt that we should have kept... See my comment on issue #283.

PR #281, opened 2022-07-08
4 Nag Mundari signs should have lb=CM

markusicu:
Clerical error noted by David Corbett during Unicode 15 beta. Modified LineBreak.txt from Ken Whistler.


Ken-Whistler (approved):
Changes as expected.

PR #286, opened 2022-07-18
Unicode 15 character index additions

markusicu:
From @Ken-Whistler and Joe Becker.

First tranche: The usual small set of additions of entries for new blocks.

Second: the rest of the recommendations from Joe, with some adjustments


Ken-Whistler (approved):
Look as expected.

PR #287, opened 2022-07-20
Fix minimally-qualified sequences

nedley:
Fixes [171-A119](https://www.unicode.org/cgi-bin/GetL2Ref.pl?171-A119)


nedley:
@macchiati ping

macchiati (approved):
excellent, thanks!

PR #288, opened 2022-07-21
fix 6 Latin char names

markusicu:
From Ken:
I just regenerated allkeys.txt (and the CTT for 14651) with the fix for the 6 Latin characters with those left hook names we changed.
No other implications or changes in behavior. The character names are just in comments on the lines.


Ken-Whistler (approved):
Look as expected.

PR #294, opened 2022-08-05
John Unihan 15 updates post UTC 172

markusicu:
This update accounts for a large number of the CJK action items for Michel and for @Tsengtsz.


Ken-Whistler (approved):
I can't comment on the details, but these look like the expected batch of changes for John's latest additions. Ken Lunde reviewed those in detail, so these should be o.k. here.

Tsengtsz (approved):

PR #298, opened 2022-08-08
VerticalOrientation.txt updated: U+13440 gc=Mn

markusicu:
Laurentiu caught and fixed a leftover range break issue resulting from the change of 13440 from gc=Cf --> Mn.


Ken-Whistler (approved):
As expected. Looks o.k.

PR #302, opened 2022-08-09
UCD 15.0: USourceGlyphs and USourceRSChart updated

markusicu:
From Ken: ... USourceGlyphs.pdf and USourceRSChart.pdf with their cover sheets ... These should now be final for the release


Ken-Whistler (approved):
I presume this is o.k. The changes can't be viewed, because they are binary pdf files.

Tsengtsz (approved):

PR #303, opened 2022-08-10
UCA 15: sort 2 Tibetan chars like other bindus

markusicu:
From Ken:

I've taken care of my action 172-A18, which was to change the weighting for 0F82/0F83 to match other bindus. The change in the weighting is very localized, as it just updated those two from `[0000.0000.0000]` to have the same secondary weight as other bindus. The change is easily diffable in the output. This also involved updating the decomps.txt file, as it reflects these artificial equivalences added to the input.

I also posted the input file, unidata.txt, along with the two others in my incoming dropoff directory.

If you compare the input file with the previous version, you may note that I also dropped a few other name notes for related Tibetan characters. Those are just old values from the obsolete Unicode_1_Name field that I hadn't gotten around to cleaning up. They have no impact on the DUCET generation.


Ken-Whistler (approved):
Changes appear as expected. The additions for 10EFD, 1E4EC, etc. in the test files seem to result from generation after the library knows that those new characters are combining marks, so the order wrt 0334 might matter for collation.

PR #307, opened 2022-08-15
Names List updated to final for 15.0

markusicu:
From Ken: This rolls in all the open actions I knew about for beta feedback, etc. It is mostly a bunch of small, local annotation fixes. I synched the notice for the Kazakh letters 0675..0678


Ken-Whistler (approved):
Changes look as expected.

PR #308, opened 2022-08-15
Include end names in description of emoji ranges

nedley:
Fixes [162-A13](https://www.unicode.org/cgi-bin/GetL2Ref.pl?162-A13) “Action Item for Mark Davis: Make data/tooling fixes documented in [L2/20-050](http://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/20-050), based on Charlotte Buff's feedback”


macchiati (approved):
Thanks!

nedley:
Thinking some more I’m not sure about doing this for 15.0; my concern is that the range here is not in a comment and runs the risk, however small, of confusing some parser somewhere. Maybe it would be better to restructure this file in a future release to stop exposing _any_ CLDR short names as a property field in this file.

Ken-Whistler (commented):
Changes look good. It seems better to do it than to not do it at this point, but decide quickly and stick to the decision for 15.0, whichever way it goes.

macchiati:
I wouldn't go so far as to remove it in 15.0. Any implementation that uses
the names is a failure anyway because of the ranges.

(As a matter of fact, I know of implementations that pull the names out of
emoji-test.txt, but there are no ranges there, so every name is discrete)

nedley:
> I wouldn't go so far as to remove it in 15.0. Any implementation that uses the names is a failure anyway because of the ranges.

True, which is why I would suggest moving the names to a comment (similar to emoji-data.txt) in a future version.

> (As a matter of fact, I know of implementations that pull the names out of emoji-test.txt, but there are no ranges there, so every name is discrete)

Yes, but those are also in a comment…

macchiati:
> True, which is why I would suggest moving the names to a comment (similar
to emoji-data.txt) in a future version.
Ah, I was thinking you meant removal. Yes, into a comment would be fine
post 15.

Issue #289, opened 2022-07-27
Consider test: InSc=Bindu --> Alphabetic with exceptions


markusicu:
The [UTC 172 script ad hoc report](https://www.unicode.org/L2/L2022/22128-script-adhoc-rept.pdf) item VII 27 “Nonalphabetic bindus” points out that “Most characters with InSC=Bindu have Alphabetic=Yes.”

In discussion, attendees agreed, and wished for an invariant test, but there could be exceptions. Consider adding a test, with a comment that this is not a hard & fast rule, and to allow for exceptions.

markusicu:
https://github.com/unicode-org/unicodetools/blob/main/docs/build.md#invariant-checking --> https://github.com/unicode-org/unicodetools/blob/main/unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt

markusicu:
https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AInSc%3DBindu%3A%5D-%5B%3AAlphabetic%3A%5D&g=gc&i=

Currently 5 exceptions.
- is the rule roughly right?
- if so, are these intentional exceptions?
- otherwise, are these wrong? should these 5 be alphabetic?

josh-hadley:
Once #318 is sorted, I will follow-up with SAH on the questions Markus mentions above.

markusicu:
Thanks for the test!
Should we keep this issue open for the SAH review of the exceptions?
Or do you want to track that in some other way?

josh-hadley:
I'd like to keep it open until we complete the review w/SAH (starting an email on that right now).

josh-hadley:
@markusicu reviewing the discussion from UTC #172 on this, SAH determined that those 5 exceptions are in fact mistakes and should be changed to `Alphabetic=Yes` we have made that change in #296 (AI 172-A17). So probably that is why our original implementation of the rule failed -- the exceptions are already removed. The way the rule is written now passes even with the change in place, but I think it should actually be changed to `[\p{InSC=Bindu} - \p{Alphabetic}] = []`.

macchiati:
Agreed.

markusicu:
Thanks for following up, and sorry for the outdated info in this issue. I wrote this issue during the UTC meeting, apparently before the AI for the property change, or else not thinking of the connection.

[...]

PR #318, opened 2022-08-23
Add test for non-alphabetic InSC=Bindu

josh-hadley:
- Updated UnicodeInvariantTest.txt (Misc. section) with check for non-Alphabetic InSc=Bindus. See #289.
- Added step to cli-build-instructions.yml workflow to upload generated `UnicodeTestResults.html` file as a build artifact (out-of-source build). This step is set to run `always()` so it will attempt to upload the artifact even if previous steps fail.

### DIFF
--- a/.github/workflows/cli-build-instructions.yml
+++ b/.github/workflows/cli-build-instructions.yml
@@ -76,14 +76,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           
-      - name: Upload UnicodeTestResults
-        if: always()
-        uses: actions/upload-artifact@v3
-        with:
-          name: unicode-test-results
-          path: |
-            Generated/UnicodeTestResults.*
-
   out-of-source-build:
     name: Out-of-source Instructions
     runs-on: ubuntu-latest
@@ -125,6 +117,14 @@ jobs:
           MAVEN_OPTS="-ea" mvn -s .github/workflows/mvn-settings.xml package -DCLDR_DIR=$(cd ../../../cldr/mine/src ; pwd)  -DUNICODETOOLS_GEN_DIR=$(cd ../Generated ; pwd)  -DUNICODETOOLS_REPO_DIR=$(pwd)  -DUVERSION=$CURRENT_UVERSION
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Upload UnicodeTestResults artifact
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: unicode-test-results
+          path: |
+            unicodetools/mine/Generated/UnicodeTestResults.*
 
       - name: Run command - Make Unicode Files
         run: |
--- a/.github/workflows/cli-build-instructions.yml
+++ b/.github/workflows/cli-build-instructions.yml
@@ -75,7 +75,7 @@ jobs:
         run: MAVEN_OPTS="-ea" mvn -s .github/workflows/mvn-settings.xml exec:java -Dexec.mainClass="org.unicode.text.UCD.Main"  -Dexec.args="version $CURRENT_UVERSION build MakeUnicodeFiles"  -pl unicodetools  -DCLDR_DIR=$(cd ../cldr ; pwd)  -DUNICODETOOLS_GEN_DIR=$(cd Generated; pwd)  -DUNICODETOOLS_REPO_DIR=$(pwd)  -DUVERSION=$CURRENT_UVERSION
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          
+
   out-of-source-build:
     name: Out-of-source Instructions
     runs-on: ubuntu-latest
--- a/.github/workflows/cli-build-instructions.yml
+++ b/.github/workflows/cli-build-instructions.yml
@@ -75,6 +75,14 @@ jobs:
         run: MAVEN_OPTS="-ea" mvn -s .github/workflows/mvn-settings.xml exec:java -Dexec.mainClass="org.unicode.text.UCD.Main"  -Dexec.args="version $CURRENT_UVERSION build MakeUnicodeFiles"  -pl unicodetools  -DCLDR_DIR=$(cd ../cldr ; pwd)  -DUNICODETOOLS_GEN_DIR=$(cd Generated; pwd)  -DUNICODETOOLS_REPO_DIR=$(pwd)  -DUVERSION=$CURRENT_UVERSION
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          
+      - name: Upload UnicodeTestResults
+        if: always()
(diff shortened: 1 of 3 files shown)


.github/workflows/cli-build-instructions.yml
```diff
@@ -118,6 +118,14 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Upload UnicodeTestResults artifact
+        if: always()
+        uses: actions/upload-artifact@v3
```
echeran, line 123: Observation: This looks good. The artifact will be available for manual download when viewing the page for this particular workflows run instance.

If we want to do something fancier (have that HTML be always visible and up-to-date at a particular URL), we could look at that in a later issue.

echeran (approved):

.github/workflows/cli-build-instructions.yml
josh-hadley: > have that HTML be always visible and up-to-date at a particular URL

Oooh, yeah I like that idea, like publish it via Pages or something? Maybe other reports too... 🔮 future project maybe 😸

josh-hadley: The artifact for the most recent run on this PR can be found here: https://github.com/unicode-org/unicodetools/actions/runs/2914569939  ("Artifacts" section at the bottom, click the name which will download a .zip containing the HTML file).
I'm open to adding other files from `Generated` if they would be useful. I mainly added this because there was previously no way to see the output of that test from GitHub CI runs and it's handy for troubleshooting Invariant test failures.

unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
```diff
@@ -431,6 +431,10 @@ Show [\u20b9]
 # [\p{Alphabetic}] ∥ \p{Script=Common}
 #  & [\p{Decomposition_Type=None} \p{Decomposition_Type=Canonical}]
 
+# Nonalphabetic bindus
```
markusicu, line 434:
```suggestion
# The UTC 172 script ad hoc report (L2/22-128) item VII 27 “Nonalphabetic bindus” points out that
# “Most characters with InSC=Bindu have Alphabetic=Yes.” but there could be exceptions.
# TODO: Review with SAH:
# - is the rule roughly right?
# - if so, are these intentional exceptions?
# - otherwise, are these wrong? should these 5 be alphabetic?
```

.github/workflows/cli-build-instructions.yml
markusicu: I suggest adding a sentence to https://github.com/unicode-org/unicodetools/blob/main/docs/build.md#invariant-checking about the CI providing this artifact, and how to fetch it.

markusicu (commented):
nice!

unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
```diff
@@ -431,6 +431,10 @@ Show [\u20b9]
...
+Let $nonAlphaBindus = [\u0C04\u0F82\u0F83\U00011080\U00011081]
+[\p{InSc=Bindu} - $nonAlphaBindus - \p{Alphabetic}] = []
```
macchiati, line 436: By the way, I don't quite understand why the other formulation wasn't working, but this is a better formulation anyway, because we don't care that they are exactly equal, just that if we subtract them as well as alphabetic, we are down to zero.

It could also be expressed as:

\p{InSc=Bindu} ⊆ [$nonAlphaBindus  \p{Alphabetic}]

.github/workflows/cli-build-instructions.yml
```diff
@@ -75,7 +75,7 @@ jobs:
         run: MAVEN_OPTS="-ea" mvn -s .github/workflows/mvn-settings.xml exec:java -Dexec.mainClass="org.unicode.text.UCD.Main"  -Dexec.args="version $CURRENT_UVERSION build MakeUnicodeFiles"  -pl unicodetools  -DCLDR_DIR=$(cd ../cldr ; pwd)  -DUNICODETOOLS_GEN_DIR=$(cd Generated; pwd)  -DUNICODETOOLS_REPO_DIR=$(pwd)  -DUVERSION=$CURRENT_UVERSION
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
+
```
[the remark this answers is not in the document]
josh-hadley: Fixed in https://github.com/unicode-org/unicodetools/pull/318/commits/b84e46831b084dfa39687a8226ddc10c7905917a

.github/workflows/cli-build-instructions.yml
```diff
@@ -236,4 +244,3 @@ jobs:
           mvn -s .github/workflows/mvn-settings.xml -Dexec.mainClass="org.unicode.propstest.CheckProperties" -Dexec.classpathScope=test test-compile  -Dexec.args="COMPARE ALL $PREVIOUS_UVERSION" exec:java  -pl unicodetools  -DCLDR_DIR=$(cd ../../../cldr/mine/src ; pwd)  -DUNICODETOOLS_GEN_DIR=$(cd ../Generated ; pwd)  -DUNICODETOOLS_REPO_DIR=$(pwd)  -DUVERSION=$CURRENT_UVERSION
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
josh-hadley: Not sure what's going on with this, it seems to have an LF

unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
josh-hadley: Done in https://github.com/unicode-org/unicodetools/pull/318/commits/e610d63ab47717ed45dd59f652e474f6f833a90e

.github/workflows/cli-build-instructions.yml
josh-hadley: Done in https://github.com/unicode-org/unicodetools/pull/318/commits/e0e6afa39664db148c02920e66285577a1e1059e

unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
josh-hadley: Thanks, Mark! We'll revisit this after checking in with SAH as to the results.

markusicu: I think we got these five exceptions from the Unicode utilities running on Unicode 14. It's possible that the set of exceptions is slightly different now. As we discuss the TODO we could experiment with the exceptions set, such as emptying it and looking at the invariant test failure.

markusicu (approved):
lgtm tnx!

markusicu:
(please *squash*-and-merge when ready)

PR #322, opened 2022-08-26
security data files without UTF-8 BOM

### LINKED ISSUE #123: omit UTF-8 signature (BOM) from generated files
Ken:

In double-checking all the data files for up-to-date copyright notices,
I stumbled on the fact that *all* of the security data files in the
release (other than the ReadMe.txt) have an unnecessary initial UTF-8
BOM on the file. This appears to be the result of some bogus conversion
step for output somewhere in the generating tools.

This is not a regression, because I checked that it has been the case
for all these files going back at least to Unicode 10.0. But it
definitely should be fixed in the tools, as it is somewhat embarrassing
to be officially recommending against unnecessary initial UTF-8 BOMs in
UTF-8 text files, but then going ahead and using them in the release of
our *security* data files. ;-)

See, e.g. confusables.txt, IdentifierType.txt, intentional.txt in the
security/14.0.0/ directory.

Steven:

ICU and CLDR have automated tools which check for stray BOMs.  These could be used as part of a test case.  If it’s the unicodetools repository that has these files, it could have an automated step to scan - excluding any known historical BOMs. This would prevent check-in of any such files.

markusicu:
The file-initial Byte Order Mark is not recommended for UTF-8 (where it is more properly a ["signature byte sequence"](https://www.unicode.org/reports/tr6/tr6-4.html#Signature)). Stop writing it for security data files.

Fixes #123

### DIFF
--- a/unicodetools/data/security/dev/IdentifierStatus.txt
+++ b/unicodetools/data/security/dev/IdentifierStatus.txt
@@ -1,5 +1,5 @@
-﻿# IdentifierStatus.txt
-# Date: 2022-08-22, 23:13:14 GMT
+# IdentifierStatus.txt
+# Date: 2022-08-26, 16:49:09 GMT
 # © 2022 Unicode®, Inc.
 # Unicode and the Unicode Logo are registered trademarks of Unicode, Inc. in the U.S. and other countries.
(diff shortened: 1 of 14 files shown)


Ken-Whistler (approved):
Looks fine.

PR #326, opened 2022-09-02
Unicode 15 final readmes

markusicu:
For each file, I am using the same "final" wording as in Unicode 14.


rickmcgowan (approved):

Ken-Whistler (approved):
Changes look fine.

josh-hadley (approved):

Issue #271, opened 2022-06-03
Jsps to Unicode 15 beta


markusicu:
I suggest updating the Jsps to Unicode 15 beta, and advertising that on the unicode/unicore lists, to help with beta review.

@macchiati @srl295

srl295:
followon to #140

srl295:
- pull https://github.com/unicode-org/icu/releases/tag/cldr%2F2022-06-27
- pull similar CLDR version?

macchiati:
Missed this earlier, yes, a good idea. I haven't looked at this in a while: The binary versions of the properties have to be updated in the jsp directories, and there is a switch that enables it. I think just the main version of CLDR would work.

srl295:
@markusicu @macchiati  I see this:

> Z-Other … Emoji_Keycap_Sequence

srl295:
OK. I'm able to build again. Now to make the build actually work..

srl295:
Deployed to https://util.unicode.org/

macchiati:
Added https://github.com/unicode-org/unicodetools/issues/336

PR #332, opened 2022-09-16
chore: Jsps to Unicode 15 final

srl295:
For #271


markusicu:
Push to GCR failed.

markusicu:
Thanks!
This looks rubber-stamp-worthy to me.
For next week, I plan to
- download the final Unicode 15 data files and sync back into this repo
- copy the "dev" folders to "15.0[.0]" folders

Does this change when you want to upgrade the JSPs to 15?

srl295:
> push failed

yes, fix in progress


> Thanks! This looks rubber-stamp-worthy to me. For next week, I plan to
> 
> * download the final Unicode 15 data files and sync back into this repo
> * copy the "dev" folders to "15.0[.0]" folders
> 
> Does this change when you want to upgrade the JSPs to 15?

I'm … not sure.   I built the data with '-DUVERSION=15.0.0'

srl295:
@markusicu  live @ https://util.unicode.org/UnicodeJsps/

srl295:
CLDR is now(?!) needed to be included in the build image… have to revamp the docker image… will work on it monday

.github/workflows/push-jsp-on-tag.yml
```diff
@@ -12,6 +12,15 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 11
+    - uses: actions/checkout@v2
```
markusicu, line 15: As its own list item with just this single line, this looks like a no-op. What am I missing? (Other than general GHA ignorance.)

.github/workflows/build-jsp.yml
```diff
@@ -17,6 +17,9 @@ jobs:
       with:
         repository: unicode-org/cldr
         path: cldr
+    - name: Backup Unicodetools and CLDR for jsps  # this is needed only for the Docker build
```
markusicu, line 20: dumb question: Where does it get backed up to, and how does another step find it?

.github/workflows/push-jsp-on-tag.yml
```diff
@@ -12,6 +12,15 @@ jobs:
...
+    - name: Check out CLDR
+      uses: actions/checkout@v2
+      with:
+        repository: unicode-org/cldr
+        path: cldr
+    - name: Backup Unicodetools and CLDR for jsps
```
markusicu, line 21: dumb question: Why do two workflows need to do the same backup?

markusicu (approved):
I trust you, and the JSPs work, but I have more questions (uninformed ones) than real opinions... Rubber-stamping.

.github/workflows/build-jsp.yml
srl295: not a dumb question.  It gets copied to `UnicodeJsps/target/cldr-unicodetools.tgz` as a temporary file, and is then picked up in the Dockerfile. It's a copy of the repositories not including any build artifacts or git metadata, to keep the docker image smaller.

.github/workflows/push-jsp-on-tag.yml
srl295: because they both attempt to build the Docker image.  This one is only run on tag, the other is run on every commit.

srl295: i had neglected the build-jsp.yml at first which is why the build failed before

PR #340, opened 2022-09-30
Unicode 15.1 version numbers

markusicu:
Updating version numbers, with latestVersion=15.1.0.

I also modified GenerateEnums.java to write `@Override` annotations that Eclipse adds automatically, so that when one touches UcdPropertyValues.java in Eclipse, the CI check does not complain about unexpected diffs.

The second commit makes minor version number changes in generated segmentation .html files.


.github/workflows/build-jsp.yml
```diff
@@ -43,16 +43,16 @@ jobs:
       run: >
         mkdir -pv $(pwd)/output/Generated/ &&
         mvn -s .github/workflows/mvn-settings.xml -B exec:java
-        -Dexec.mainClass="org.unicode.text.UCD.Main" -Dexec.args="version 15.0.0 build MakeUnicodeFiles"
+        -Dexec.mainClass="org.unicode.text.UCD.Main" -Dexec.args="version 15.1.0 build MakeUnicodeFiles"
```
nedley, line 46: Is there any way to expand this from the `CURRENT_UVERSION` token in cli-build-instructions.yml?

markusicu: I *suspect* not, since it's in a different workflow file. But I am no expert in GitHub Actions.

nedley: I am _entirely_ unsure myself, but even if we needed to create a new file containing just the constant such that it could be used by the others I think it would be worth doing.

markusicu: --> #344

markusicu:
I ran the UCD file generator and added a second commit with minor version number changes in generated segmentation .html files.

Manishearth (approved):

PR #342, opened 2022-10-05
Run exec:java targets without install

### LINKED ISSUE #317: Run CI jobs without `install` for subprojects
In Unicode Tools, we have a situation where one subproject (`UnicodeJsps`) depends on another (`unicodetools`).

In some of our CI jobs, an initial step will run `mvn install` on the dependency subproject in order to be able to run commands in the dependent subproject.  But the problem with `mvn install` is that it affects the local Maven cache, which is undesirable as it could create caching problems.

Is it possible to avoid running `install` and still run the desired commands within the desired subprojects succesfully?  Would the Maven [`-am`](https://maven.apache.org/ref/current/maven-embedder/cli.html) option in combination with `-pl` work, or not?

echeran:
Fixes #317

Follows the details in the comment that I left on [this Stack Overflow answer](https://stackoverflow.com/a/26448447/2077918).


echeran:
CC @yumaoka

markusicu (approved):
(partial rs)lgtm tnx!

markusicu:
Unfortunately, you have a merge conflict with my PR #340 :-/

srl295 (dismissed):

markusicu (approved):
lgtm tnx

PR #351, opened 2022-11-30
Support U+ notation in character.jsp

### LINKED ISSUE #350: The Character Properties utility should support U+ notation
From @PeterConstable on properties@unicode.org, 2022-08-17:
> Why doesn’t the [Character Properties utility](https://util.unicode.org/UnicodeJsps/character.jsp?a=5&B1=Show) allow you to type code points in “U+” notation, which is the conventional citation form for code points?

> After trying that and getting properties for “U”, it’s not at all obvious that the non-conventional form without “U+” would work, though I suppose one might argue that the report shows the other form.

I replied:
> I have been bitten by that and shown the properties of LATIN CAPITAL LETTER U many times, so I very much agree that this should support U+ notation.

> This should be easy enough to add, the parsing code appears to be [here](https://github.com/unicode-org/unicodetools/blob/70dce2c89f185c65b436c28404ae5b7bdb32c2d1/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeJsp.java#L211) (called from [here](https://github.com/unicode-org/unicodetools/blob/fc25a0e1d7b1123bc8da239cd7a9c7f98f6e8a90/UnicodeJsps/src/main/webapp/character.jsp#L27)), but I have yet to do the [setup](https://github.com/unicode-org/unicodetools/blob/main/docs/unicodejsps/index.md) to build this stuff.

KenW replied:

> For what it is worth, the Code2Chart code chart lookup CGI utility used on the code charts index page:

> https://www.unicode.org/charts/

> has long supported entry as U+212A, as well as \u212A, \U212A, and for that matter 0x212A, as well as bare hex entry 212A, and lowercase hex digits, too, too: 212a, face, etc.

eggrobin:
As well as a number of other hex escape and hex numeric literal syntaxes.

This fixes #350.

Provide some help in an error message if we still get multiple code points, pointing to the pages for the individual code points, which can be useful when one is interested in combining marks; see screenshots below.

Drive-by changes to documentation so that it doesn’t suggest a running maven in a place where it doesn’t see the POM, and to .gitignore so that it ignores the Generated directory suggested by the build instructions.

Unfortunately this breaks this important feature of character.jsp: https://util.unicode.org/UnicodeJsps/character.jsp?a=%F0%9F%90%88%22/%3E%3Cscript%3Ealert(%22meow%22)%3C/script%3E%3Cspan%20class=%22&B1=Show.

Screenshots:

| U+ notation | Multiple code points | Escaping |
|---|---|---|
| image | image | image |

### DIFF
--- a/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeJsp.java
+++ b/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeJsp.java
@@ -214,7 +214,8 @@ public class UnicodeJsp {
         // Accept U+ notation and standalone hexadecimal digits, as well as a variety
         // of hexadecimal character escape and numeric literal syntaxes.
         Matcher matcher =
-                Pattern.compile("(?:U\\+|\\\\[ux]\\{?|&#x|0x|16#|&H)?([0-9a-f'_]+)[\\};#]?",
+                Pattern.compile(
+                                "(?:U\\+|\\\\[ux]\\{?|&#x|0x|16#|&H)?([0-9a-f'_]+)[\\};#]?",
                                 Pattern.CASE_INSENSITIVE)
                         .matcher(trimmed);
         String digits = matcher.matches() ? matcher.group(1).replaceAll("['_]", "") : null;
--- a/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeJsp.java
+++ b/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeJsp.java
@@ -208,28 +208,37 @@ public class UnicodeJsp {
         UnicodeUtilities.getDifferences(setA, setB, abbreviate, abResults, abSizes, abLinks);
     }
 
-    public static int parseCode(String text, String nextButton, String previousButton) {
+    public static int[] parseCode(String text, String nextButton, String previousButton) {
         // text = fromHTML.transliterate(text);
         String trimmed = text.trim();
-        if (trimmed.length() > 1) {
+        // Accept U+ notation and standalone hexadecimal digits, as well as a variety
+        // of hexadecimal character escape and numeric literal syntaxes.
+        Matcher matcher = Pattern.compile(
+            "(?:U\\+|\\\\[ux]\\{?|&#x|0x|16#|&H)?([0-9a-f'_]+)[\\};#]?",
+            Pattern.CASE_INSENSITIVE).matcher(trimmed);
+        String digits = matcher.matches() ? matcher.group(1).replaceAll("['_]", "") : null;
+        if (digits != null && digits.length() > 1) {
             try {
-                text = UTF16.valueOf(Integer.parseInt(trimmed, 16));
+                text = UTF16.valueOf(Integer.parseInt(digits, 16));
             } catch (Exception e) {
             }
         }
-        int cp = UTF16.charAt(text, 0);
+        int[] result = text.codePoints().toArray();
+        int cp = result[0];
         if (nextButton != null) {
             cp += 1;
             if (cp > 0x10FFFF) {
                 cp = 0;
             }
+            return new int[]{cp};
         } else if (previousButton != null) {
             cp -= 1;
             if (cp < 0) {
                 cp = 0x10FFFF;
             }
+            return new int[]{cp};
         }
-        return cp;
+        return result;
     }
 
     public static String getConfusables(String test, int choice) {
--- a/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeJsp.java
+++ b/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeJsp.java
@@ -213,9 +213,10 @@ public class UnicodeJsp {
         String trimmed = text.trim();
         // Accept U+ notation and standalone hexadecimal digits, as well as a variety
         // of hexadecimal character escape and numeric literal syntaxes.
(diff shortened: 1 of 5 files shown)


eggrobin:
FWIW maven spotless:apply (and check) fails on my machine with the following error:
> [ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.22.5:apply (default-cli) on project unicodetools-testutils: Execution default-cli of goal com.diffplug.spotless:spotless-maven-plugin:2.22.5:apply failed: java.lang.reflect.InvocationTargetException: class com.google.googlejavaformat.java.JavaInput (in unnamed module @0x692e028d) cannot access class com.sun.tools.javac.parser.Tokens$TokenKind (in module jdk.compiler) because module jdk.compiler does not export com.sun.tools.javac.parser to unnamed module @0x692e028d

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeJsp.java
```diff
@@ -208,28 +208,39 @@ public static void getDifferences(
...
+        String digits = matcher.matches() ? matcher.group(1).replaceAll("['_]", "") : null;
+        if (digits != null && digits.length() > 1) {
             try {
-                text = UTF16.valueOf(Integer.parseInt(trimmed, 16));
+                text = UTF16.valueOf(Integer.parseInt(digits, 16));
             } catch (Exception e) {
             }
         }
-        int cp = UTF16.charAt(text, 0);
+        int[] result = text.codePoints().toArray();
```
echeran, line 228: observation: `.codePoints()` is a Java 8 method on `CharSequence` that returns an `IntStream` (a `Stream` of ints). I didn't know that existed in Java 8.

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeJsp.java
```diff
@@ -208,28 +208,39 @@ public static void getDifferences(
         UnicodeUtilities.getDifferences(setA, setB, abbreviate, abResults, abSizes, abLinks);
     }
 
-    public static int parseCode(String text, String nextButton, String previousButton) {
+    public static int[] parseCode(String text, String nextButton, String previousButton) {
```
echeran, line 211: observation: I'm unopinionated on the return type, and what you have seems reasonable. I don't know if matters whether it is an int array `int[]` or a collection type object that implements `List<T>`.

UnicodeJsps/src/main/webapp/character.jsp
```diff
@@ -24,14 +24,44 @@ th           { text-align: left }
...
 		String text = utfParameters.getParameter("a", "\u2615", "\u2615");
 
-		int cp = UnicodeJsp.parseCode(text,null,null);
-		//text = UTF16.valueOf(cp);
-	    String nextHex = "character.jsp?a=" + Utility.hex(cp < 0x110000 ? cp+1 : 0, 4);
+		int[] codePoints = UnicodeJsp.parseCode(text,null,null);
+		int cp = codePoints[0];
+    String nextHex = "character.jsp?a=" + Utility.hex(cp < 0x110000 ? cp+1 : 0, 4);
 		String prevHex = "character.jsp?a=" + Utility.hex(cp > 0 ? cp-1 : 0x10FFFF, 4);
+    if (codePoints.length > 1) {
```
echeran, line 31: praise: this is useful, as are other drive-by fixes in this PR

docs/unicodejsps/index.md
```diff
@@ -15,7 +15,7 @@ need to actually need to download or build CLDR itself.
 
 ### Building from the command line
 
-Run `mvn package` in the UnicodeJsps directory to build the JSPs.
+Run `mvn package -DskipTests` in the repository directory to build the JSPs.
```
echeran, line 18:
```suggestion
Run `mvn -am -pl UnicodeJsps package -DskipTests` at the root directory of the repository to build the JSPs.
```

echeran (approved):
LGTM. The multiple extra fixes are appreciated. I left one minor suggested change.

echeran (approved):

markusicu:
very nice! merci beaucoup!

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeJsp.java
```diff
@@ -208,28 +208,39 @@ public static void getDifferences(
...
         // text = fromHTML.transliterate(text);
         String trimmed = text.trim();
-        if (trimmed.length() > 1) {
+        // Accept U+ notation and standalone hexadecimal digits, as well as a variety
+        // of hexadecimal character escape and numeric literal syntaxes.
+        Matcher matcher =
+                Pattern.compile(
+                                "(?:U\\+|\\\\[ux]\\{?|&#x|0x|16#|&H)?([0-9a-f'_]+)[\\};#]?",
```
srl295, line 218: Ideally this should go into a separate tool over in unicodetools/src/java/org/unicode somewhere and have its own tests. could be future.

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeJsp.java
srl295: nice

srl295 (commented):
good. has some unrelated improvements

Issue #327, opened 2022-09-03
add Public/<version>/ReadMe.txt into the repo?


markusicu:
For each data file folder that we track in the Unicode Tools repo, we include almost all of the files (with exceptions like a transmogrified form of the Unihan data) including ReadMe.txt files.

We are missing the UCD top-level readme: https://www.unicode.org/Public/15.0.0/ReadMe.txt

Consider adding it to the repo, so that we can track its contents and mostly publish the UCD from out of the repo.
Since the tools don't need access to multiple versions of this file at the same time, it could just go into unicodetools/data/ucd/ .

markusicu:
We could give it a different name inside .../ucd/dev/

Easier to just put it into unicodetools/data/ucd/

markusicu:
I am having second thoughts.
- This file is part of the set of UCD files for each version, so it does make sense to keep it in `dev` which would require a different filename.
- There is another ReadMe.txt file that is equally tied to the set of UCD files for each version, which means we need to get creative about filenames anyway: https://www.unicode.org/Public/zipped/15.0.0/ReadMe.txt

New proposal:
- Add https://www.unicode.org/Public/15.0.0/ReadMe.txt as .../data/ucd/dev/version-ReadMe.txt
- Add https://www.unicode.org/Public/zipped/15.0.0/ReadMe.txt as .../data/ucd/dev/zipped-ReadMe.txt

ok? @josh-hadley @echeran @macchiati @srl295

echeran:
Makes sense, sounds ok to me

josh-hadley:
no objections from me

PR #394, opened 2023-01-30
update several readmes for 15.1 alpha/beta

markusicu:
including a few that we do not plan to publish as part of the initial alpha review

New folder structure in Public for the alpha & beta:
https://github.com/unicode-org/unicodetools/blob/main/docs/data-workflow.md

These are fairly straight updates. Elsewhere we have an ongoing discussion about the different styles of readme files: https://github.com/unicode-org/properties/issues/8

This is also missing a couple of files. I will add those in another PR: https://github.com/unicode-org/unicodetools/issues/327

### DIFF
--- a/unicodetools/data/emoji/dev/ReadMe.txt
+++ b/unicodetools/data/emoji/dev/ReadMe.txt
@@ -1,21 +1,21 @@
 # Unicode Emoji
-# © 2022 Unicode®, Inc.
+# © 2023 Unicode®, Inc.
 # Unicode and the Unicode Logo are registered trademarks of Unicode, Inc. in the U.S. and other countries.
 # For terms of use, see https://www.unicode.org/terms_of_use.html
 
-This directory contains final data files for Unicode Emoji, Version 15.0
+This directory contains draft data files for Unicode Emoji, Version 15.1
 
-Public/emoji/15.0/
+Public/draft/emoji/
 
   emoji-sequences.txt
   emoji-zwj-sequences.txt
   emoji-test.txt
 
-The following related files are found in the UCD for Version 15.0
+The following related files are found in the UCD for Version 15.1
 
-Public/15.0.0/ucd/emoji/
+Public/draft/UCD/ucd/emoji/
 
   emoji-data.txt
   emoji-variation-sequences.txt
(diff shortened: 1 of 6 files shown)


unicodetools/data/emoji/dev/ReadMe.txt
```diff
@@ -1,21 +1,21 @@
...
 
   emoji-sequences.txt
   emoji-zwj-sequences.txt
   emoji-test.txt
 
-The following related files are found in the UCD for Version 15.0
+The following related files are found in the UCD for Version 15.1
 
-Public/15.0.0/ucd/emoji/
+Public/draft/UCD/ucd/emoji/
```
echeran, line 16: observation: it's curious how the segment `15.0.0` is being replaced by `draft/UCD`, given that `draft` seems like the version string that corresponds to `15.0.0`. In other words, it's a little asymmetrical, so slightly confusing, but not ultimately harmful.

unicodetools/data/ucd/dev/emoji/ReadMe.txt
```diff
@@ -1,22 +1,22 @@
...
+# Date: 2023-01-30
+# © 2023 Unicode®, Inc.
 # Unicode and the Unicode Logo are registered trademarks of Unicode, Inc. in the U.S. and other countries.
 # For terms of use, see https://www.unicode.org/terms_of_use.html
 
-This directory contains final data files for Unicode, Version 15.0
+This directory contains draft data files for Unicode, Version 15.1
 
-Public/15.0.0/emoji/
+Public/draft/UCD/ucd/emoji/
```
echeran, line 9: Ditto. Although I know we've discussed the location of emoji data and how it's a strange case given certain data that's in UCD vs. stored separately.

echeran (approved):

unicodetools/data/emoji/dev/ReadMe.txt
markusicu: Yes, in the new scheme for alpha & beta (only), we have one single "draft" folder that contains the other pieces, and each of those without a version level. Documented in data-workflow.md.

PR #396, opened 2023-01-31
update UAX-45 data files

markusicu:
@Tsengtsz updated these according to decisions in UTC-174.


Ken-Whistler (approved):
Change to USourceData.txt looks correct. I also checked the two pdf files, and the addition of UTC03336 to each also looks fine.

PR #397, opened 2023-02-01
Emoji 15.1 alpha data

nedley:
Resolves [174-A61](https://www.unicode.org/cgi-bin/GetL2Ref.pl?174-A61) and #152.

PR #402, opened 2023-02-03
alpha data workflow update: emoji files deduped, zip draft folder

markusicu:
@nedley de-duplicated two emoji files in PR #397 so the copy-alpha script need not delete those any more.

Workflow: Don't assume that the script is run on the unicode.org server; send Rick a zip file with the contents of the “draft” target folder.


nedley (approved):
It’s true, I did that.

rickmcgowan (approved):

PR #403, opened 2023-02-03
UCA 15.1

markusicu:
From @Ken-Whistler:

## UCA 15.1.0: Drop for first delta

I've done the code and input file updates, and have generated an initial
drop for you for UCA 15.1.0.

For this delta, I just created a baseline by adding the 5 new
ideographic description characters for 15.1 and turning the crank.

## UCA 15.1.0: Drop for second delta

For [173-A15](https://www.unicode.org/cgi-bin/GetL2Ref.pl?173-A15), moving U+10A7F to sort in the punctuation range.


markusicu:
I just realized that this PR here was still pending but I have since merged the changes for Ken's delta 3, see https://github.com/unicode-org/unicodetools/pull/460.
Those subsume the changes here but we won't see the deltas separately in the main branch history :-(

PR #404, opened 2023-02-07
more updates for 15.1 data workflow

markusicu:
More docs changes. For issue #172 I looked through the docs and made changes as necessary.

I did not read every word on every page... after merging this PR, I intend to check off the items in #172, close that issue, and when we need to make more changes, we do.

### DIFF
--- a/docs/inputdata.md
+++ b/docs/inputdata.md
@@ -6,7 +6,9 @@ Starting with Unicode 15, we are developing most of the Unicode data files
 in this Unicode Tools project, and publish them to the Public folder
 only for alpha/beta/final releases.
 That is, we are reversing the flow of files.
-(See [issue #144](https://github.com/unicode-org/unicodetools/issues/144).)
+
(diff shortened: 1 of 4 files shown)


docs/inputdata.md
```diff
@@ -15,6 +17,34 @@ and we continue to ingest them as before.
 
 ## Source Files
 
+*Starting with Unicode 15.1, the “source of truth” for most data files is in the repo,
+and most of this section is obsolete. See [data workflow](data-workflow.md).
+The biggest exception is Unihan.zip, which we don't track in the repo; see the Unihan section below.
+Also, it's still useful to delete the BIN files/folders after changing data files.*
+
+### Unihan
```
markusicu, line 25: FYI: I moved this section, unchanged, up from below. This Unihan section is still relevant. The other ones only remain for historical information (nostalgia?).

macchiati (approved):

PR #406, opened 2023-02-08
publish alpha: everyone can read files, search dirs; zip files

markusicu:
Set file & dir permissions.
Add a command to zip the files, to remove a manual step.

### DIFF
--- a/pub/copy-alpha-to-draft.sh
+++ b/pub/copy-alpha-to-draft.sh
@@ -19,6 +19,12 @@ rm $DRAFT/UCD/ucd/zipped-ReadMe.txt
 mkdir -p $DRAFT/emoji
 cp $UNITOOLS_DATA/emoji/dev/* $DRAFT/emoji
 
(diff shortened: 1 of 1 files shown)


rickmcgowan (approved):
Probably OK in practice, but last I read the chmod documentation the "X" (capital X) only sets the "x" (execute/search) bits if the file already has some execute permission. I.e.: man chmod: " execute/search only if the file is a directory or already has execute permission for some user (X), " So it may not be certain to _only_ add "x" to non-directories and could add "x" for all to regular files that happen to already have some execute perms, e.g., for owner.

markusicu:
> ... man chmod: " execute/search only if the file is a directory or already has execute permission for some user (X), " So it may not be certain to _only_ add "x" to non-directories and could add "x" for all to regular files that happen to already have some execute perms, e.g., for owner.

As the docs say, it makes directories searchable, and it extends file executability for files that are executable for anyone.
The only files that should have any x bits are shell and Python scripts.

PR #407, opened 2023-02-09
Replace references to UTR #51 with UTS #51

nedley:
Fix code that generates UCD/StandardizedVariants.txt and emoji/emoji-sequences.txt.


josh-hadley (approved):
LGTM

PR #411, opened 2023-02-21
fix(build): jsp: fix checkout for push to GCR

srl295:
#271

understandably, they've removed v1.2 as a tag. But the SHA is still there

We need this version because we need the functionality to checkout a short hash (from the CLDR pom version)


eggrobin (approved):

srl295:
@eggrobin  PTAL https://unicode-jsps-staging-5ocgitonaa-uw.a.run.app/UnicodeJsps/

eggrobin (approved):

eggrobin:
> PTAL https://unicode-jsps-staging-5ocgitonaa-uw.a.run.app/UnicodeJsps/

Looks good.

PR #416, opened 2023-03-06
Myanmar Extended-C

eggrobin:
For unicode-org/properties#57, according to [L2/22-046](https://www.unicode.org/L2/L2022/22046-myanmar-numerals.pdf), per [171-C20](https://www.unicode.org/L2/L2022/22061.htm#171-C20):
> [[171-C20](https://www.unicode.org/cgi-bin/GetL2Ref.pl?171-C20)] Consensus: Accept 20 Myanmar numerals (U+116D0…U+116E3) for Eastern Pwo Karen and Pa’O in a new Myanmar Extended-C block, with range U+116D0..U+116FF for a future version of the standard. (Reference [L2/22-046](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/22-046))

The proposal did not supply information for LineBreak.txt, so I inferred that from the GC ([GC=Nd is almost always LB=NU](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3ANd%3A%5D&g=LB&i=)).

```
116D0..116E3   ; NU # Nd    [20] MYANMAR PAO DIGIT ZERO..MYANMAR EASTERN PWO KAREN DIGIT NINE
```

Generated files updated as follows.
```powershell
mvn compile exec:java '-Dexec.mainClass="org.unicode.text.UCD.Main"'  '-Dexec.args="build MakeUnicodeFiles"' -am -pl unicodetools  "-DCLDR_DIR=..\cldr\"  "-DUNICODETOOLS_GEN_DIR=Generated"  "-DUNICODETOOLS_REPO_DIR=."
cp .\Generated\UCD\15.1.0\* .\unicodetools\data\ucd\dev -recurse -force
rm unicodetools\data\ucd\dev\zzz-unchanged-*
rm unicodetools\data\ucd\dev\*\zzz-unchanged-*
rm .\unicodetools\data\ucd\dev\extra\*
rm .\unicodetools\data\ucd\dev\cldr\*
```

PR #417, opened 2023-03-06
Sidetic

eggrobin:
For unicode-org/properties#102, according to L2/23-019.
> [[174-C17](https://www.unicode.org/cgi-bin/GetL2Ref.pl?174-C17)] Consensus: Provisionally assign 29 code points U+10940..U+1095C in a new Sidetic block at U+10940..U+1095F, for 29 Sidetic characters as described in [L2/23-019](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/23-019) and section 1 of [L2/23-012](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/23-012).


unicodetools/src/main/java/org/unicode/props/UcdPropertyValues.java
```diff
@@ -1831,6 +1831,7 @@ public enum Script_Values implements Named {
         Math_Symbols("Zmth"),
         Emoji_Symbols("Zsye"),
         Other_Symbols("Zsym"),
+        Sidetic("Qabw"),
```
markusicu, line 1834: This file should ideally be generated from PVA.txt by GenerateEnums.java. (And then you need to reformat using `mvn spotless apply` ...)

https://github.com/unicode-org/unicodetools/blob/main/docs/newunicodeproperties.md#run-generateenumsjava

unicodetools/src/main/java/org/unicode/text/UCD/UCD_Names.java
```diff
@@ -592,6 +594,8 @@ public final class UCD_Names implements UCD_Types {
         // Unicode 15
         "Kawi",
         "Nagm",
+        // Unicode 16?
+        "Qabw",
```
markusicu, line 598: I would add a comment like `"Qabw" // Sidetic` while we use a private use code.

unicodetools/src/main/java/org/unicode/props/UcdPropertyValues.java
eggrobin: I was following https://github.com/unicode-org/unicodetools/blob/main/docs/build.md#new-blocks which starts from the Java rather than PVA; I guess that section is outdated?

markusicu: Maybe I am just outdated. I certainly don't remember all of the docs that we have...
Looking at that:

> Add to ShortBlockNames.txt (you need to know what the short name is, you can find it in PropertyValueAliases.txt)

This suggests that PVA.txt wants to be updated first.

My comment about how it *should* work is a wish for the future. Clearly you updated the seemingly-redundant internal file, and the tools added the block to PVA.txt. But what I *wish* for, in some future, is to clearly have PVA.txt be an input, and not to need SBN.txt.

> Add long & short names to UcdPropertyValues.java enum Block_Values

Again, clearly you can do this manually, but if you do it differently from GenerateEnums+spotless, then you should get a CI check failure.

eggrobin: > Again, clearly you can do this manually, but if you do it differently from GenerateEnums+spotless, then you should get a CI check failure.

Generally it sounds like we need a CI check for GenerateEnums, as you noted on #416. We can discuss that tomorrow night.

(FWIW spotless fails for me with an issue that looks like `https://github.com/google/google-java-format/issues/787`)

markusicu: > > Again, clearly you can do this manually, but if you do it differently from GenerateEnums+spotless, then you should get a CI check failure.
> 
> Generally it sounds like we need a CI check for GenerateEnums, as you noted on #416. We can discuss that tomorrow night.

We have one: https://github.com/unicode-org/unicodetools/blob/main/.github/workflows/cli-build-instructions.yml#L203

eggrobin: Ah right, it just runs after the UCA validation, which fails.

unicodetools/src/main/java/org/unicode/props/UcdPropertyValues.java
```diff
@@ -62,8 +62,7 @@ public enum Age_Values implements Named {
         V12_0("12.0"),
         V12_1("12.1"),
         V13_0("13.0"),
-        V13_1("13.1"), // TODO: there is no Unicode 13.1, see
-        // https://github.com/unicode-org/unicodetools/issues/100
+        V13_1("13.1"),
```
eggrobin, line 65: It looks like GenerateEnums + spotless removes this comment.

markusicu: Yes. Please restore it. Known issue.

markusicu:
Preliminary script code: Sidt

markusicu (commented):
lgtm

unicodetools/data/ucd/dev/LineBreak.txt
```diff
@@ -2748,6 +2748,7 @@ FFFD           ; AI # So         REPLACEMENT CHARACTER
 1091F          ; BA # Po         PHOENICIAN WORD SEPARATOR
 10920..10939   ; AL # Lo    [26] LYDIAN LETTER A..LYDIAN LETTER C
 1093F          ; AL # Po         LYDIAN TRIANGULAR MARK
+10940..1095C   ; AL # Lo    [29] SIDETIC LETTER N1..SIDETIC LETTER N29
```
Marcos323232, line 2753: Marcos32rocha32@gmail.com

Marcos323232 (approved):

markusicu (commented):
latest changes lgtm2

markusicu (approved):

PR #422, opened 2023-03-10
 Line breaking at orthographic syllable boundaries

eggrobin:
Per L2/23-072: L2/22-080 without the change to the line breaking class of ◌, giving the Brahmi Number Joiner lb=GL rather than lb=WJ, and amending LB28b to special-case ◌.

[[175-C27](https://www.unicode.org/cgi-bin/GetL2Ref.pl?175-C27)] Consensus: Add line breaking classes AF, AK, AP, AS, VI, and VF, as well as a new line breaking rule LB 28b, and change Line_Break property values, as described in [L2/23-072](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/23-072).

[[175-A81](https://www.unicode.org/cgi-bin/GetL2Ref.pl?175-A81)] Action Item for Robin Leroy, PAG: Update LineBreak.txt and PropertyValueAliases.txt as described in [L2/23-072](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/23-072). For Unicode Version 15.1.


eggrobin:
Norbert had a few comments over email:

- [ ] Peter hinted that we need to update the header of LineBreak.txt, which indeed has a list of Line_Break property values that now needs a few more.

- [ ] Apparently some tool generates a lot of test cases in LineBreakTest.txt for corner cases which are unlikely to occur but should nevertheless be handled in a well-defined way. I’m more interested in test cases for actual text, starting with the ones in the Example section of L2/22-080R2. How would I add those?

- [ ] Something is reordering entries in ScriptExtensions.txt – that should be prevented.

- [ ] LineBreakTest.html shows lots of changes in the source, but has neither the new classes nor the new rule.

macchiati:
For the header of generated/formatted data files, you generally modify
MakeUnicodeFiles.txt. But in the case of linebreak I think it is all
maintained manually.

GenerateBreakTest is used to generate the segmentation tests. It uses
Segmenter.java, with 2 different sets of rule files:

   - SegmenterCldr.txt
   - SegmenterDefault.txt

These need to be in sync, except for the CLDR additions.

For the test cases, many are generated with various sample characters, one
chosen from each property value (I think with some characters for the rules
that use other Unicode properties*).

The extra ones are in the subclasses of XGenerateBreakTest, such as

GenerateLineBreakTest extends XGenerateBreakTest

This also generates the .html files with the processed rules and examples.

So for extra examples for LB, you'd add strings there.

LineBreakTest.html. The problem is that if you change the sets of
characters with a property, it can change the default character that is
chosen. I have long had on my (long) TODO list to correct that. My thought
was to write out the list of representative characters used in version X.
When generating the representative characters for X+1, if one of the X
representative characters fits the bill, use it.

* Note: because some of the files use properties outside of the main one,
ideally one would create a full partition of the code space, whereby all
the characters in a given partition behave the same in the algorithm. So,
for example, we would get from OP two partitions
[OP-[\p{ea=F}\p{ea=W}\p{ea=H}]] and [OP&[\p{ea=F}\p{ea=W}\p{ea=H}]]

eggrobin:
Apparently I forgot to copy this line of Norbert’s email:
- [ ] SentenceBreakProperty.txt and WordBreakProperty.txt lost a number of entries for digits in Brahmic scripts. Are these entries derived from line breaking data?

They are, https://www.unicode.org/reports/tr29/#Table_Sentence_Break_Property_Values and https://www.unicode.org/reports/tr29/#Table_Word_Break_Property_Values.

This means we need a matching amendment to #&zwnj;29.
Right now Numeric is defined as follows for Word_Break:

Numeric | Line_Break = Numeric or any of the following: U+FF10 (０) FULLWIDTH DIGIT ZERO..U+FF19 (９) FULLWIDTH DIGIT NINE and not U+066C ( ٬ ) ARABIC THOUSANDS SEPARATOR
-- | --

and as follows for Sentence_Break:

Numeric | Line_Break = Numeric or any of the following: U+FF10 (０) FULLWIDTH DIGIT ZERO..U+FF19 (９) FULLWIDTH DIGIT NINE
--|--

[`Line_Break = Numeric or any of the following: U+FF10 (０) FULLWIDTH DIGIT ZERO..U+FF19 (９) FULLWIDTH DIGIT NINE` can be turned into `Line_Break = Numeric or General_Category = Nd`](https://util.unicode.org/UnicodeJsps/unicodeset.jsp?a=[[:lb=NU:][:Nd:]]&b=[:sb=NU:]), and that would be preserved by the Line_Break change.

macchiati:
Agreed, they are identical.

https://util.unicode.org/UnicodeJsps/unicodeset.jsp?a=[\p%7BLine_Break=Numeric%7D[\uFF10-\uFF19]]&b=[\p%7BLine_Break=Numeric%7D\p%7BGeneral_Category=Nd%7D]

macchiati (commented):
Looking good

unicodetools/data/ucd/dev/DerivedAge.txt
```diff
@@ -1,5 +1,5 @@
 # DerivedAge-15.1.0.txt
-# Date: 2023-04-26, 18:59:33 GMT
+# Date: 2023-05-02, 14:11:40 GMT
```
markusicu, line 2: It would be nice not to check in files where only the time stamp changes.

markusicu (dismissed):
(partial rs)lgtm except for files with only time stamp changes

markusicu:
oh, "This branch cannot be rebased due to conflicts"

eggrobin:
> This branch cannot be rebased due to conflicts

Ah, I understand now why you see conflicts that I don’t: you have the selector set to _rebase and merge_, not _squash and merge_. This probably shouldn’t be rebased anyway.

markusicu:
> > This branch cannot be rebased due to conflicts
> 
> Ah, I understand now why you see conflicts that I don’t: you have the selector set to _rebase and merge_, not _squash and merge_. This probably shouldn’t be rebased anyway.

Why would that make a difference? Either way the PR proposes to make the same set of changes, and if there is a conflict, then there's a conflict. (Not criticizing, just trying to understand.)

markusicu (approved):
lgtm
and thanks for reverting time-stamp-only files!

PR #424, opened 2023-03-23
Don’t manually comment a generated file.

### LINKED ISSUE #100: Whither 13.1.0?
https://github.com/unicode-org/unicodetools/pull/44 introduces a "13.1.0" data subdirectory. Perhaps it means Emoji 13.1.0 as opposed to Unicode 13.1.0.  Revisit whether 13.1.0 can be removed from the data directory.

Should the emoji version be kept separately from the Unicode version?

eggrobin:
Someday, and that day may never come, I will call upon you to fix #100.

But until that day, I would like manually reverting that line not to be a part of the process for adding new scripts, blocks, or other property values.


markusicu (approved):
lgtm tnx

PR #425, opened 2023-03-24
Add CI workflow to prevent Maven cache from getting evicted

### LINKED ISSUE #421: CI workflow to keep Maven artifact cache fresh
In #418, Unicode Tools' CI on Github Actions is experiencing the same problem as the CI for ICU's Maven build, which is that we have ephemeral download errors from Maven Central (for some reason on the same artifact), and the cache TTL for Github Actions is only 7 days.

We can copy [my workaround in ICU](https://github.com/unicode-org/icu/pull/2281), which is just to make use of the cache regularly with period < TTL.

echeran:
Fixes #421 

This is basically a copy of the [ICU CI workflow](https://github.com/unicode-org/icu/blob/main/.github/workflows/cache_retain.yml) that does the same. Except that instead of doing _something_, I think doing _nothing_ is sufficient because the Github's cache action only cares about the job success status.

cc @eggrobin

### DIFF
--- a/.github/workflows/cache_retain.yml
+++ b/.github/workflows/cache_retain.yml
@@ -55,4 +55,4 @@ jobs:
       # invoked and the job completes successfully.
       #
       # It seems sufficient for the CI job here to finish successfully by
-      # trivially run nothing.
+      # trivially running nothing.
--- a/.github/workflows/cache_retain.yml
+++ b/.github/workflows/cache_retain.yml
@@ -1,0 +1,58 @@
+# Copyright (C) 2023 and later: Unicode, Inc. and others.
+# License & terms of use: http://www.unicode.org/copyright.html
+#
+# This workflow is designed to keep specific caches on the main
+# branch from getting evicted according to the Github Actions policy
+# (currently in 2023: 7 days) in cases where the cost to construct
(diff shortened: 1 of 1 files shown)


srl295 (dismissed):

.github/workflows/cache_retain.yml
```diff
@@ -0,0 +1,58 @@
...
+          mkdir -p unicodetools/mine/Generated/BIN
+
+      # Omitting running any build instructions.
+      #
+      # The point is to touch the cache before the TTL causes it to get evicted.
+      # For Github Actions, the cache is preserved only when the cache action is
+      # invoked and the job completes successfully.
+      #
+      # It seems sufficient for the CI job here to finish successfully by
+      # trivially run nothing.
```
markusicu, line 58:
```suggestion
      # trivially running nothing.
```

markusicu (approved):
lgtm tnx

However, there are failing CI checks, apparently from the github.com certificate change. How do we get around those?

srl295:
requested a rerun

markusicu (approved):

srl295 (approved):

Issue #426, opened 2023-03-30
Get rid of casefolding using getCase


eggrobin:
It is based on a definition of case folding which is independent of the normative one (and inconsistent, at least for simple case folding), which is terrifying.

See https://github.com/unicode-org/unicodetools/pull/423#discussion_r1153459329.

In order to stop the bleeding I will be adding a consistency check to GenerateCaseFolding.txt, but we should just stop having independent definitions.

PR #427, opened 2023-03-30
Fix case folding using getCase

eggrobin:
Add a consistency check.

#426 for follow-up.

This one is going to be easier to review without the spaces: https://github.com/unicode-org/unicodetools/pull/427/files?w=1.


eggrobin:
Note: running `build MakeUnicodeFiles` prior to 7abdb62 failed, as it should, with the following error:
```
java.lang.AssertionError: UCD.getCase("\u0130", UCD.SIMPLE, UCD.FOLD)="i", should be "İ" per CaseFolding.txt
```

macchiati (approved):

PR #428, opened 2023-03-31
Move GCB=Prepend out of WB=Format

eggrobin:
Mostly into WB=Numeric.

See L2/23-079 item 𒇧𒇧, 174-C𒇧𒇧, 174-A𒇧𒇧.


markusicu (approved):
lgtm nice!

Issue #53, opened 2021-04-30
UnicodeSet util: Extended_Pictographic missing unassigned code points


markusicu:
https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AExtPict%3A%5D%26%5B%3ACn%3A%5D&g=&i= comes up empty but should show more than 1000 unassigned code points such as the following.

https://www.unicode.org/Public/UCD/latest/ucd/emoji/emoji-data.txt
```
1F0D0..1F0FF  ; Extended_Pictographic# E0.0  [48] (🃐..🃿)    <reserved-1F0D0>..<reserved-1F0FF>
...
1F8AE..1F8FF  ; Extended_Pictographic# E0.0  [82] (🢮..🣿)    <reserved-1F8AE>..<reserved-1F8FF>
...
1FC00..1FFFD  ; Extended_Pictographic# E0.0[1022] (🰀..🿽)    <reserved-1FC00>..<reserved-1FFFD>
```
@macchiati FYI

macchiati:
Thanks. I need to fix the underlying code. There is a property flag that says "I have all the same values for unassigned code points", in place for performance. That flag is manually set, but should be computed at build time instead.

markusicu:
Mark said in https://github.com/unicode-org/unicodetools/issues/97#issuecomment-883030395 that this issue here is a duplicate of the newer one, and the newer one has more information.

PR #435, opened 2023-04-14
Four Arabic Pegon Characters

eggrobin:
[[172-C3](https://www.unicode.org/cgi-bin/GetL2Ref.pl?172-C3)] Consensus: Accept the following four characters for a future version of the standard, with property values as given in [L2/22-116](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/22-116):

 	U+0897 ARABIC PEPET
	U+10EC2 ARABIC LETTER DAL WITH TWO DOTS VERTICALLY BELOW
	U+10EC3 ARABIC LETTER TAH WITH TWO DOTS VERTICALLY BELOW
	U+10EC4 ARABIC LETTER KAF WITH TWO DOTS VERTICALLY BELOW
(Reference: [L2/22-116](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/22-116) and section 4b of [L2/22-128](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/22-128))


roozbehp (changes requested):
ArabicShaping.txt has issues with Joining_Type of the dal-like character and the schematic names of all three.

roozbehp:
> I think it should be:
> 
> ```
> # Arabic Extended-D Characters
> 10EC2; DAL WITH TWO DOTS VERTICALLY BELOW; R; DAL
> 10EC3; TAH WITH TWO DOTS VERTICALLY BELOW; D; TAH
> 10EC4; KAF WITH TWO DOTS VERTICALLY BELOW; D; KAF
> ```

Yes, but with "TWO" replaced by "2".

unicodetools/data/ucd/dev/UnicodeData.txt
```diff
@@ -2123,6 +2123,7 @@
 088E;ARABIC VERTICAL TAIL;Lo;0;AL;;;;;N;;;;;
 0890;ARABIC POUND MARK ABOVE;Cf;0;AN;;;;;N;;;;;
 0891;ARABIC PIASTRE MARK ABOVE;Cf;0;AN;;;;;N;;;;;
+0897;ARABIC PEPET;Mn;0;NSM;;;;;N;;;;;
```
markusicu, line 2126: In the [issue](https://github.com/unicode-org/properties/issues/54#issuecomment-1518308495), @LornaSIL writes:
> the proposal document has a combining class of 0 for this character. I would think it should be 230. Maybe it's already been fixed?

Looking at [L2/22-116](https://www.unicode.org/L2/L2022/22116-four-pegon-chars.pdf), the pepet indeed looks like a combining mark above which is normally ccc=230.

The proposal disunifies the Pepet from U+0653 Arabic Maddah Above which has ccc=230.

@roozbehp @Ken-Whistler @dwanders-A is there a reason for the Pepet to have ccc=0?

Ken-Whistler: I can't see a good reason for 0897 ARABIC PEPET to have ccc=0 instead of ccc=230. It doesn't seem to be explicitly discussed in the SAH discussion thread or recommendations. Note that PEPET is *not* actually a disunification of 0653. It is a separate tashkil for Pegon which some people were using a maddah to represent, because that was the closest in appearance. Most newly encoded tashkil above are just ccc=230, however, so that might be right for ccc. Roozbeh and Lorna would have to clarify whether there is any conceivable interaction with the MSM in UTR #53. Note that the AMTRA algorithm moves MCM characters away from madda and ahead of the reordered shadda and other tashkil. Likely not an issue for PEPET, but I don't fully understand all of this.

eggrobin: FWIW: The set [`[:Arab:]&[:Mn:]&[:ccc=0:]`](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AArab%3A%5D%26%5B%3AMn%3A%5D%26%5B%3Accc%3D0%3A%5D&g=&i=) (Arabic nonspacing marks with ccc=0) is empty. Maybe we should make that an invariant test (probably extended to the scripts outside chapters 12 through 17).

markusicu: > FWIW: The set [`[:Arab:]&[:Mn:]&[:ccc=0:]`](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AArab%3A%5D%26%5B%3AMn%3A%5D%26%5B%3Accc%3D0%3A%5D&g=&i=) (Arabic nonspacing marks with ccc=0) is empty. Maybe we should make that an invariant test (probably extended to the scripts outside chapters 12 through 17).

sgtm, at least as a trip-wire test that triggers a review, not an encoding principle (yet)

roozbehp: 1) PEPET should have ccc=230. 

2) We will have a ccc=0 combining mark in Unicode 16.0 for the first time, namely U+10EFC ARABIC COMBINING ALEF OVERLAY.

3) I think these pull requests follow the proposals too blindly. That's not what we had been doing in the past. In the past, when assigning properties to new characters, unless there was an explicit note from UTC about a certain property, we tried to assign properties carefully ourselves instead of looking at what the proposal had. I have been personally doing that for Indic and Arabic properties, as well as ScriptExtensions. I double check with the proposal after deciding the property value, and I found that several times the proposal had made mistakes. Most of the time, the error was so obvious I didn't even need to raise it with anybody, I just assigned the correct property...

eggrobin: > PEPET should have ccc=230.

Thanks; I will fix that in this PR, and mention it in unicode-org/properties#115.

> We will have a ccc=0 combining mark in Unicode 16.0 for the first time, namely U+10EFC ARABIC COMBINING ALEF OVERLAY.

Good to know; it may still be worth explicitly listing such things as exceptions in an invariant test, since they are not the norm.

> I think these pull requests follow the proposals too blindly. That's not what we had been doing in the past. In the past, when assigning properties to new characters, unless there was an explicit note from UTC about a certain property, we tried to assign properties carefully ourselves instead of looking at what the proposal had.

My understanding is that in the future, since there is now a PAG review of encoding proposals once the SAH thinks they should be recommended for provisional assignment, these adjustments will get done at that time, and will be reflected in amendments to the proposal.

The proposals in the 16.0 pipeline predate that process, so we will have to review the properties as we go along. I had originally created this draft pull request and the others in https://github.com/unicode-org/unicodetools/labels/data-for-new not so much for that, but to sort out some concerns Markus had about merging multiple changes to the data files and the data file generation process.

It seems that these pull requests have been very effective at getting additional eyeballs on the data files, which is a good thing: we do not want more BRAKCETs :-)

roozbehp:
See also https://github.com/unicode-org/unicodetools/issues/454 for catching issues in the schematic name.

eggrobin:
Checked consistency with Ken’s UnicodeData-16.0.0d6.txt and LineBreak-16.0.0d2.txt.

roozbehp (approved):
LGTM

PR #442, opened 2023-04-18
5 Smalltalk symbols

eggrobin:
[[170-C17](https://www.unicode.org/cgi-bin/GetL2Ref.pl?170-C17)] Consensus: UTC accepts the following 5 Smalltalk symbols, as documented in [L2/21-234](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/21-234), for a future version of the standard:

U+1CEB0 HORIZONTAL ZIGZAG LINE
U+1CEB1 KEYHOLE
U+1CEB2 OLD PERSONAL COMPUTER WITH MONITOR IN PORTRAIT ORIENTATION
U+1CEB3 BLACK RIGHT TRIANGLE CARET
U+1F8B2 RIGHTWARDS ARROW WITH LOWER HOOK


eggrobin:
Checked consistency with Ken’s UnicodeData-16.0.0d6.txt and LineBreak-16.0.0d2.txt; some LineBreak ranges merge with others from other proposals, but the assignments appear consistent.

unicodetools/data/ucd/dev/extracted/DerivedLineBreak.txt
```diff
@@ -1430,6 +1430,7 @@ FFED..FFEE    ; AL # So   [2] HALFWIDTH BLACK SQUARE..HALFWIDTH WHITE CIRCLE
 1BC80..1BC88  ; AL # Lo   [9] DUPLOYAN AFFIX HIGH ACUTE..DUPLOYAN AFFIX HIGH VERTICAL
 1BC90..1BC99  ; AL # Lo  [10] DUPLOYAN AFFIX LOW ACUTE..DUPLOYAN AFFIX LOW ARROW
 1BC9C         ; AL # So       DUPLOYAN SIGN O WITH CROSS
+1CEB0..1CEB3  ; AL # So   [4] HORIZONTAL ZIGZAG LINE..BLACK RIGHT TRIANGLE CARET
```
nedley, line 1433: I’m guessing AL was chosen based on surrounding blocks but I think this should be ID to match 1F8B2.

eggrobin: I forget how I chose it; So is 54% lb=AL and 37% lb=ID without much in the way of obvious patterns.
One pattern I _am_ seeing is that the arrows tend to be lb=**AL**, except for U+1F8B0 and U+1F8B1:
https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3Aname%3D%2F%5CbARROW%5Cb%2F%3A%5D&g=lb&i=

 So maybe we should recommend an LB change for those two and make the new one AL?

Looking in the mapping tables at some of the related already-encoded characters that are not otherwise widely used is not that much more enlightening:
- ☞ is ID;
- ⏿ is AL;
- ⦂ is AL.

For their meaning in Smalltalk see the screenshot below from http://www.bitsavers.org/pdf/xerox/parc/techReports/SSL-76-6_Smalltalk-72_Instruction_Manual_Mar76.pdf, p. 49.
More than you likely wanted to know about Smalltalk-72

For context, that range contains the following characters:

U+1CEB0 HORIZONTAL ZIGZAG LINE
U+1CEB1 KEYHOLE
U+1CEB2 OLD PERSONAL COMPUTER WITH MONITOR IN PORTRAIT ORIENTATION
U+1CEB3 BLACK RIGHT TRIANGLE CARET

nedley: OK, considered as operators I will accept AL (or possibly AI).

eggrobin: Changed the new arrow to LB=AL; the two ID ones in that block, as well as the block default, will be dealt with separately in #580 if the UTC approves of LB=AL for them.

unicodetools/data/ucd/dev/VerticalOrientation.txt
```diff
@@ -2178,6 +2178,7 @@ FFFC..FFFD     ; U  # So     [2] OBJECT REPLACEMENT CHARACTER..REPLACEMENT CHARA
 1BC9D..1BC9E   ; R  # Mn     [2] DUPLOYAN THICK LETTER SELECTOR..DUPLOYAN DOUBLE MARK
 1BC9F          ; R  # Po         DUPLOYAN PUNCTUATION CHINOOK FULL STOP
 1BCA0..1BCA3   ; R  # Cf     [4] SHORTHAND FORMAT LETTER OVERLAP..SHORTHAND FORMAT UP STEP
+1CEB0..1CEB3   ; R  # So     [4] HORIZONTAL ZIGZAG LINE..BLACK RIGHT TRIANGLE CARET
```
nedley, line 2181: Since this file says code points “that are not listed[…] are given the value R”, we may want to _not_ add new characters since that feels like we are saying we have some evidence of how they ought to be used in vertical typesetting when we don’t.

markusicu: Looking at the current version of the file, it looks like it lists every character that is assigned or have a value different from R. That is, it seems to omit only those R code points that are unassigned. And, this file gets regenerated, so omitting certain characters would require both a policy change and a code change.

We should remain consistent with current practice until we (PAG+UTC) decide otherwise.

eggrobin: While this is unwritten in the file, all assigned code points are automatically listed, even when we don’t have a clue about their vertical typesetting. I don’t think we want to change that—if we want to have information in here we want it to part of the property. Maybe we should have a new value No_Clue, but that seems like another can of worms.

nedley (approved):

PR #444, opened 2023-04-18
Ignore conflict markers in ConvertUCD

eggrobin:
We might need to do that for other parsers at some point; this is the only one that posed problems when merging #442 and #441.

Note that after merging and before running `MakeUnicodeFiles`, the Derived files need to be reset to a non-conflicting state, otherwise the ranges either side of the conflict markers overlap. UnicodeData is obviously immune to that.

This will probably be a problem for some source files eventually (LineBreak or Script), so I am leaving #419 open; we’ll cross that bridge when we get to it.

PR #449, opened 2023-04-20
Regenerate UnicodeData in MakeUnicodeFiles; fix a bug in Unicode_1_Name

eggrobin:
This turned out much easier than expected, because Mark had done half the work more than 10 years ago (I can’t tell how much more, git history does not go further for this repository; nor do I know why Mark stopped halfway through). This should make the processing of new proposals easier: no need to manually merge-sort the lines from the proposal, and with #444, conflict markers will be automatically eliminated when regenerating the UCD.

This also fixes a (fairly uninteresting) bug: the Unicode_1_Name property (obsolete, but neither deprecated nor stabilized, and importantly not blanked like ISO_Comment was) was incorrectly handled by the tools, which relied on an ICU API for it (ICU deprecated that API in [ICU-9013](https://unicode-org.atlassian.net/browse/ICU-9013), https://github.com/unicode-org/icu/commit/c39e5af48c1008d9d88ec5d7c0cbc9fddef2a714).

The new options for BagFormatter should allow us to generate files without spaces either side of the semicolon, like the 15.0 and earlier LB and EAW, should that be needed.

The function `dumbFraction` is extended to guess the correct denominator based on the name; if we want the fraction fields in DerivedNumericValues to match UnicodeData for the Meroitic cursive twelfths, we could pass the name there too; for now I am leaving the data files unchanged.

### DIFF
--- a/unicodetools/data/ucd/dev/EastAsianWidth.txt
+++ b/unicodetools/data/ucd/dev/EastAsianWidth.txt
@@ -1,11 +1,11 @@
 # EastAsianWidth-15.1.0.txt
-# Date: 2022-11-16, 09:33:27 GMT [KW, LI]
-# © 2022 Unicode®, Inc.
+# Date: 2023-04-19, 18:36:42 GMT
+# © 2023 Unicode®, Inc.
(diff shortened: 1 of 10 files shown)


macchiati (approved):
Looks great, thanks!

PR #455, opened 2023-05-01
NamesList.html from Ken jan19

markusicu:


Ken-Whistler (approved):
Looks correct. Note that the massive difference from 15.0 is partly because of the HTML 4.01 -> 5 conversion, and partly because of the need to reconcile the exact format of the BNF against changes made in Asmus' separate copy of this documentation. I've checked this validates o.k.

PR #462, opened 2023-05-06
Emoji 15.1 beta

nedley:
- Programmatically flip images for directional sequences
- Fix three family sequence names
- Recategorize of gender-inclusive families as symbols
- Update proposal data

I am waiting on three minor image updates before I can generate CLDR data and build the charts.


macchiati (approved):
Looks great! I can try out the files next week.

Issue #95, opened 2021-07-15
Fix emoji chart tofu


nedley:
Charts like https://unicode.org/emoji/charts-14.0/emoji-released.html use U+1F3FF to denote a skin-tone modifier, but it seems to be rendered as tofu by Chrome.

markusicu:
Could this be caused by whatever font is set in the styles/CSS? In https://github.com/unicode-org/cldr/blob/main/common/collation/root.xml `& [last primary ignorable]<<*🦰🦱🦳🦲🏻🏼🏽🏾🏿` it shows up fine on a browser that shows tofu on the emoji chart page.

Ned thinks it might have to do with what character precedes it.

Might be safest to just use an embedded color swatch image.

josh-hadley:
I wonder if some minor modifications to the CSS might fix this? I tried hacking/overriding the resolved `font-family` in Chrome (Mac), adding `'Apple Color Emoji'` at the end on one of the places this happens, and that seemed to do the trick. See attached screen captures. It would be nice not to have to embed an image if it can be avoided.

![before](https://user-images.githubusercontent.com/51727949/185441725-5eb34329-3595-4a0c-b459-41edb480711f.png)
![after](https://user-images.githubusercontent.com/51727949/185441740-3d1076dd-4f9c-4e54-8dff-5612f9ba4748.png)

nedley:
The problem is that this appears to be a cross-platform issue with Chrome, so I would like to avoid platform-specific fixes if at all possible. And if it requires a color swatch I would use pure CSS rather than an image.

josh-hadley:
Understood and I agree it would be better to do something that isn't platform-specific (_and_ is not an image).

I did a little more tinkering and it does indeed look like its the preceding character that triggers this. Actually, looks like _any_ preceding character that is not also an emoji. It seems like once Chrome gets a "hit" in the font fallback scheme, it stops falling back and uses the last font found to render the remainder of the string within the same block-level (?) element. So for example, `<p>&#x1F3ff;</p>` will render fine, but `<p>a &#x1F3ff;</p>` will fail on the skintone character because it tries to use the same font that it found for `a`. Similarly, `<div><span>a &#x1F3ff;</span><span>&#x1F3ff;</span></div>` fails on both because they're in the same block-level element (outer `<div>`). So for the example I used in the screen captures above, it wouldn't be fixed by simply moving the skintone character out of the `<span>`. It would need to be moved to a new `<div>` or other block-level element.

[...]

PR #463, opened 2023-05-08
Fix proposals for sequences containing U+FE0F

nedley:
While reviewing charts I discovered a longstanding issue with the one for proposals “by emoji”: it was omitting all sequences containing U+FE0F due to an overzealous skeleton comparison. For example, if you look at https://www.unicode.org/emoji/charts-15.0/emoji-proposals.html the proposal for heart on fire is listed but that is the only reference to that proposal on the page, plus the pilot sequences are missing (among others).

This also applies the fix for #95 to the proposals chart.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/tools/emoji/ChartUtilities.java
+++ b/unicodetools/src/main/java/org/unicode/tools/emoji/ChartUtilities.java
@@ -153,8 +153,6 @@ public class ChartUtilities {
     }
 
     public static String htmlSpanForSkintone(String modifier) {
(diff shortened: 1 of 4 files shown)


macchiati (approved):

nedley:
@markusicu @srl295 The “license/cla” check appears to be stalled, how do I kick it?

srl295:
@nedley  the URL for recheck is `http://cla-assistant.io/check/unicode-org/unicodetools?pullRequest=463` – you can change the repo and PR # to taste.  On success you get dumped back to the github homepage.

PR #466, opened 2023-05-09
NamesList.html from Ken may09

markusicu:
From Ken:

This updated a bullet that explains the lookbehind logic for
intercepting various standards identifiers and preventing four digit
identifiers for them (e.g. ISO 6429) from turning into Unicode
characters in the output line. Asmus has added detection of IEEE, KSC X,
and GB to the list that is supported for this special processing.


Ken-Whistler (approved):
Looks as expected.

asmusf:
Ken found another typo that I fixed in my own copies of the syntax description.

PR #467, opened 2023-05-10
UCA 15.1d4: secondary diffs for fancy quotes

markusicu:
From Ken:

I've implemented the changes proposed in L2/23-016 (see UTC-174-C4 and UTC-174-A19).

I used the PUA range F8F9..F8FC to add the 4 additional variant
secondaries. With that choice, the diff for the input file is a nice,
concise change, easy to understand. The diff for decomps.txt is also
quite straightforward, contained, and self-explanatory.

The impact on allkeys.txt is well-contained, but not diffable, because
it removed a bunch of primary weights for the quotation marks, and then
shifted all the weights that followed. However, the significant change
for the output can be seen in lines 9629..9642, where all the secondary
variant weighted quotation marks show up in allkeys.txt, along with
geresh and gershayim. The entries look a bit strange in place there,
because they consist of variables (starred) followed by the
non-ignorable secondary weight as the second element. These are unusual
in allkeys.txt, but not unique. The one prior example we had is U+2ADC
FORKING, which has a canonical decomposition to a (variable) symbol plus
a combining mark. Note that I arbitrarily chose to assign F8FC (weight =
0000.0124.0004) for both geresh and gershayim, so they would both show
up with identical variant secondary weights in the output.

There are corresponding changes to the sifter code, but those are also
very contained and comprehensible.

I also annotated the entries for the 4 new PUA variant secondaries in
the unidata.txt file to warn that any change to them would require a
corresponding (careful) change to the sifter code, so future maintainers
don't mess them up by accident.


Ken-Whistler (approved):
Changes are as expected, both for the sifter code and the unidata and decomps files.

Issue #386, opened 2023-01-17
invariant: stability of simple case mappings


markusicu:
In L2/23-008 item 1.2, based on https://github.com/unicode-org/properties/issues/63, we ask for “a new derived property NFKC_SimpleCasefold”, and that “the Case Folding and Case Pair stability policies be extended to simple as well as full case transformations”.

Add invariant tests for these stability policies.

PR #469, opened 2023-05-12
New stability policy, and stability policies from the future

### LINKED ISSUE #385: invariant: Default_Ignorable_Code_Point changes do not affect NFKC_Casefold
In L2/23-008 item 1.1, based on https://github.com/unicode-org/properties/issues/43, we ask to “document for NFKC_Casefold that if the Default_Ignorable_Code_Point value changes for an existing XID_Continue that its NFKC_Casefold mapping must not change.”

Add an invariant test for this.

eggrobin:
Fix #385.

Fix most of #386, except for the part about case mapping, where we don’t yet have an invariant test for the full version either.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
@@ -354,7 +354,7 @@ public class TestUnicodeInvariants {
         }
 
         private static final UnicodeSet PROPCHARS =
-                new UnicodeSet("[a-zA-Z0-9\\:\\-\\_\\u0020\\p{pattern white space}]");
+                new UnicodeSet("[a-zA-Z0-9.\\:\\-\\_\\u0020\\p{pattern white space}]");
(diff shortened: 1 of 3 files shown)


unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -354,7 +354,7 @@ enum Type {
         }
 
         private static final UnicodeSet PROPCHARS =
-                new UnicodeSet("[a-zA-Z0-9\\:\\-\\_\\u0020\\p{pattern white space}]");
+                new UnicodeSet("[a-zA-Z0-9.\\:\\-\\_\\u0020\\p{pattern white space}]");
```
eggrobin, line 357: Note: this is so you can use something like `U5.2.0:NFKC_Casefold` in an `In` line. (I ended up not putting tests for the earliest applicable version, because we don’t have those for the other policies. We could add those later if we feel like it.)

macchiati: Makes sense

macchiati (approved):

PR #470, opened 2023-05-12
NamesList.txt 15.1 may12

markusicu:
Ken via file header:

Tweak use of notices to suppress year expansions.
Update annotations for 06F4..06F7.

Ken via email:

This delta fixes some annotations for Eastern Arabic digits, and also
removes some unnecessary notice prefixes on annotation lines that
contain numbers for standards (like "ISO 2047", etc.)


Ken-Whistler (approved):
Changes as expected.

PR #472, opened 2023-05-12
add Idna2008-15.1.0.txt

markusicu:
From Ken: derivation of IDNA2008_Category for 15.1.0


Ken-Whistler (approved):
That's the file and that's the right place to put it. ;-)

PR #473, opened 2023-05-15
IDNA 15.1: new CJK; no NFD validity check

markusicu:
For
[[175-A86](https://www.unicode.org/cgi-bin/GetL2Ref.pl?175-A86)] Action Item for Mark Davis, Markus Scherer, PAG: In IdnaMappingTable.txt, change U+2260 (≠), U+226E (≮), and U+226F (≯) from disallowed_STD3_valid to valid, for Unicode 15.1.

In the code for [Mapping Table Derivation, Step 7](https://unicode.org/reports/tr46/#TableDerivationStep7), we no longer check whether the NFD versions of characters and mappings are valid. We had agreed with feedback that that is not necessary because domain names are in composed form, not decomposed.

----
Also adds the other characters that are new in Unicode 15.1:
- 5 new IDS characters
- 603 new characters in CJK Extension I

### DIFF
--- a/UnicodeJsps/src/test/java/org/unicode/jsptest/TestJsp.java
+++ b/UnicodeJsps/src/test/java/org/unicode/jsptest/TestJsp.java
@@ -913,7 +913,7 @@ public class TestJsp extends TestFmwkMinusMinus {
         logln(uts46unic + ", " + error[0]);
         checkValues(error, Uts46.SINGLETON);
         checkValidIdna(Uts46.SINGLETON, "À｡÷");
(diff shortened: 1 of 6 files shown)


Ken-Whistler (dismissed):
Updates to the data file look fine. There are no implications for the derivation of Idna2008-15.1.0.txt. I'm not reviewing the Java code for testing.

macchiati (dismissed):

macchiati (approved):

echeran (approved):

PR #475, opened 2023-05-16
Text from Ken for 174-A5

eggrobin:
[[174-A5](https://www.unicode.org/cgi-bin/GetL2Ref.pl?174-A5)] Action Item for Ken Whistler, PAG: In DerivedCoreProperties.txt, document the implications of changing Default_Ignorable_Code_Point, for Unicode Version 15.1. See document [L2/23-008](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/23-008) item 1.1.


Ken-Whistler (approved):
Looks good to me.

PR #477, opened 2023-05-16
Update the rule files for 175-C26

eggrobin:
[[175-C26](https://www.unicode.org/cgi-bin/GetL2Ref.pl?175-C26)] Consensus: Modify Indic grapheme clusters in UAX29 as described in [L2/23-079](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/23-079) item 5.5, for Unicode Version 15.1.

[[175-A76](https://www.unicode.org/cgi-bin/GetL2Ref.pl?175-A76)] Action Item for Mark Davis, PAG: In UAX29, add the three macros Virama, LinkingConsonant, and ExtCccZwj, as well as the new rule 9.c, as described in [L2/23-079](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/23-079) item 5.5, for Unicode 15.1. Break out the list of scripts into a separate script set. Also fix the lack of sc=.


markusicu (approved):
lgtm tnx!

PR #478, opened 2023-05-16
UCA & security 15.1 generated files

markusicu:


Ken-Whistler (approved):
The files for which a visual diff results in something reasonable looked o.k. to me. Collation test changes looked like they were turning up the right things. I have no idea what some of the stuff in dev/data or dev/data/source is.

PR #480, opened 2023-05-16
script for beta, idna2008 readme

markusicu:


Ken-Whistler (approved):
The ReadMe.txt for the idna2008derived directory looks right. No comment on the data-workflow notes. The deployment script looks like it should do the expected deployment, but without more visibility on unicodetools I can't evaluate it in detail. I'll see what results from running it and see if that will work for beta.

josh-hadley (approved):
lgtm

markusicu:
> Just fwiw, line 108 still says "alpha" in the beta script.

Yes, that's in the “Publish an alpha snapshot” section. It's workflow documentation, not only for the beta.

PR #485, opened 2023-05-24
Delete all but one *Invariant*.txt

### LINKED ISSUE #380: delete all but one *Invariant*.txt
[In the repo, use "Go to file"](https://github.com/unicode-org/unicodetools/find/main) and look for files named like "invariant.txt". There is a confusing set of similarly-named files. Some of them are old according to their names or in-file comments.

The "real" invariants test file is https://github.com/unicode-org/unicodetools/blob/main/unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt

Can we delete each of the others? Or is any one of them still used by some code?
- (?) https://github.com/unicode-org/unicodetools/blob/main/unicodetools/src/main/resources/org/unicode/text/UCD/InvariantTest.txt
- (says "OLD") https://github.com/unicode-org/unicodetools/blob/main/unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariants.txt
- (named "attic...") https://github.com/unicode-org/unicodetools/blob/main/unicodetools/src/main/resources/org/unicode/text/UCD/atticUnicodeInvariants.txt
- (empty, named "scratch...") https://github.com/unicode-org/unicodetools/blob/main/unicodetools/src/main/resources/org/unicode/text/UCD/ScratchInvariantTest.txt
- (named "reallyold") https://github.com/unicode-org/unicodetools/blob/main/unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariants-reallyold.txt

eggrobin:
Scripsit @macchiati:
> You can remove all but the "real" one[.]

Fix #380.


macchiati (approved):

macchiati:
I'm assuming you looked these over and the remaining one is a superset.

markusicu (approved):
lgtm tnx

> I'm assuming you looked these over and the remaining one is a superset.

Maybe not a bad idea to look, but I would hope/assume that files called "attic" or "scratch" etc. don't have stuff that we should really keep.

eggrobin:
As far as I can tell, everything interesting in these old files has equivalents in the modern file.

Further, they probably cannot be run since https://github.com/unicode-org/unicodetools/commit/cd4afe40c7bd0d0a0689517f10496aa3b9fefa9c (2009), as the syntax has changed.

It is worth noting that the three that look vaguely like a test suite (namely atticUnicodeInvariants.txt, UnicodeInvariants-reallyold.txt, and the one plainly called UnicodeInvariants.txt), as opposed to a playground like InvariantTest.txt, have at least one test that is broken since Unicode Version 5.1, namely
`$Default_Ignorable_Code_Point = [[$Other_Default_Ignorable_Code_Point $GC:Cf $GC:Cc $GC:Cs $Variation_Selector $Noncharacter_Code_Point] - [$White_Space\uFFF9-\uFFFB]]`
which still has U+0600 and friends as DI.


Even if we want to do archæology, having them around is not actually useful, because they have been deleted and re-created several times in strange history of this source tree.

### For unicodetools archæologists:

#### History of the InvariantTest.txt playground:
https://github.com/unicode-org/unicodetools/blame/2d1a65225b0b376ebeb4396c95fcfac192c25855/org/unicode/text/UCD/InvariantTest.txt

#### History of the actual invariant tests:
| Time span | Blame |
|---|---|
| 2013/present | [UnicodeInvariantTest.txt](https://github.com/unicode-org/unicodetools/blame/main/unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt) |
| 2008/2013 | [UnicodeInvariantTest.txt](https://github.com/unicode-org/unicodetools/blame/2d1a65225b0b376ebeb4396c95fcfac192c25855/org/unicode/text/UCD/UnicodeInvariantTest.txt) |
| 2006 | No version controlled changes; -old and -reallyold created by https://github.com/unicode-org/unicodetools/commit/9e23f4af69b511db2f93c45d098bf8e011fce3d4 (old later renamed to attic by https://github.com/unicode-org/unicodetools/commit/6daa2d7d068df5ef40c60da72c2fd9a548dcd1ce). |
| 2004/2005 | [UnicodeInvariants.txt](https://github.com/unicode-org/unicodetools/blame/2d1a65225b0b376ebeb4396c95fcfac192c25855/org/unicode/text/UCD/UnicodeInvariants.txt) |
| 2004* | [UnicodeInvariants.txt](https://github.com/unicode-org/unicodetools/blame/5c930e8c1b8ed10f91529c82fa5dcf74d19f51b1/org/unicode/text/UCD/UnicodeInvariants.txt) |

\* UnicodeInvariants is spuriously deleted on 2004-10-14 (https://github.com/unicode-org/unicodetools/commit/64540d64203d9ee217570dbf84f5c38b02089c22, recreated in https://github.com/unicode-org/unicodetools/commit/ec384129cfb2377ea7adc43c0f3a521034ff583e).

Issue #484, opened 2023-05-24
The versioning of ToolUnicodePropertySource is neither tested nor correct


eggrobin:
Found while digging into the invariants tests, looking at a `U5.0.0:` property:
```
Attempt to reset value for 070F when that is disallowed. Old: Format; New: ALetter


java.lang.UnsupportedOperationException: Attempt to reset value for 070F when that is disallowed. Old: Format; New: ALetter
	at com.ibm.icu.dev.util.UnicodeMap._put(UnicodeMap.java:297)
	at com.ibm.icu.dev.util.UnicodeMap._putAll(UnicodeMap.java:377)
	at com.ibm.icu.dev.util.UnicodeMap.putAll(UnicodeMap.java:436)
	at org.unicode.text.UCD.ToolUnicodePropertySource.(ToolUnicodePropertySource.java:1212)
	at org.unicode.text.UCD.ToolUnicodePropertySource.make(ToolUnicodePropertySource.java:78)
```

The changes made to the derivation of the properties should be gated behind `version`, otherwise we are seeing incorrect old versions in the invariants tests, so we could silently break an invariant by changing the derivation.

We should test that for old versions we generate something consistent with what we published.

eggrobin:
Discussed in the 2023-05-30 meeting: we should test this by comparing IndexUnicodePropertySource against ToolUnicodePropertySource.

markusicu:
Long-term, we should probably drop the older ToolUnicodePropertySource.

eggrobin:
Well, something has to do the derivation if we want derived properties; by design IndexUnicodePropertySource has none of that logic.

Issue #432, opened 2023-04-06
VersionedProperty support in the JSPs


eggrobin:
The invariant test have a versioned property syntax, _e.g._, `\p{U4.1:XID_Continue}`, implemented in `VersionedProperty`.
https://github.com/unicode-org/unicodetools/blob/70dce2c89f185c65b436c28404ae5b7bdb32c2d1/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java#L1033-L1072

Something like that would be useful for list-unicodeset.jsp.
It sounds like there is limited support for multiple versions with the β syntax, but I cannot figure out how that one is implemented.

eggrobin:
> For the beta, the idea is that you have multiple release versions plus 1 dev version.

So, matching the subfolders of `ucd`? That was what I had in mind.

> Do you think we need multiple?

Multiple different βs? That sounds like it would confuse things rather than help.

macchiati:
When I read "limited support for multiple versions with the β syntax", I
thought you were asking for multiple betas. Glad you're not!

eggrobin:
> Should the invariants use it instead of ToolUnicodePropertySource?

Wait, they currently use _both_, with a preference for ToolUnicodePropertySource. So I guess the question is whether this should be inverted.

```java
         propSource = getProperties(version);  // ToolUnicodePropertySource.
         property = propSource.getProperty(xPropertyName); 
         if (property == null) { 
             propSource = getIndexedProperties(version);   // IndexUnicodeProperties.
             property = propSource.getProperty(xPropertyName); 
```

macchiati:
I agree, it should be inverted. There are a few constructed properties
(non-UCD) that are only available via Tool... That is, the extra
normalization properties. So those should fallback to the Tool

eggrobin:
> it should be inverted
Ah, that doesn’t work, because IndexUnicodeProperties, while closer to the truth for non-default values, it has issues with default values (I think it relies on @&zwnj;missing lines that are not always there).

At a glance, it is missing defaults at least for:
1. CCC, 
2. the non-special values of the properties in SpecialCasing (those used to have @&zwnj;missing lines, which IndexUnicodeProperties.txt knows how to read, but they got removed in 7.0), 
3.  the case foldings.

(The failings of IndexUnicodeProperties should probably turn into their own issue, but I am on a train with spotty internet, so I will be happy if this comment goes through.)

macchiati:
[...]

PR #487, opened 2023-05-26
Fix defaults for IndexUnicodeProperties.

### LINKED ISSUE #486: The properties provided by IndexUnicodeProperties have incorrect defaults
See https://github.com/unicode-org/unicodetools/issues/432#issuecomment-1563223360:

the following incorrectly have null defaults:
1. CCC,
2. the non-special values of the properties in SpecialCasing (those used to have @‌missing lines, which IndexUnicodeProperties.txt knows how to read, but they got removed in 7.0),
3. the case foldings.

eggrobin:
Fix #486, paving the way for the inversion discussed in https://github.com/unicode-org/unicodetools/issues/432#issuecomment-1562938248, and for consistency checks for #484.

The new spot checks all failed (with nulls) prior to this change, and so did some additional tests for U+FFFF in the loop (see the additions to ExtraPropertyValueAliases.txt).

Note that Bidi_Class is removed from ExtraPropertyValueAliases, as we have its @&zwnj;missing lines from the DerivedExtracted file used as the source in IndexUnicodeProperties.


macchiati:
This is a subset of the other one, eh?

eggrobin:
> This is a subset of the other one, eh?

Yes, that is why the other one is still a draft (contrary to Critique, GitHub doesn’t have diffbase, so I can’t give you a nice diff for a PR on top of another).

macchiati (approved):

PR #488, opened 2023-05-26
Prefer IndexUnicodeProperties over ToolUnicodePropertySource

eggrobin:
DRAFT: on top of #487.

See discussion in https://github.com/unicode-org/unicodetools/issues/432#issuecomment-1562938248:
@eggrobin:
> […] the question is whether this should be inverted.
> ```java
>         propSource = getProperties(version);  // ToolUnicodePropertySource.
>         property = propSource.getProperty(xPropertyName); 
>         if (property == null) { 
>             propSource = getIndexedProperties(version);   // IndexUnicodeProperties.
>             property = propSource.getProperty(xPropertyName); 
> ```

@macchiati:
> I agree, it should be inverted. There are a few constructed properties
(non-UCD) that are only available via Tool... That is, the extra
normalization properties. So those should fallback to the Tool

This uses IndexUnicodeProperties except for the following, where ToolUnicodePropertySource is used:
1. the toNF[&nbsp;] pseudoproperties;
2. properties that did not exist in that version of Unicode, _if_ the version is requested with an R (for _Retroactive_) rather than a U.

This is used for the invariant about NFKC_SCF, which does not exist until 15.1, but whose derivation would be stable very far back.

This also fixes defaults (incorrectly null instead of `<code point>`) in the NFKC_CF and NFKC_SCF returned by ToolUnicodePropertySource.


eggrobin:
> `TestTestUnicodeInvariants.testUnicodeInvariants:35 TestUnicodeInvariants.testInvariants() failed ==> expected: <0> but was: <2>`

That’s odd: on my machine it works.

macchiati (approved):
Nice work

Issue #325, opened 2022-08-30
Clean up CheckXmlProperties


macchiati:
See https://unicode-org.github.io/unicodetools/newunicodeproperties.html#checking-xml-properties

1.  Make into JUnit test
2. Fix the failures and warnings due to differences in interpretation (eg the UcdProperty values include defaults) and problems with the test (interpreting #, and \u00xxx.)

markusicu:
Mark, in email you wrote about the Meroitic fractions: “the UcdProperties have the reduced values”
Is that necessary? As long as we still store the fraction strings, it would be more illustrative to keep the fractions as-is, rather than simplifying them.

macchiati:
The software right now treats them as numerical values — which is the way
most software will. So I think it is a mistake to treat them just as the
surface form.

I can just smarten the test to compare the numeric value of both the xml
format and the UcdProperty format.

markusicu:
> The software right now treats them as numerical values — which is the way most software will. So I think it is a mistake to treat them just as the surface form.

Depends on what the goal is. I doubt that most software does anything with numeric values of characters other than gc=Nd.
For a precise, readable, understandable presentation of the value, these fractions are best written as in UnicodeData.txt.

> I can just smarten the test to compare the numeric value of both the xml format and the UcdProperty format.

ok -- presumably with an epsilon for dealing with rounding issues

macchiati:
Can do exact match: we have a Rational number class in CLDR.

Issue #125, opened 2021-09-11
Remove false positives from the ucdxml consistency tests


macchiati:
A lot of spurious errors/warnings occur; here is a listing:

I found instructions here:
https://github.com/unicode-org/unicodetools/blob/main/docs/newunicodeproperties.md#checking-xml-properties
including
NOTE: the following (until the test is fixed) are false differences

There are also comments with "TODO Known problems" at the top of the program source:
https://github.com/unicode-org/unicodetools/blob/main/unicodetools/org/unicode/propstest/CheckXmlProperties.java

Ran the test program -->
Failing tests: 9
Warning tests: 4

TESTING Numeric_Value

Property cp xml unicodetools
Numeric_Value 109F7 [2/12] [1/6]
Numeric_Value 109F8 [3/12] [1/4]
Numeric_Value 109F9 [4/12] [1/3]
Numeric_Value 109FB [6/12] [1/2]
Numeric_Value 109FD [8/12] [2/3]
Numeric_Value 109FE [9/12] [3/4]
Numeric_Value 109FF [10/12] [5/6]
*FAIL* Numeric_Value with 7 errors.
...
*FAIL* Bidi_Paired_Bracket with 1113984 errors.
...
TESTING Lowercase_Mapping

Property cp xml unicodetools
Lowercase_Mapping 0041 [a] [A]
Lowercase_Mapping 0042 [b] [B]
Lowercase_Mapping 0043 [c] [C]
Lowercase_Mapping 0044 [d] [D]
Lowercase_Mapping 0045 [e] [E]
Lowercase_Mapping 0046 [f] [F]
Lowercase_Mapping 0047 [g] [G]
Lowercase_Mapping 0048 [h] [H]
Lowercase_Mapping 0049 [i] [I]
Lowercase_Mapping 004A [j] [J]
*FAIL* Lowercase_Mapping with 1405 errors.
...
*FAIL* Titlecase_Mapping with 1377 errors.
*FAIL* Uppercase_Mapping with 1423 errors.
*FAIL* Name with 100878 errors.
*FAIL* Canonical_Combining_Class with 829834 errors.
*FAIL* Joining_Group with 1113286 errors.
*FAIL* Joining_Type with 1113284 errors.

*WARNING* kAccountingNumeric with 1114086 “empty” differences.
*WARNING* kOtherNumeric with 1114082 “empty” differences.
*WARNING* kPrimaryNumeric with 1114095 “empty” differences.
[...]

PR #489, opened 2023-05-27
UCDXML 15.1 beta from may22

markusicu:
Two commits:
- Make a copy of the 15.0 data. I forgot to do that last year.
- 15.1 data from @liancu-uni

I ran CheckXmlProperties and it has a lot of complaints, but at a glance they all look like the kinds of things noted in issues #125 & #325.

### DIFF
--- a/unicodetools/data/ucdxml/15.0.0/ucdxml.readme.txt
+++ b/unicodetools/data/ucdxml/15.0.0/ucdxml.readme.txt
@@ -1,0 +1,33 @@
+XML Representation of Unicode 15.0.0 UCD
+
+
+© 2022 Unicode®, Inc.
+For terms of use, see https://www.unicode.org/terms_of_use.html
+
+
+This directory contains the representation in XML of Version 15.0.0 of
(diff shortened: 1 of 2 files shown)


Ken-Whistler (commented):
Why a second copy of the ucdxml.readme.txt ? Shouldn't this go in one place?

markusicu:
> Why a second copy of the ucdxml.readme.txt ? Shouldn't this go in one place?

We have the same three files for each version. See
- https://github.com/unicode-org/unicodetools/tree/main/unicodetools/data/ucdxml/13.0.0
- https://github.com/unicode-org/unicodetools/tree/main/unicodetools/data/ucdxml/14.0.0
- https://github.com/unicode-org/unicodetools/tree/main/unicodetools/data/ucdxml/dev

If the Unicode Tools didn't try to make old versions available in parallel, then we wouldn't need folder names with version numbers, and different folders with different versions of the same files.

Ken-Whistler (approved):
Not attempting to manually compare the xml file diffs. Changes to the readme look o.k.

PR #490, opened 2023-05-30
Don’t duplicate spotless failures as build failures

eggrobin:
Right now if some code is improperly formatted, the `Formatter + Style checker` job fails (as expected), but the `Out-of-source Instructions` (build and test) job also fails, which has repeatedly confused me.

This is because the GenerateEnums test checks for differences after running GenerateEnums and reformatting the whole source tree.

This PR changes that to reformatting only the generated files, avoiding the spurious test failure:
- Workflow run with only formatting errors: https://github.com/eggrobin/unicodetools/actions/runs/5098935042;
- Workflow run with modified (but stylistically correct) UcdProperty: https://github.com/unicode-org/unicodetools/actions/runs/5121706678/jobs/9209817846.

The PR also removes the fancy #100-related diff filtering: while #100 is still around, it no longer induces diffs since #424.

### DIFF
--- a/.github/workflows/cli-build-instructions.yml
+++ b/.github/workflows/cli-build-instructions.yml
@@ -218,9 +218,6 @@ jobs:
             echo "Failure: GenerateEnums has changes that need to be committed."
             exit 1
           fi
-          # TODO(#100): Note: when running locally on a clean checkout, this command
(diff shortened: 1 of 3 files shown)


macchiati (dismissed):

markusicu (approved):
lgtm tnx -- please squash-and-merge

PR #511, opened 2023-07-28
UTS46 map capital to lowercase sharp s

markusicu:
Changes for new action items from UTC-176:
- In UTS46 Mapping Table Derivation step 1 (base mapping) substep 1 (exceptional characters), map U+1E9E capital sharp s to U+00DF small sharp s. Add a note that this changes the mapping of the capital sharp s. For Unicode 15.1.
- In IdnaMappingTable.txt map U+1E9E capital sharp s to U+00DF small sharp s instead of to "ss", for Unicode 15.1.

*Please read the following and advise!*

It turns out that the first action item (adding a base  mapping) is insufficient: Table derivation step 6 checks that the mapping contains only characters in the base valid set before using the mapping. Lowercase sharp s is not in the base valid set (it has an NFKC_Casefold mapping). Thus, the mapping table is generated with capital sharp s disallowed.

I chose to add lowercase sharp s to the base valid set. That seemed like the most straightforward and understandable way out.

Then it turns out that that is not sufficient either: Table derivation step 7 also checks that every character in a mapping is in the valid set. Lowercase sharp is in the deviation set, not in the valid set. Thus, capital sharp s is still disallowed.

Here I chose to also allow deviation characters in mappings. I was not sure whether that would inadvertently allow more mappings in, but it doesn't (at least not currently).

Please let me know if I should do something different. For example, in step 7 we could just add lowercase sharp s alone to the characters allowed in mappings. We could even just hack this one desired mapping into step 7 and not make any of the other changes.

Note that whatever derivation changes we agree on I will also have to make in the UTS46 spec text.

### DIFF
--- a/unicodetools/data/idna/dev/IdnaTestV2.txt
+++ b/unicodetools/data/idna/dev/IdnaTestV2.txt
@@ -1,5 +1,5 @@
 # IdnaTestV2.txt
-# Date: 2023-06-06, 23:32:28 GMT
+# Date: 2023-07-27, 23:16:06 GMT
 # © 2023 Unicode®, Inc.
 # Unicode and the Unicode Logo are registered trademarks of Unicode, Inc. in the U.S. and other countries.
 # For terms of use, see https://www.unicode.org/terms_of_use.html
@@ -217,6 +217,8 @@ o\u0308bb; öbb; ; xn--bb-eka; ; ;  # öbb
 Öbb; öbb; ; xn--bb-eka; ; ;  # öbb
 O\u0308bb; öbb; ; xn--bb-eka; ; ;  # öbb
 xn--bb-eka; öbb; ; xn--bb-eka; ; ;  # öbb
+FAẞ.de; faß.de; ; xn--fa-hia.de; ; ; [V6] # faß.de
+FAẞ.DE; faß.de; ; xn--fa-hia.de; ; ; [V6] # faß.de
 βόλος.com; ; ; xn--nxasmm1c.com; ; xn--nxasmq6b.com;  # βόλος.com
 βο\u0301λος.com; βόλος.com; ; xn--nxasmm1c.com; ; xn--nxasmq6b.com;  # βόλος.com
 ΒΟ\u0301ΛΟΣ.COM; βόλοσ.com; ; xn--nxasmq6b.com; ; ;  # βόλοσ.com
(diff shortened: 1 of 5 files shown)


markusicu:
@michelsu please advise on the questions in the description.

unicodetools/data/idna/dev/IdnaTestV2.txt
```diff
@@ -217,6 +217,8 @@ o\u0308bb; öbb; ; xn--bb-eka; ; ;  # öbb
 Öbb; öbb; ; xn--bb-eka; ; ;  # öbb
 O\u0308bb; öbb; ; xn--bb-eka; ; ;  # öbb
 xn--bb-eka; öbb; ; xn--bb-eka; ; ;  # öbb
+FAẞ.de; faß.de; ; xn--fa-hia.de; ; ; [V6] # faß.de
+FAẞ.DE; faß.de; ; xn--fa-hia.de; ; ; [V6] # faß.de
```
roozbehp, line 221: Why is the same line repeated?

asmusf:
The problem seems to be specific to how the tools are defined, which in turn is based on the assumptions that go into defining the transition between two incompatible standards.

Your description sounds reasonable, but I cannot validate it, not knowing the ins and outs of the existing tools.

macchiati (dismissed):

markusicu:
> The problem seems to be specific to how the tools are defined

Not in this case. The code follows the spec very closely.
https://unicode.org/reports/tr46/proposed.html#TableDerivationStep6

markusicu:
Well, I screwed up. I pushed a UTS46 spec change, but had neglected to create a branch, so I ended up directly pushing to unicode-org main :-(

I had meant https://github.com/unicode-org/unicode-reports/commit/1d665bdfe43b8210a66a01dac438dfae3215366f to be in a pull request so that y'all could take another look at what the spec changes really look like that correspond to the code changes. Instead, it's already up at https://unicode-org.github.io/unicode-reports/tr46/tr46.html

So @macchiati @Ken-Whistler @michelsu @asmusf @PeterConstable let me know if you disagree with my choices. I will wait for a couple of hours before I merge this unicodetools PR, and we can always adjust things later anyway.

markusicu:
Force-pushed with a change in a backwards-compatibility test that was complaining that the mapping of capital sharp s differed from before.

echeran (approved):
RSLGTM

PR #515, opened 2023-08-08
176-C27 Change rules LB15a and LB15b in UAX14 so that they treat class ZW like BK

eggrobin:
[[176-A93](https://www.unicode.org/cgi-bin/GetL2Ref.pl/?176-A93)] Action Item for Robin Leroy, PAG: Change the tooling that generates LineBreakTest.txt and LineBreakTest.html to adjust for the changes to the Proposed Update for UAX #&zwnj;14 described in L2/23-160 item 4.4. For Unicode Version 15.1.

Added a new test since the auto-generated ones do not see the difference.


markusicu (approved):
lgtm

macchiati (approved):

PR #520, opened 2023-08-15
UTS46 transitional: map capital sharp s to ss

markusicu:
... to keep the mapping idempotent, and to avoid the small sharp s failing the mode=transitional validity check

FYI @michelsu


markusicu:
ICU's UTS46Test passes with this updated test file!

Manishearth (approved):

PR #524, opened 2023-08-18
Tulu-Tigalari

eggrobin:
UTC-170-C9 Consensus: UTC accepts 78 Tulu-Tigalari characters in a new Tulu-Tigalari block U+11380..U+113FF as documented in [L2/22-031](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/22-031), for encoding in a future version of the standard.

UTC-176-C16 Consensus: The UTC approves the code point change for TULU-TIGALARI LETTER DHA to U+113A4, from the conflicting U+113A5 approved by UTC-170-C9. TULU-TIGALARI LETTER NA remains at U+113A5.

Plus correcting a typo in the InSC data from L2/22-031.

LineBreak according to [L2/22-080](https://www.unicode.org/cgi-bin/GetDocumentLink?L2/22-080).


eggrobin:
CI does not pass because of collation.

eggrobin:
Ken noted that the avagraha should probably be LB=BA like Grantha, but L2/22-080R2 recommends LB=ID.
Sent an email to Norbert to see if the discrepancy was intentional.

eggrobin:
The discrepancy was intentional.

Scripsit Norbert:

> For Tulu-Tigalari, one of the experts told us that line breaks can occur pretty much at any spacing character. OpenType doesn’t let you reorder characters across line breaks, and doesn’t deal well with line breaks before combining marks, so we can’t quite replicate that. But avagrahas with gc=Lo are treated as a base, so there’s no problem with breaking before them.

markusicu:
Several non-UCA CI failures...

unicodetools/data/ucd/dev/DerivedCoreProperties.txt
```diff
@@ -12738,6 +12816,9 @@ ABED          ; Grapheme_Link # Mn       MEETEI MAYEK APUN IYEK
 11235         ; Grapheme_Link # Mc       KHOJKI SIGN VIRAMA
 112EA         ; Grapheme_Link # Mn       KHUDAWADI SIGN VIRAMA
 1134D         ; Grapheme_Link # Mc       GRANTHA SIGN VIRAMA
+113CE         ; Grapheme_Link # Mn       TULU-TIGALARI SIGN VIRAMA
```
roozbehp, line 12819: 113CE and 113CF are pure killers. Shouldn't they be Grampheme_Extend instead?

unicodetools/data/ucd/dev/DerivedCoreProperties.txt
```diff
@@ -12954,6 +13035,8 @@ FE20..FE2F    ; InCB; Extend # Mn  [16] COMBINING LIGATURE LEFT HALF..COMBINING
 1133B..1133C  ; InCB; Extend # Mn   [2] COMBINING BINDU BELOW..GRANTHA SIGN NUKTA
 11366..1136C  ; InCB; Extend # Mn   [7] COMBINING GRANTHA DIGIT ZERO..COMBINING GRANTHA DIGIT SIX
 11370..11374  ; InCB; Extend # Mn   [5] COMBINING GRANTHA LETTER A..COMBINING GRANTHA LETTER PA
+113CE         ; InCB; Extend # Mn       TULU-TIGALARI SIGN VIRAMA
```
roozbehp, line 13038: Why is 113CF not here?

unicodetools/data/ucd/dev/auxiliary/GraphemeBreakProperty.txt
```diff
@@ -27,14 +27,15 @@
 110BD         ; Prepend # Cf       KAITHI NUMBER SIGN
 110CD         ; Prepend # Cf       KAITHI NUMBER SIGN ABOVE
 111C2..111C3  ; Prepend # Lo   [2] SHARADA SIGN JIHVAMULIYA..SHARADA SIGN UPADHMANIYA
+113D1         ; Prepend # Lo       TULU-TIGALARI REPHA
```
roozbehp, line 30: This is yet another invariant from InSC if you don't already have it: Consonant_Preceding_Repha is Prepend.

roozbehp (changes requested):

unicodetools/data/ucd/dev/DerivedCoreProperties.txt
markusicu: Looks like those properties are not mutually exclusive. Grapheme_Link is deprecated, and derived from ccc=9 (Virama).

The non-looped virama *is* Grapheme_Extend = Me + Mn + Other_Grapheme_Extend

The looped virama is gc=Mc not Mn, so if we want that to be Grapheme_Extend then it needs to get Other_Grapheme_Extend. UAX44 points to core spec chapter 3 D59 for that; I haven't looked at that yet.

And because the looped virama is not Grapheme_Extend, it is currently a Grapheme_Base.

unicodetools/data/ucd/dev/DerivedCoreProperties.txt
markusicu: Because the looped virama is not gcb=Extend, because it's not Grapheme_Extend, because it's gc=Mc and not Other_Grapheme_Extend.

unicodetools/data/ucd/dev/auxiliary/GraphemeBreakProperty.txt
eggrobin: We have it: https://github.com/unicode-org/unicodetools/blob/1f791fe19c7b88e190c03aee67de3f20e6a9564b/unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt#L725 (and that is the derivation anyway: https://www.unicode.org/reports/tr29/#Prepend).

unicodetools/data/ucd/dev/DerivedCoreProperties.txt
markusicu: Comparison shopping: None of the existing gc=Mc, ccc=9 characters are Grapheme_Extend. [`[:^ccc=0:]-[:Mn:]`](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3A%5Eccc%3D0%3A%5D-%5B%3AMn%3A%5D&g=gc+Grapheme_Extend&i=ccc)

eggrobin: > Shouldn't they be Grampheme_Extend instead?

The implication here is that they extend Extended Grapheme Clusters (because they are GCB=SpacingMark), but not Legacy Grapheme Clusters (which is what GCB=Grapheme_Extend does).
In other words, they behave like Mc dependent vowels for grapheme clusters (both legacy and extended).
That doesn’t seem unreasonable to me.

markusicu: I looked over all of the changes. They look plausible to me.

It seems like the major question from @roozbehp is whether the looped virama should be Other_Grapheme_Extend or even gc=Mn (instead of Mc). @Ken-Whistler ?

roozbehp: > It seems like the major question from @roozbehp is whether the looped virama should be Other_Grapheme_Extend or even gc=Mn (instead of Mc).

I think it should still be Mc. If that affects Other_Grapheme_Extend, fine, I have no objections. If Other_Grapheme_Extend is independent, we should consider it (I don't understand all the implications).

markusicu: I am looking at the core spec now.

D53 Nonspacing mark: A combining character with the General Category of Nonspacing
Mark (Mn) or Enclosing Mark (Me). The position of a nonspacing mark in presentation depends on its base character. *It generally does not consume space along the visual baseline in and of itself*.

D55 Spacing mark: A combining character that is not a nonspacing mark.

--> It makes sense for the Tulu-Tigalari virama to be Mn and the looped virama to be Mc.

D59 Grapheme extender: A character with the property Grapheme_Extend.
- Grapheme extender characters consist of all nonspacing marks, zero width joiner, zero width non-joiner, U+FF9E halfwidth katakana voiced sound mark, U+FF9F halfwidth katakana semi-voiced sound mark, and
a small number of spacing marks.
- ...
- The small number of spacing marks that have the property Grapheme_Extend are all the second parts of a two-part combining mark.

--> So, by definition, the looped virama does not qualify for Grapheme_Extend.

It does attach to the preceding character, and even graphically join (ligate) with it, so it seems weird that it's a Grapheme_Base and starts its own cluster.

In other words, if we want the looped virama to be Grapheme_Extend, we need to give it Other_Grapheme_Extend, but *we also have to add this case to D59*.

[Here is the list of spacing marks with Grapheme_Extend](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3Agc%3DMc%3A%5D%26%5B%3AGrapheme_Extend%3A%5D&g=&i=). Are all of these “the second parts of a two-part combining mark”?

eggrobin: > It does attach to the preceding character, and even graphically join (ligate) with it, so it seems weird that it's a Grapheme_Base and starts its own cluster.

It starts its own _legacy_ grapheme cluster, and that is exactly the kind of weird thing that happens with legacy grapheme clusters.
Being a spacing mark, it extends _extended_ grapheme clusters by [GB9a](https://www.unicode.org/reports/tr29/#GB9a).

markusicu: Ok, thanks. I see that you basically said that already last night. It makes sense to me now.

For my own reference: UAX29 Table 1a. [Sample Grapheme Clusters](https://www.unicode.org/reports/tr29/#Table_Sample_Grapheme_Clusters)

@roozbehp as Robin said, this then behaves similarly to various gc=Mc vowels. Ok?

One more query: [gc=Mc by Grapheme_Extend=No vs. Yes](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3Agc%3DMc%3A%5D&g=Grapheme_Extend&i=)

roozbehp: > @roozbehp as Robin said, this then behaves similarly to various gc=Mc vowels. Ok?

Ok. My only remaining comment is the comment to be added to the InPC.txt header then.

roozbehp (approved):
LGTM

PR #527, opened 2023-08-18
UCDXML 15.1 2023aug16

markusicu:
@liancu-uni provided the UCDXML versions of what should be the final Unicode 15.1 data.

I ran the CheckXmlProperties program. The "failures" that it prints all look like the known value representation issues as before (see issue #325).


Ken-Whistler (approved):
Diffs are too big to look at. Assume the updated data is o.k.

PR #530, opened 2023-08-22
templatize CollationTest.html

markusicu:
like the ReadMe.txt files, replacing routine manual updates


eggrobin (approved):

Ken-Whistler (approved):
Seems fine to me to template the file this way, on the assumption that it will be clearly documented that all these templated files need to be systematically generated for deployment.

PR #339, opened 2022-09-30
copy data/*/dev folders to 15.0(.0)

markusicu:
Post Unicode 15 release task:
- copy `data/*/dev` folders to `15.0(.0)`
- fix issue #157

After this step, we can continue to develop the data files for the next version in the `dev` folders.

PR #539, opened 2023-09-19
Unicode 15.1→16.0

markusicu:
Best to review one commit at a time:
- copy the data "dev" folders to "15.1(.0)"
- add version 16.0, adjust latest & last version numbers
- generated files
- fix tests

Compare with moving to 15.1 last year:
- #339
- #340


unicodetools/src/main/resources/org/unicode/text/UCD/MakeUnicodeFiles.txt
```diff
@@ -233,6 +233,9 @@ Value:  V15_0
 Value:  V15_1
 # Newly assigned in Unicode 15.1.0 (September, 2023)
 
+Value:  V16_0
+# Newly assigned in Unicode 16.0.0 (September, 2023)
```
eggrobin, line 237:
```suggestion
# Newly assigned in Unicode 16.0.0 (September, 2024)
```

srl295:
Reviewed all but the "copy" commit, looks good so far

srl295 (approved):

Ken-Whistler (approved):
Presuming this is just the rollover for the 16.0 base, using the 15.1 data. No comments on the tooling changes per se.

unicodetools/src/main/resources/org/unicode/text/UCD/MakeUnicodeFiles.txt
markusicu: done tnx

markusicu:
@nedley: I amended the second commit to include tools/emoji/Emoji.java, and I made that a little less manual.

unicodetools/src/main/java/org/unicode/tools/emoji/Emoji.java
```diff
@@ -103,62 +102,49 @@ public class Emoji {
     public static final VersionInfo VERSION12_1 = VersionInfo.getInstance(12, 1);
     public static final VersionInfo VERSION12 = VersionInfo.getInstance(12);
     public static final VersionInfo VERSION11 = VersionInfo.getInstance(11);
+    public static final VersionInfo VERSION10 = VersionInfo.getInstance(10);
```
nedley, line 105: What are these for? They don’t correspond to any emoji versions I’m aware of.

markusicu: Rather than duplicating some VersionInfo instances in both VERSIONxx and UCDxx I made everything just VERSIONxx. There were some UCDxx that didn't already have VERSIONxx. Having just one kind of constant for a VersionInfo constant seems much simpler. The VersionInfo itself doesn't know whether it's for UCD or emoji.

These are used in EMOJI_TO_UNICODE_VERSION, on the right side.

markusicu:
I fixed the test failure. CheckProperties has another parser of numeric values. I don't know why it fails only with the previous version having multiple kPrimaryNumericValues but works if the current version has them. Now it only looks at the first value.

nedley (approved):

josh-hadley (approved):

PR #546, opened 2023-10-02
Script codes from the future

eggrobin:
Following up on #443; these codes were added on 2023-09-12: https://unicode.org/iso15924/codechanges.html.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/text/UCD/UCD_Names.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/UCD_Names.java
@@ -415,18 +415,18 @@ public final class UCD_Names implements UCD_Types {
         "Kawi",
         "Nag_Mundari",
         // Unicode 16
-        "Sunuwar",
-        "Tulu_Tigalari",
-        "Kirat_Rai",
-        "Todhri",
         "Garay",
         "Gurung_Khema",
+        "Kirat_Rai",
         "Ol_Onal",
+        "Sunuwar",
+        "Todhri",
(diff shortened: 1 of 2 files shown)


unicodetools/src/main/java/org/unicode/text/UCD/UCD_Names.java
```diff
@@ -414,7 +414,7 @@ public final class UCD_Names implements UCD_Types {
         // Unicode 15
         "Kawi",
         "Nag_Mundari",
-        // A future version of Unicode
+        // Unicode 16
         "Sunuwar",
         "Tulu_Tigalari",
         "Kirat_Rai",
```
markusicu, line 420: FYI -- not really important / nice-to-have

For the last several versions, we have sorted scripts alphabetically within each version.

markusicu (dismissed):
lgtm

unicodetools/src/main/java/org/unicode/text/UCD/UCD_Names.java
eggrobin: Alphabetically by script name or by script code?

markusicu: Good question. Judging from the Unicode 11 list, it's by script _code_, at least at that time.
```
        "Sogdian",
        "Old_Sogdian",
    ...
        "Sogd",
        "Sogo",
```

markusicu (approved):
lgtm tnx

PR #448, opened 2023-04-19
Regenerate EAW and VO, fix a UnicodeSet bug

eggrobin:
Follow-up on #438: MakeUnicodeFiles now does the three Ken files.

Correct an obsolete comment in LB and EAW.

Also fix the VerticalOrientation property created by ToolUnicodePropertySource, which caused UnicodeSet to be wrong about sets based on VO: compare

https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3ACn%3A%5D%26%5B%3Ablk%3DTangut%3A%5D%26%5B%3AVO%3DU%3A%5D&g=&i=

and 

https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3ACn%3A%5D%26%5B%3Ablk%3DTangut%3A%5D&g=VO

PR #547, opened 2023-10-02
Make MakeUnicodeFiles regenerate IndicMeowCategory.txt

eggrobin:
Follow-up on #438, #448, and #449; see rationale in #438.

### DIFF
--- a/unicodetools/data/ucd/dev/IndicPositionalCategory.txt
+++ b/unicodetools/data/ucd/dev/IndicPositionalCategory.txt
@@ -1,5 +1,5 @@
 # IndicPositionalCategory-16.0.0.txt
-# Date: 2023-10-02, 18:04:25 GMT
+# Date: 2023-10-02, 22:58:33 GMT
 # © 2023 Unicode®, Inc.
 # Unicode and the Unicode Logo are registered trademarks of Unicode, Inc. in the U.S. and other countries.
 # For terms of use, see https://www.unicode.org/terms_of_use.html
--- a/unicodetools/data/ucd/dev/IndicPositionalCategory.txt
+++ b/unicodetools/data/ucd/dev/IndicPositionalCategory.txt
@@ -1,11 +1,11 @@
-# IndicPositionalCategory-15.1.0.txt
-# Date: 2023-01-05
+# IndicPositionalCategory-16.0.0.txt
+# Date: 2023-10-02, 18:04:25 GMT
 # © 2023 Unicode®, Inc.
 # Unicode and the Unicode Logo are registered trademarks of Unicode, Inc. in the U.S. and other countries.
 # For terms of use, see https://www.unicode.org/terms_of_use.html
 #
-# For documentation, see UAX #44: Unicode Character Database,
-# at https://www.unicode.org/reports/tr44/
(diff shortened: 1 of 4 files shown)


unicodetools/data/ucd/dev/IndicPositionalCategory.txt
```diff
@@ -1,11 +1,11 @@
...
+# IndicPositionalCategory-16.0.0.txt
+# Date: 2023-10-02, 22:58:33 GMT
 # © 2023 Unicode®, Inc.
 # Unicode and the Unicode Logo are registered trademarks of Unicode, Inc. in the U.S. and other countries.
 # For terms of use, see https://www.unicode.org/terms_of_use.html
 #
-# For documentation, see UAX #44: Unicode Character Database,
-# at https://www.unicode.org/reports/tr44/
+# Unicode Character Database
+#   For documentation, see https://www.unicode.org/reports/tr44/
```
markusicu, line 8: question only: Better consistent with other files but not with Roozbeh's files here? Will we see flip-flopping between Roozbeh sending updates and us regenerating these files?

unicodetools/src/main/resources/org/unicode/text/UCD/MakeUnicodeFiles.txt
```diff
@@ -908,6 +908,376 @@ Format: kenFile skipValue=Rotated
...
+#   * U+1A69 TAI THAM VOWEL SIGN U and U+1A6A TAI THAM VOWEL SIGN UU have
+#     contextually variable placement in Tai Tham.
+#
+# 4. The following character is assigned the positional category Left, but
+# may have different positions in different styles:
+#   * U+119D2 NANDINAGARI VOWEL SIGN I has stylistically variable placement
+#     in Nandinagari.
+Property: Indic_Positional_Category
+Format:	roozbehFile separateValues=false valueStyle=short skipValue=NA
+Value: Right
```
markusicu, line 1022: question only: Is there any check for this list of values matching the one in PropertyValueAliases.txt = in our generated enum?

markusicu (approved):
yeah, fancy Java! ;-)

unicodetools/data/ucd/dev/IndicPositionalCategory.txt
eggrobin: We will see flip-flopping unless or until Roozbeh updates this part of his template, something similar (though it was just the indentation) happened with Ken’s LineBreak.txt.

This header is generated in a different place, so it would be mildly annoying to pass a Roozbeh flag around to customize it.

unicodetools/src/main/resources/org/unicode/text/UCD/MakeUnicodeFiles.txt
eggrobin: This will not work if the values are not value aliases of the property, I wanted to add a check that the list was exhaustive, I forgot.
Added one and checked that when commenting out a couple of lines I got `java.lang.IllegalArgumentException: All values must be listed when using roozbehFile; missing [Left, Top_And_Left]`.

markusicu (approved):

PR #548, opened 2023-10-03
Make the canonical decomposition invariant test easier to maintain

eggrobin:
Test what we actually want to test, instead of having to pile on more inscrutable exceptions whenever we add decompositions.


macchiati (approved):

macchiati (approved):
I didn't catch that either

PR #552, opened 2023-10-03
Garay

eggrobin:
[[171-C18](https://www.unicode.org/cgi-bin/GetL2Ref.pl?171-C18)] Consensus: Accept 69 Garay characters in a new Garay block (U+10D40..U+10D8F) for encoding in a future version of the standard. (Reference [L2/22-048](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/22-048))

Note that the proposed UnicodeData.txt lines have a duplicate code point, see bbfffdc499fade67283dc8fe38c2d9474678ffa4. Added to unicode-org/properties#169.


eggrobin:
CI does not pass because of collation.

roozbehp:
This also needs Script Extensions properties added for Arabic comma, question mark, and semicolon (see the proposal, Section 4).

markusicu (commented):
- so far so good!
- please rebase on top of merged PR #546
- see Roozbeh’s request for scx additions

unicodetools/data/ucd/dev/PropList.txt
markusicu, line 1: no Diacritic or Extender for any Garay characters?

eggrobin: I would also tend to think that GARAY VOWEL SIGN E should be at least one of Alphabetic and Diacritic.

unicodetools/data/ucd/dev/PropList.txt
```diff
@@ -1044,6 +1045,8 @@ FFE3          ; Diacritic # Sk       FULLWIDTH MACRON
 10AE5..10AE6  ; Diacritic # Mn   [2] MANICHAEAN ABBREVIATION MARK ABOVE..MANICHAEAN ABBREVIATION MARK BELOW
 10D22..10D23  ; Diacritic # Lo   [2] HANIFI ROHINGYA MARK SAKIN..HANIFI ROHINGYA MARK NA KHONNA
 10D24..10D27  ; Diacritic # Mn   [4] HANIFI ROHINGYA SIGN HARBAHAY..HANIFI ROHINGYA SIGN TASSI
+10D4E         ; Diacritic # Lo       GARAY VOWEL LENGTH MARK
```
markusicu, line 1048: The Garay length mark does not look like a diacritic to me.
It looks more like a U+30FC KATAKANA-HIRAGANA PROLONGED SOUND MARK ... which is both Extender and, to my surprise, also Diacritic.
And, that one is gc=Lm while the Garay length mark is gc=Lo. There are many Lm Diacritics but few Lo ones.

So:
- Should the Garay length mark be gc=Lm?
- Why is U+30FC a Diacritic?

@Ken-Whistler advice?

unicodetools/data/ucd/dev/PropList.txt
```diff
@@ -1134,6 +1137,8 @@ AADD          ; Extender # Lm       TAI VIET SYMBOL SAM
 AAF3..AAF4    ; Extender # Lm   [2] MEETEI MAYEK SYLLABLE REPETITION MARK..MEETEI MAYEK WORD REPETITION MARK
 FF70          ; Extender # Lm       HALFWIDTH KATAKANA-HIRAGANA PROLONGED SOUND MARK
 10781..10782  ; Extender # Lm   [2] MODIFIER LETTER SUPERSCRIPT TRIANGULAR COLON..MODIFIER LETTER SUPERSCRIPT HALF TRIANGULAR COLON
+10D4E         ; Extender # Lo       GARAY VOWEL LENGTH MARK
+10D6A         ; Extender # Mn       GARAY CONSONANT GEMINATION MARK
```
markusicu, line 1141: I would expect the U+10D6F GARAY REDUPLICATION MARK to also be an Extender, like U+3005 IDEOGRAPHIC ITERATION MARK.

unicodetools/data/ucd/dev/PropList.txt
markusicu: I agree that vowel sign E should be Alphabetic, like the Hebrew points and many Indic vowel signs.

unicodetools/data/ucd/dev/PropList.txt
eggrobin: For context:

[Diacritic](https://www.unicode.org/reports/tr44/#Diacritic) |	B	| I	| Characters that linguistically modify the meaning of another character to which they apply. Some diacritics are not combining characters, and some combining characters are not diacritics.
-- | -- | -- | --

unicodetools/data/ucd/dev/UnicodeData.txt
```diff
@@ -19327,6 +19327,75 @@ FFFD;REPLACEMENT CHARACTER;So;0;ON;;;;;N;;;;;
...
+10D44;GARAY DIGIT FOUR;Nd;0;AN;;4;4;4;N;;;;;
+10D45;GARAY DIGIT FIVE;Nd;0;AN;;5;5;5;N;;;;;
+10D46;GARAY DIGIT SIX;Nd;0;AN;;6;6;6;N;;;;;
+10D47;GARAY DIGIT SEVEN;Nd;0;AN;;7;7;7;N;;;;;
+10D48;GARAY DIGIT EIGHT;Nd;0;AN;;8;8;8;N;;;;;
+10D49;GARAY DIGIT NINE;Nd;0;AN;;9;9;9;N;;;;;
+10D4A;GARAY VOWEL SIGN A;Lo;0;R;;;;;N;;;;;
+10D4B;GARAY VOWEL SIGN I;Lo;0;R;;;;;N;;;;;
+10D4C;GARAY VOWEL SIGN O;Lo;0;R;;;;;N;;;;;
+10D4D;GARAY VOWEL SIGN EE;Lo;0;R;;;;;N;;;;;
```
markusicu, line 19343: FYI The proposal calls this "vowel sign EU" as well as "vowel sign EE": https://github.com/unicode-org/sah/issues/55#issuecomment-1766934556

unicodetools/data/ucd/dev/PropList.txt
markusicu: I know, I read that, and the glossary entry. But without contrasting examples, and without knowing how this property is used, I still find these definitions fuzzy. If Diacritic covers any sort of "to which they apply" then there is little difference with gc=Lm. These two length marks certainly don't _look like_ what I think of as a diacritic.

eggrobin: > These two length marks certainly don't _look like_ what I think of as a diacritic.

I don’t think the appearance actually matters to diacriticness; the modification here is linguistic, not graphical stickingoutness.

unicodetools/data/ucd/dev/PropList.txt
eggrobin: Hm, it reduplicates a word rather than a character though.

But  U+3031〱 and U+3032 〲 are also extenders, so this should indeed be an extender. (Assuming their function as described in https://en.wikipedia.org/wiki/Iteration_mark is accurate enough.)

unicodetools/data/ucd/dev/PropList.txt
Ken-Whistler: My recommendations:
10D69 Alphabetic (definitely)
10D6A..10D6D Diacritic
10D6A could also be Extender (for consistency with other gemination marks)
10D6F could be Extender (for consistency with other reduplication marks).
We haven't been entirely consistent with Extender -- could use more review. But this would work for Garay, I think.

eggrobin: I agree with the above, but in addition I would say that 10D69 should be a diacritic, because it serves as a prenasalization mark as well as a vowel sign, and that 10D4E should be a diacritic and an extender (like U+30FC or U+1C7B).

That last one of those raises the question brought up by Markus in another thread here: should U+10D4E be Lm rather than Lo? (I would think so.)

Ken-Whistler: No objection to also making 10D69 Diacritic. It really is a unique animal in this script.
As for 10D4E, I agree that gc=Lm makes a little more sense. Most length marks are gc=Mn or gc=Mc, so formally dependent in the encoding sense, but a standalone letter that is serving as a length mark seems like gc=Lm would suit it better than gc=Lo. The countervailing argument can be seen for tone letters. Tai Le and New Tai Lue have them as gc=Lo, whereas Lisu and Miao have them as gc=Lm. So consistency won't push us hard one way or the other, and there is little in the way of algorithmic consequences that would distinguish Lm from Lo, anyway.

markusicu (approved):
lgtm

markusicu (approved):

PR #562, opened 2023-10-07
Split the UCA checks into their own job, check UCD consistency

eggrobin:
The ucd-and-smoke-tests job will leave an error message in the diff on every file that needs to be regenerated with MakeUnicodeFiles, see https://github.com/unicode-org/unicodetools/pull/562/files/81e0e3708a2e26fe89b46b95b487e44eebc63e67 (scroll to the bottom).

See multiple threads on properties@ and https://github.com/unicode-org/unicodetools/pull/445#issuecomment-1751523905.


markusicu:
major credit for creative commit messages!

eggrobin:
> major credit for creative commit messages!


@robertbastian once told me about [a PR on ICU4X](https://github.com/unicode-org/icu4x/pull/2898) that my commit messages tell a story, and that’s definitely the case here (I was testing with the build on push on eggrobin/unicodetools, but the tree was being checked out from unicode-org/unicodetools, so I was running the Python script without my changes…).

py/copygenerateducd.py
```diff
@@ -17,9 +17,10 @@
 
 
 def main():
+    out_of_source = '--out-of-source' in sys.argv[1:]
```
josh-hadley, line 20: As a reminder for our future selves: if we need to use additional args for this script, we should probably make use of Python's `argparse` module for robust handling. Not a concern for this PR though.

josh-hadley (approved):
😸 LGTM and I'm happy you were able to make use of the Python script

PR #566, opened 2023-10-09
Tulu-Tigalari gemination mark and pluta

eggrobin:
[[173-C27](https://www.unicode.org/cgi-bin/GetL2Ref.pl?173-C27)] Consensus: Accept U+113D2 TULU-TIGALARI GEMINATION MARK and U+113D3 TULU-TIGALARI SIGN PLUTA, with names, glyphs, and properties as given in [L2/22-260](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/22-260), for a future version of the standard. (Ref. section 7 of [L2/22-248](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/22-248)).

Inconsistencies with Ken’s drafts: Ken’s UnicodeData d8 is missing a trailing semicolon on U+113D3 (like the proposal).
Ken's LineBreak d3 gives Pluta lb=ID rather than BA.


eggrobin:
I made them both diacritic extenders, since that felt appropriate for a gemination mark and a vowel length mark.

The Grantha pluta is extender but not diacritic, but that seems like an oddity (the sort of things that feel to me like they should be non-diacritic extenders are reduplication/iteration/repetition marks etc.)

unicodetools/data/ucd/dev/LineBreak.txt
```diff
@@ -2930,6 +2930,8 @@ FFFD           ; AI # So         REPLACEMENT CHARACTER
 11362..11363   ; CM # Mc     [2] GRANTHA VOWEL SIGN VOCALIC L..GRANTHA VOWEL SIGN VOCALIC LL
 11366..1136C   ; CM # Mn     [7] COMBINING GRANTHA DIGIT ZERO..COMBINING GRANTHA DIGIT SIX
 11370..11374   ; CM # Mn     [5] COMBINING GRANTHA LETTER A..COMBINING GRANTHA LETTER PA
+113D2          ; CM # Mn         TULU-TIGALARI GEMINATION MARK
+113D3          ; BA # Lo         TULU-TIGALARI SIGN PLUTA
```
markusicu, line 2934: Why lb=BA "Break After" for this length mark? It looks like other length marks are lb=CM.

markusicu:
It looks like this one wants to be reviewed after the script as a whole is merged. Or maybe its changes want to be added to the main script addition PR.

unicodetools/data/ucd/dev/LineBreak.txt
eggrobin: See discussion with Ken:
> > You gave the Tulu-Tigalari sign pluta lb=ID, but the Grantha sign pluta, mentioned in the proposal, is lb=BA; is there a reason for lb=ID here?
>
> I don't really know what the correct assignments are here. The 3 plutas we have currently are inconsistent:
> 
> 1135D          ; BA # Lo         GRANTHA SIGN PLUTA
> 113D3          ; ID # Lo         TULU-TIGALARI SIGN PLUTA
> 11A9D          ; AL # Lo         SOYOMBO MARK PLUTA
> And there is a similar inconsistency for avagrahas, which are defaulting to AL for most scripts, but:
> 
> 0D3D           ; AL # Lo         MALAYALAM SIGN AVAGRAHA (etc., many others)
> 1133D          ; BA # Lo         GRANTHA SIGN AVAGRAHA
> 113B7          ; ID # Lo         TULU-TIGALARI SIGN AVAGRAHA
> L2/22-080R explicitly suggests AL->BA for the Grantha pluta and the Grantha avagraha. So I presume that both should 
> be treated the same way for Tulu-Tigalari, as well, for consistency. I will update those two.

eggrobin:
> It looks like this one wants to be reviewed after the script as a whole is merged.

Probably.

unicodetools/data/ucd/dev/LineBreak.txt
markusicu: Help me understand. The doc (L2/22-260) says “Pluta denotes ‘extra-normal’ length for a vowel”. The handwritten examples seem to mostly show this in the middle of a word. (The printed example has an arrow pointing to a totally different glyph.) Why should there be a Break After the vowel length mark?

@Ken-Whistler

eggrobin: Note the context of L2/22-080R; this is one of those scripts that break lines within words. We avoid breaking within orthographic syllables (though see the discussion in that doc, that might even be OK for Tulu-Tigalari), so not between the vowel and its lengthener. But after is fine.

markusicu: > Note the context of L2/22-080R; this is one of those scripts that break lines within words.

Thanks, that was what I was missing.

eggrobin:
The Pluta should probably be LB=ID after all. See Norbert's comment on the avagraha:
> For Tulu-Tigalari, one of the experts told us that line breaks can occur pretty much at any spacing character. OpenType doesn’t let you reorder characters across line breaks, and doesn’t deal well with line breaks before combining marks, so we can’t quite replicate that. But avagrahas with gc=Lo are treated as a base, so there’s no problem with breaking before them.

unicodetools/data/ucd/dev/PropertyValueAliases.txt
markusicu, line 1: nit: you could revert this file now that only the time stamp changes

markusicu (approved):

markusicu (approved):

PR #569, opened 2023-10-10
Imbue ARABIC PEPET with the Alphabetic property

eggrobin:
Missed in #435.

See L2/22-116:
> This vowel sign is used to represent vowel /ə/ in Pegon script, and commonly known as
“pepet”, which is cognate with javanese vowel sign pepet (U+A9BC). This vowel sign has
similar shape with arabic maddah above (U+0653), so the users of Pegon script are
commonly using it to write Pepet in the digital text. […]

Both pre-existing characters cited are alphabetic: https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%5CuA9BC%5Cu0653%5D&g=&i=alphabetic


Ken-Whistler (approved):
This change looks correct to me.

Issue #419, opened 2023-03-08
Make the tool parser ignore the conflict marker lines <<<< ==== >>>>


eggrobin:
Discussed in the meeting of 2023-03-07.

Need to either check that the conflicting lines are comments or whitespace, or that they are logically nonconflicting.

eggrobin:
We should also ignore conflicts in PropList.txt:
```
java.lang.reflect.InvocationTargetException
    at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)
    at jdk.internal.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:77)
    at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke (DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke (Method.java:568)
    at org.codehaus.mojo.exec.ExecJavaMojo$1.run (ExecJavaMojo.java:293)
    at java.lang.Thread.run (Thread.java:833)
Caused by: com.ibm.icu.util.ICUException: Prepended_Concatenation_Mark( from: C:\Users\robin\Projects\Unicode\unicodetools\unicodetools\data\ucd\dev\PropList.txt)
    at org.unicode.props.IndexUnicodeProperties.load (IndexUnicodeProperties.java:446)
    at org.unicode.props.IndexUnicodeProperties.loadEnum (IndexUnicodeProperties.java:294)
    at org.unicode.props.IndexUnicodeProperties.loadEnum (IndexUnicodeProperties.java:285)
    at org.unicode.text.UCD.DerivedProperty$13.<init> (DerivedProperty.java:748)
    at org.unicode.text.UCD.DerivedProperty.<init> (DerivedProperty.java:731)
    at org.unicode.text.UCD.DerivedProperty.getCached (DerivedProperty.java:62)
    at org.unicode.text.UCD.DerivedProperty.make (DerivedProperty.java:46)
    at org.unicode.text.UCD.ToolUnicodePropertySource.<init> (ToolUnicodePropertySource.java:899)
    at org.unicode.text.UCD.ToolUnicodePropertySource.make (ToolUnicodePropertySource.java:78)
    at org.unicode.text.UCD.MakeUnicodeFiles.generatePropertyFile (MakeUnicodeFiles.java:1127)
    at org.unicode.text.UCD.MakeUnicodeFiles.generateFile (MakeUnicodeFiles.java:528)
[...]

PR #583, opened 2023-10-24
Ignore conflict markers in ScriptExtensions

eggrobin:
It just so happens that #419 is only _mostly_ fixed.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/jsp/FileUtilities.java
+++ b/unicodetools/src/main/java/org/unicode/jsp/FileUtilities.java
@@ -76,6 +76,11 @@ public final class FileUtilities {
                     if (line == null) {
                         break;
                     }
+                    if (line.startsWith("<<<<<<<")
+                            || line.startsWith("=======")
+                            || line.startsWith(">>>>>>>")) {
+                        continue;
+                    }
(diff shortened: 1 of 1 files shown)


unicodetools/src/main/java/org/unicode/jsp/FileUtilities.java
```diff
@@ -76,6 +76,11 @@ public SemiFileReader process(BufferedReader in, String fileName) {
                     if (line == null) {
                         break;
                     }
+                    if (line.startsWith("<<<<<<<")
+                            || line.startsWith("=======")
+                            || line.startsWith(">>>>>>>")) {
+                        continue;
+                    }
```
markusicu, line 83:
```suggestion
                    // Ignore merge conflict markers.
                    if (line.startsWith("<<<<<<<")
                            || line.startsWith("=======")
                            || line.startsWith(">>>>>>>")) {
                        continue;
                    }
```

markusicu (approved):

eggrobin:
(Note that org.unicode.text.UCD.ScriptExtensions uniquely depends on a parser from the _JSPs_! The rest of the parsers involved in MakeUnicodeFiles are in more reasonable locations.)

Issue #306, opened 2022-08-12
connect the sifter code to the Unicode Tools


markusicu:
The [UCA sifter](https://github.com/unicode-org/unicodetools/tree/main/c/uca/sifter) code that is checked into the Unicode Tools calls into ICU for access to character properties.

This is inconvenient:
- It relies on extra setup to make it build against a recent version of ICU.
- During UCA development, which is when we run the sifter, we need to use the latest draft Unicode properties data. By relying on ICU, we need to first integrate that snapshot of properties data into ICU and rebuild the sifter against that.

Quite cumbersome to do often, which means we don't run this version of the sifter as often as we should.

We should make the sifter independent of ICU, or at least independent of a version of ICU that has the latest draft properties data.

For example, a new Unicode Tool could write `*_data.h` files with data initializers for the data of relevant properties, and the sifter code could either use that data directly, or build ICU tries from it on the fly.

@Ken-Whistler @macchiati FYI

markusicu:
Wish list:
- Write a `*_data.h` file, e.g., `ucd_data.h`, with C initializers. Include that into [unisifex.c](https://github.com/unicode-org/unicodetools/blob/main/c/uca/sifter/unisifex.c). Check it into the same sifter folder.
- For each boolean unisifex “property”, write an inversion list of range limit code points like in ICU UnicodeSet.
- In unisifex.c, write a function like [`UnicodeSet::findCodePoint(c)`](https://github.com/unicode-org/icu/blob/main/icu4c/source/common/uniset.cpp#L361-L404) to do the lookup. Plus [the last few lines from `UnicodeSet::contains(c)`](https://github.com/unicode-org/icu/blob/main/icu4c/source/common/uniset.cpp#L354-L358) which is just above findCodePoint().
- Sample C++ code for writing an array of values: ICU [toolutil/writesrc.cpp usrc_writeArray()](https://github.com/unicode-org/icu/blob/main/icu4c/source/tools/toolutil/writesrc.cpp#L148)
- There are some unisifex.c functions that map a code point to an integer value (e.g., Numeric_Value, which might be used only for gc=Nd). We should write an inversion _map_ for each of those. Array of (range limit, value) pairs for consistency with the ICU inversion list, or maybe (range start, value). Maybe like [struct CheckRange](https://github.com/unicode-org/icu/blob/main/icu4c/source/test/cintltst/ucptrietest.c#L32-L43) in some ICU test code.

[...]

PR #584, opened 2023-10-24
A standalone C++ unisifex

eggrobin:
See https://github.com/unicode-org/unicodetools/issues/306#issuecomment-1747114315.


c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,433 @@
+#include <algorithm>
```
markusicu, line 1:
```suggestion
// © 2023 and later: Unicode, Inc. and others.
// License & terms of use: http://www.unicode.org/copyright.html
#include <algorithm>
```

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,433 @@
...
+  ~CheckFailure() {
+    std::terminate();
+  }
+  void operator&(std::ostream&) {}
+};
+
+#define CHECK(condition) \
+  (condition)            \
+      ? (void)0          \
+      : CheckFailure() & std::cerr << ("Check failed: " #condition "\n")
```
markusicu, line 23:
```suggestion
#define CHECK(condition) \
  ((condition)            \
      ? (void)0          \
      : CheckFailure() & std::cerr << ("Check failed: " #condition "\n"))
```
... so that, if someone accidentally puts this into a larger expression, we don't get funny operator precedence effects.

c/uca/sifter/unisifeggs.cpp
markusicu, line 1: I assume that this wants to replace the contents of unisifex.c (and rename that to .cpp), rather than a whole new file name / basename.

Maybe first a PR to rename the existing file to .cpp, and then rebase this on top of that and replace the contents?

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,433 @@
...
+
+class CodePointRange {
+ public:
+  constexpr CodePointRange(char32_t front, char32_t back)
+      : front_(front), back_(back) {
+    if (empty()) {
+      front_ = 1;
+      back_ = 0;
+    }
```
markusicu, line 32: Suggestion:
- rename "front" to "start"
- rename "back" to "end"
- internally store "start_" and "limit_" where limit_ = end + 1
- empty() is start_ == limit_ or maybe start_ >= limit_

This from experience working with the API and innards of UnicodeSet. Exclusive-end "limit" values tend to be easier to work with than inclusive-end "end" values.

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,433 @@
...
+  }
+
+  constexpr bool empty() const {
```
markusicu, line 35: Let's please call this isEmpty(). empty() sounds like a command to make it so.

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,433 @@
...
+#include <iostream>
+#include <map>
+#include <ranges>
+#include <string>
+#include <string_view>
+#include <vector>
+
+class CheckFailure {
+ public:
+  ~CheckFailure() {
```
markusicu, line 14: Would you mind using an indentation of 4 spaces per level?

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,433 @@
...
+  constexpr bool contains(char32_t c) const {
+    return c >= front_ && c <= back_;
+  }
+
+  constexpr CodePointRange intersection(CodePointRange const& other) const {
+    return CodePointRange(std::max(front_, other.front_),
+                          std::min(back_, other.back_));
+  }
+
+  constexpr char32_t front() const {
```
markusicu, line 48: ... getStart(), getEnd(), maybe even getLimit() if there are call sites that add 1 ...

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,433 @@
...
+  iterator end() const {
+    return {back_ + 1};
+  }
+
+ private:
+  char32_t front_;
+  char32_t back_;
+};
+
+constexpr CodePointRange convex_hull(CodePointRange const& r1,
```
markusicu, line 84: I prefer camelCase()

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,433 @@
...
+
+  bool contains(char32_t c) const {
+    auto const it = std::partition_point(
+        ranges_.begin(), ranges_.end(), [c](CodePointRange const& range) {
+          return range.back() < c;
+        });
+    return it != ranges_.end() && it->contains(c);
+  }
+
+  void Include(CodePointRange range) {
```
markusicu, line 103: addRange(range)

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,433 @@
...
+    auto it = std::partition_point(
+        ranges_.begin(), ranges_.end(), [&range](CodePointRange const& r) {
+          return r.back() < range.front();
+        });
+    if (it == ranges_.end() || it->intersection(range).empty()) {
```
markusicu, line 108: optional: The second test could be simpler/faster `range.getEnd() < it->getStart()`, right?

Also, if you were testing `range.getLimit() < it->getStart()` here (limit not end), then the immediately-adjacent case would automatically go to the "else" branch and simplify your "if-then" branch.

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,433 @@
...
+      if (it != ranges_.begin() && (it - 1)->back() + 1 == it->front()) {
+        *(it - 1) = convex_hull(*(it - 1), *it);
+        ranges_.erase(it);
+      }
+    } else {
+      *it = convex_hull(*it, range);
+    }
+  }
+
+  CodePointSet Include(CodePointSet const& other) {
```
markusicu, line 123: addAll(other)

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,433 @@
...
+constexpr std::array<std::string_view, n> fields(std::string_view const line) {
+  std::array<std::string_view, n> result;
+  auto split = std::views::split(line, ';');
+  auto it = split.begin();
+  for (auto& field : result) {
+    CHECK(it != split.end()) << line;
+    std::string_view field_with_spaces(*it++);
+    auto const begin = std::min(field_with_spaces.find_first_not_of(" "),
+                                field_with_spaces.size());
+    auto const end = field_with_spaces.find_last_not_of(" ") + 1;
```
markusicu, line 148: s/end/limit/
(because exclusive)

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,433 @@
...
+}
+
+constexpr char32_t parse_hex_codepoint(std::string_view hex) {
+  std::uint32_t c;
+  std::from_chars(hex.data(), hex.data() + hex.size(), c, 16);
+  return static_cast<char32_t>(c);
+}
+
+constexpr CodePointRange parse_hex_codepoint_range(std::string_view hex) {
+  auto const front_end = hex.find("..");
```
markusicu, line 161: confusing... I would call it "pos"

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,433 @@
...
+  if (front_end == std::string_view::npos) {
+    char32_t const front = parse_hex_codepoint(hex);
```
markusicu, line 163: s/front/start/

markusicu (commented):
I got to just before "class UCD" before dinner...

c/uca/sifter/unisifeggs.cpp
eggrobin: No, this is intentional so you can output additional context into it (see the glog macros).

c/uca/sifter/unisifeggs.cpp
eggrobin: `begin`/`end` are the idiomatic terms for that in C++, see the containers and ranges of the standard library and range-based for loops.

c/uca/sifter/unisifeggs.cpp
eggrobin: The name should be `front` to match conventions for C++ sequence containers (`front()` for the first element). See https://en.cppreference.com/w/cpp/named_req/SequenceContainer.

c/uca/sifter/unisifeggs.cpp
eggrobin: https://en.cppreference.com/w/cpp/named_req/Container.

c/uca/sifter/unisifeggs.cpp
eggrobin: I’m not sure I want to get rid of the existing one until we know other people can run it, and there is little enough in common that having both is fine.

(The name may be slightly silly. Then again, we are not above bags of walnuts.)

c/uca/sifter/unisifeggs.cpp
eggrobin: See my other comments (though would be fine with storing begin_ and end_ rather than front_ and back_).
`begin` and `end` in particular are needed and need to have the right semantics for range-based for loops to work (but more generally, having the right names is required for code that works with containers to work with this).

c/uca/sifter/unisifeggs.cpp
eggrobin: Yeah this is not a good name here (it is not even the end of the text representation of the variable called `front` below!).

c/uca/sifter/unisifeggs.cpp
markusicu: The more the C++ standard progresses, the less I like C++ ... :-(

c/uca/sifter/unisifeggs.cpp
markusicu: Gosh these names are bad. push_back() makes some sense, where "back" indicates *where* you push, not *what* you push. front() and back() for the first/last elements are just sucky names...

c/uca/sifter/unisifeggs.cpp
markusicu: Grumble C++ STL grumble...

Please
- name the fields start_ and end_
- add a limit() function that returns end_ + 1, and use that where it makes sense

c/uca/sifter/unisifeggs.cpp
markusicu: Using "end" to mean both inclusive and exclusive ends is confusing. Where we are not bound by bad STL naming conventions, such as in variable names, I would prefer being unambiguous.

eggrobin: `end` never means inclusive end in C++ naming conventions; I would be deeply confused if it were used that way.

I have no intuition of what `limit` means, and in particular I would be confused as to whether it is an inclusive or exclusive upper bound.

c/uca/sifter/unisifeggs.cpp
eggrobin: Note that I changed the type to be an iterator, rather than a char32_t; so using iterator terminology probably makes sense (the choice of type also meants that if you construct from a range [c, d[, you need to construct it from iterators rather than code points, which makes it clearer that the second bound is excluded).

eggrobin: Then again it means I am mixing iterators across range objects in some operations, which feels wrong. I will think about that some more another day.

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,440 @@
...
+                auto const record = strip_comment(line);
+                if (record.empty()) {
+                    continue;
+                }
+                auto const [cp, decimal_nv, _, rational_nv] = fields<4>(record);
+                auto const range = parse_hex_codepoint_range(cp);
+                code_points_with_numeric_value_.Include(range);
+                if (rational_nv.find_first_not_of("0123456789") ==
+                    std::string_view::npos) {
+                    for (char32_t const code_point : range) {
```
markusicu, line 240:
```suggestion
                if (rational_nv.find_first_not_of("0123456789") ==
                        std::string_view::npos) {
                    for (char32_t const code_point : range) {
```

markusicu: ... or just put it on the same line. Line width of 100 should be fine.

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,440 @@
...
+            return coarse_general_category_.at(gc.front());
+        }
+        auto it = general_category_.find(gc);
+        CHECK(it != general_category_.end()) << gc;
+        return it->second;
+    }
+
+    CodePointSet const& BinaryPropertySet(
+        std::string_view const binary_property_name) {
+        auto it = binary_properties_.find(binary_property_name);
```
markusicu, line 260:
```suggestion
    CodePointSet const& BinaryPropertySet(
            std::string_view const binary_property_name) {
        auto it = binary_properties_.find(binary_property_name);
```
... or put the whole function header on one line

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,440 @@
...
+    // Default 0 omitted.
+    std::map<char32_t, std::uint8_t> canonical_combining_class_;
+    std::map<char32_t, std::uint64_t> natural_numeric_value_;
+    // Characters with a non-NaN Numeric_Value.
+    CodePointSet code_points_with_numeric_value_;
+};
+
+typedef unsigned short UShort16;
+
+typedef unsigned int UInt32;
```
markusicu, line 350:
```suggestion
typedef uint16_t UShort16;

typedef uint32_t UInt32;
```

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,440 @@
...
+
+UCD* const ucd = new UCD();
+
+static bool isSifterNonSpacingCombining(char32_t c) {
```
markusicu, line 354: Rather than "static", I would put everything in this file after the includes and before the `extern "C"` functions into an anonymous namespace:

```
namespace {

// no additional indent

}  // namespace
```

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,440 @@
...
+    return 0xFA0E <= c && c <= 0xFA29 &&
+           (c <= 0xFA0F || c == 0xFA11 || c == 0xFA13 || c == 0xFA14 ||
+            c == 0xFA1F || c == 0xFA21 || c == 0xFA23 || c == 0xFA24 ||
+            0xFA27 <= c);
+}
+
+extern "C" int unisift_IsAlphabetic(UInt32 c) {
+    return ucd->BinaryPropertySet("Alphabetic").contains(c) &&
+           !isSpecialCJKIdeograph(c) &&
+           !(c == 0x1E13C || c == 0x1E13D || c == 0x16B42 || c == 0x16B43);
```
markusicu, line 371:
```suggestion
    return ucd->BinaryPropertySet("Alphabetic").contains(c) &&
            // HACK: TODO: Deal with the 12 special CJK ideographs
            // ("compatibility" but no decompositions) in sift() before
            // testing Alphabetic.
            !isSpecialCJKIdeograph(c) &&
            // Ken: Exceptions for gc=Lm extenders to match his sifter.
            // TODO: Look to modify this behavior for gc=Lm extenders in the future.
            !(c == 0x1E13C || c == 0x1E13D || c == 0x16B42 || c == 0x16B43);
```
(copied from unisifex.c)

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,440 @@
...
+
+extern "C" int unisift_ToIntValue(UInt32 c) {
+    // TODO(egg): The conversion to the return type is UB for large numbers.
+    return ucd->NaturalNumericValue(c).value_or(
+        ucd->CodePointsWithNumericValue().contains(c) ? -2 : -1);
+}
+
+extern "C" int unisift_GetCombiningClass(UInt32 c) {
+    return ucd->CanonicalCombiningClass(c);
+}
```
markusicu, line 440: Don't you also need unisift_unistrcpy() and unisift_unistrcat()?
They were in Ken's version of unisifex.c, and I assumed that the sifter code calls them.

c/uca/sifter/unisifeggs.cpp
eggrobin: Can you live with the names `first`, `past_the_end` (setting aside the 🐪 vs. 🐫 vs. 🐍 discussion for a moment), and `last` when we have non-API names corresponding to the standard `*begin=front`, `end`, and `back=*--end`?

That is consistent with wording in the standard, specifically, 
https://eel.is/c++draft/container.requirements.general#container.reqmts-28,
https://eel.is/c++draft/container.requirements.general#container.reqmts-31, and
https://eel.is/c++draft/container.requirements#sequence.reqmts-115,
and I think it should be clearer to the reader unfamiliar with standard library naming conventions.

markusicu: first and last would be ok.

I could live with past_the_end because it's clear, but it's clunky. We have been using "limit" in ICU for nearly 30 years...

markusicu: (rant)
Gotta love this one:
> https://eel.is/c++draft/container.requirements#container.reqmts-31

“**b.end()** ... Returns: An iterator which is the **past-the-end** value for the container.”

So end() does not return an iterator *at* the end, it returns one *past the end*. Then why not call it for what it does??
https://www.google.com/search?q=dunce+corner&tbm=isch
(/rant)

eggrobin: > (rant)

Yeah, it’s silly. (_end_, from _past-the-end_ by apheresis, like _bus_ from _omnibus_? 🙃)

c/uca/sifter/unisifeggs.cpp
eggrobin: Apparently it does not.

c/uca/sifter/unisifeggs.cpp
eggrobin: This was formatted with clang-format with Google settings; I would very much like not to fight the autoformatter.

c/uca/sifter/unisifeggs.cpp
```diff
@@ -41,53 +41,56 @@ class CodePointRange {
...
     constexpr bool contains(char32_t c) const {
-        return iterator{c} >= begin_ && iterator{c} < end_;
+        return c >= first_ && c < past_the_end_;
     }
 
     constexpr CodePointRange intersection(CodePointRange const& other) const {
-        return CodePointRange(std::max(begin_, other.begin_),
-                              std::min(end_, other.end_));
+        return CodePointRange({std::max(first_, other.first_)},
+                              {std::min(past_the_end_, other.past_the_end_)});
```
markusicu, line 58: What do the curly braces do here? Leftover from when you stored iterators?

c/uca/sifter/unisifeggs.cpp
markusicu: I guess I don't understand then. Inside Google, the formatter uses a double indent for continuing an if condition, for loop head, or function args list, so that there is a step back to the body of that statement/definition. It's hard to make out if the head continuation indent is at the same level as the body.

markusicu:
Thanks for the new changes! I resolved comments that you took care of.

c/uca/sifter/unisifeggs.cpp
eggrobin: Looking at it again, this code seems all wrong anyway.

eggrobin: Rewrote.

c/uca/sifter/unisifeggs.cpp
eggrobin: Ah yes, it seems increasing the indent size to 4 spaces did not increase that to more than 4…

eggrobin: I increased ContinuationIndentWidth, but that means all line continuations are very indented.

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,457 @@
...
+    bool contains(char32_t c) const {
+        const auto it = std::partition_point(
+                ranges_.begin(),
+                ranges_.end(),
+                [c](const CodePointRange& range) { return range.back() < c; });
+        return it != ranges_.end() && it->contains(c);
+    }
+
+    void addAll(CodePointRange range) {
+        const auto firstModifiedRange =
```
markusicu, line 104:
```suggestion
        // All earlier ranges end before the new one and are not adjacent.
        const auto firstModifiedRange =
```

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,457 @@
...
+                std::partition_point(ranges_.begin(),
+                                     ranges_.end(),
+                                     [&range](const CodePointRange& r) {
+                                         return r.back() + 1 < range.front();
+                                     });
+        const auto pastModifiedRanges =
```
markusicu, line 110:
```suggestion
        // First range that starts after the new one and is not adjacent.
        const auto pastModifiedRanges =
```

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,457 @@
...
+                                     [&range](const CodePointRange& r) {
+                                         return r.front() <= range.back() + 1;
+                                     });
+        char32_t insertedRangeFirst = range.front();
+        if (firstModifiedRange != ranges_.end()) {
+            insertedRangeFirst =
+                    std::min(firstModifiedRange->front(), insertedRangeFirst);
+        }
+        char32_t insertedRangeLast = range.back();
+        if (pastModifiedRanges != ranges_.begin()) {
```
markusicu, line 122: Should you check against firstModifiedRange and ranges_.end() here?
I don't have a complete image in my head of the possible combinations, but with firstModifiedRange <= pastModifiedRanges <= ranges_.end()
- if firstModifiedRange == pastModifiedRanges then we should not go to pastModifiedRanges - 1
- if pastModifiedRanges == ranges_.end() then we must not read its fields

right??

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,457 @@
...
+static bool isSpecialCJKIdeograph(char32_t c) {
+    return 0xFA0E <= c && c <= 0xFA29 &&
+           (c <= 0xFA0F || c == 0xFA11 || c == 0xFA13 || c == 0xFA14 ||
+            c == 0xFA1F || c == 0xFA21 || c == 0xFA23 || c == 0xFA24 ||
+            0xFA27 <= c);
+}
+
+}  // namespace
+
+extern "C" int unisift_IsAlphabetic(std::uint32_t c) {
```
markusicu, line 378: nit: I have never seen uint32_t qualified with `std::` ...

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,457 @@
...
+            ++value;
+        }
+
+        constexpr std::strong_ordering operator<=>(
+                const iterator& other) const = default;
+
+        char32_t value;
+    };
+
+    static constexpr CodePointRange Inclusive(char32_t first, char32_t last) {
```
markusicu, line 46:
```suggestion
    static constexpr CodePointRange inclusive(char32_t first, char32_t last) {
```
lowerCamelCase

c/uca/sifter/unisifeggs.cpp
eggrobin: It depends whether you get them from the C++ header or from the C header; I’m doing it the C++ way.

c/uca/sifter/unisifeggs.cpp
```diff
@@ -0,0 +1,457 @@
...
+
+    void addAll(CodePointRange range) {
+        const auto firstModifiedRange =
+                std::partition_point(ranges_.begin(),
+                                     ranges_.end(),
+                                     [&range](const CodePointRange& r) {
+                                         return r.back() + 1 < range.front();
+                                     });
+        const auto pastModifiedRanges =
+                std::partition_point(ranges_.begin(),
```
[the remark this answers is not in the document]
eggrobin: Probably. I’ll take another look Monday.

c/uca/sifter/unisifeggs.cpp
eggrobin: > if firstModifiedRange == pastModifiedRanges then we should not go to pastModifiedRanges - 1

Hmm. I need to think. I also need to come up with a way to have some tests for this thing…

> if pastModifiedRanges == ranges_.end() then we must not read its fields

Indeed, but we don’t: we only ever look at *(pastModifiedRanges - 1).

eggrobin: > Hmm. I need to think. I also need to come up with a way to have some tests for this thing…

I did add some tests, but I have yet to think. I will take another look some other day.

c/uca/sifter/ucd_test.cpp
```diff
@@ -0,0 +1,47 @@
+// © 2024 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+#include "ucd.hpp"
+
+namespace {
+static int testFailures = 0;
```
markusicu, line 6:
```suggestion
int testFailures = 0;
```

c/uca/sifter/ucd.cpp
```diff
@@ -0,0 +1,180 @@
...
+                        << line << "\n"
+                        << nextLine;
+                range = CodePointRange::inclusive(codePoint,
+                                                  parseHexCodePoint(lastCp));
+            }
+            coarseGeneralCategory_[gc.front()].addAll(range);
+            generalCategory_[std::string(gc)].addAll(range);
+            for (char32_t c : range) {
+                canonicalCombiningClass_.emplace(c,
+                                                 std::stoi(std::string(ccc)));
```
markusicu, line 58: optional: this should fit onto one line; allow 100 chars per line?

c/uca/sifter/ucd.cpp
```diff
@@ -0,0 +1,180 @@
...
+            const auto record = stripComment(line);
+            if (record.empty()) {
+                continue;
+            }
+            const auto [cp, decimalNV, _, rationalNV] = fields<4>(record);
+            const auto range = parseHexCodePointRange(cp);
+            codePointsWithNumericValue_.addAll(range);
+            if (rationalNV.find_first_not_of("0123456789") ==
+                std::string_view::npos) {
+                for (const char32_t codePoint : range) {
```
markusicu, line 86:
```suggestion
            if (rationalNV.find_first_not_of("0123456789") ==
                    std::string_view::npos) {
                for (const char32_t codePoint : range) {
```

c/uca/sifter/ucd.cpp
```diff
@@ -0,0 +1,180 @@
...
+                    naturalNumericValue_.emplace(
+                            codePoint, std::stoll(std::string(rationalNV)));
```
markusicu, line 88: please take the string-to-number conversion out of the loop

c/uca/sifter/ucd.hpp
```diff
@@ -0,0 +1,244 @@
...
+    constexpr CodePointRange(char32_t first, char32_t past_the_end)
+        : first_(first), pastTheEnd_(past_the_end) {}
+
+    char32_t first_;
+    char32_t pastTheEnd_;
+};
+
+// TODO(egg): Make this behave like a container.
+class CodePointSet {
+   public:
```
markusicu, line 95: An indentation of three looks very weird. I prefer labels without indentation. Indenting by half of the normal might also be ok.
```suggestion
public:
```

c/uca/sifter/ucd.hpp
```diff
@@ -0,0 +1,244 @@
...
+                                     });
+        char32_t insertedRangeFirst = range.front();
+        if (firstModifiedRange != ranges_.end()) {
+            insertedRangeFirst =
+                    std::min(firstModifiedRange->front(), insertedRangeFirst);
+        }
+        char32_t insertedRangeLast = range.back();
+        if (pastModifiedRanges != ranges_.begin() &&
+            pastModifiedRanges - 1 >= firstModifiedRange) {
+            const auto lastModifiedRange = pastModifiedRanges - 1;
```
markusicu, line 140:
```suggestion
        if (pastModifiedRanges != ranges_.begin() &&
                pastModifiedRanges - 1 >= firstModifiedRange) {
            const auto lastModifiedRange = pastModifiedRanges - 1;
```

markusicu (commented):
very nice!

c/uca/sifter/ucd.cpp
eggrobin: I really don’t want to fiddle with formatting by hand on this. I’ll see if I can get the .clang-format configuration to do something like that.

c/uca/sifter/ucd.hpp
eggrobin: Ditto, will look at doing that in .clang-format.

c/uca/sifter/ucd.cpp
eggrobin: I got the private:/public: thing to work, but not this.

c/uca/sifter/ucd.cpp
eggrobin: OBE.

c/uca/sifter/ucd.hpp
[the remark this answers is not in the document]
eggrobin: Not the most elegant thing, then again file extensions rarely shine by their elegance…
I ran into trouble with stuff (including GitHub) thinking C++ files were C, but admittedly that was 10 years ago. Then again, we do have C sitting next to C++ here, so having a distinction may be nice for the humans too.

c/uca/sifter/unisifeggs.cpp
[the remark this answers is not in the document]
markusicu: FYI: I didn't try to think through all of the range-merge logic again this time.

c/uca/sifter/ucd.cpp
markusicu: I am confused. the format options claim that they are based on Google style, and Google style is to double the indentation so that the continuation of the if/while/for is indented more than its body and doesn't visually blend in with that.

But this is not a blocker.

markusicu (approved):

c/uca/sifter/ucd.cpp
eggrobin: > Google style is to double the indentation so that the continuation of the if/while/for is indented more than its body and doesn't visually blend in with that.

The style guide has at least one example which aligns things after the ( of an if: https://google.github.io/styleguide/cppguide.html#Boolean_Expressions

markusicu: That example shows the condition-continuation with twice the indent of the if-body. In default Google style, body indent is 2 and condition-continuation is 4. Double those for more conventional indentation, and you get 4 and 8.

eggrobin: Right, the example is ambiguous as to whether is indented to the continuation indent or to the (. It appears that the clang-format implementation thinks it is the latter, and I think Google folks maintain that implementation. 🤷.

eggrobin: Also note the example below in https://google.github.io/styleguide/cppguide.html#Return_Values.

PR #585, opened 2023-10-24
Check in my UCD checklist

eggrobin:
Included @markusicu’s translation from https://github.com/unicode-org/unicodetools/pull/375#issuecomment-1758572597, all other scripts are Windows in-source.


markusicu (approved):
lgtm tnx!

PR #603, opened 2023-11-09
Indic_Positional_Category for Kirat Rai

eggrobin:
It turns out we want it, see https://github.com/unicode-org/unicodetools/pull/445#issuecomment-1798359742.

This is consistent with Roozbeh’s file.

I will fix the linewrapping of the header later, in the Gurung Khema PR.


roozbehp:
LGTM

roozbehp (approved):

roozbehp (approved):

PR #606, opened 2023-11-12
Add new Myanmar digits to IndicSyllabicCategory.txt

roozbehp:
[171-C20] Consensus: Accept 20 Myanmar numerals (U+116D0…U+116E3) for Eastern Pwo Karen and Pa’O in a new Myanmar Extended-C block, with range U+116D0..U+116FF for a future version of the standard. (Reference L2/22-046)

The characters have been added, but we had missed adding them to IndicSyllabicCategory.txt.


Ken-Whistler (approved):
Looks correct.

PR #608, opened 2023-11-14
Fix InSC of U+16D43 KIRAT RAI LETTER A to Vowel_Independent

roozbehp:
The proposal doesn't request syllabic categories, but we decided to assign them. It looks like we overlooked that the first letter is an independent vowel, not a consonant.

### DIFF
--- a/unicodetools/data/ucd/dev/IndicSyllabicCategory.txt
+++ b/unicodetools/data/ucd/dev/IndicSyllabicCategory.txt
@@ -1,5 +1,5 @@
 # IndicSyllabicCategory-16.0.0.txt
-# Date: 2023-11-13, 19:36:00 GMT
+# Date: 2023-11-14, 01:26:00 GMT
 # © 2023 Unicode®, Inc.
 # Unicode and the Unicode Logo are registered trademarks of Unicode, Inc. in the U.S. and other countries.
 # For terms of use, see https://www.unicode.org/terms_of_use.html
@@ -482,6 +482,7 @@ ABD1          ; Vowel_Independent # Lo       MEETEI MAYEK LETTER ATIYA
 11D6A..11D6B  ; Vowel_Independent # Lo   [2] GUNJALA GONDI LETTER OO..GUNJALA GONDI LETTER AU
 11F04..11F10  ; Vowel_Independent # Lo  [13] KAWI LETTER A..KAWI LETTER O
 16100         ; Vowel_Independent # Lo       GURUNG KHEMA LETTER A
+16D43         ; Vowel_Independent # Lo       KIRAT RAI LETTER A
 
 # ================================================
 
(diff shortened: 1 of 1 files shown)


eggrobin:
I think that was a conscious choice I made based on my understand of the proposal and by analogy to other scripts. Of course it is quite possible that I got that wrong, but I would like to take a closer look.

Ideally not while I am on vacation, I am trying to limit my Unicode work to cuneiform until I am back.

Ken-Whistler (approved):
I think the change is o.k. It reflects the explicit claim in the proposal. In any case, I've updated the subheads in the 16.0 names list draft to track this.

markusicu:
Robin requested that we wait until he has time to compare with his previous notes. Let's give him a few days.

eggrobin (changes requested):
I gave it InSC=Consonant by analogy with U+11EF1 𑻱 MAKASAR LETTER A and others. However, we are inconsistent here.

L2/22-043R reads:
> There is a single vowel letter which functions as a vowel carrier. Other vowels are represented as dependent signs (see
Fig. 1, 5a, 6, 7, 10). Independent forms of other vowels are produced by writing those vowels with this letter.

This is the same as for Makasar, on which the Standard reads:
> There is only one independent vowel sign, U+11EF1 𑻱 MAKASAR LETTER A. Vowel signs
can be attached to this character to produce other vowel sounds when a syllable has no
consonant, such as at the beginning of a word.

Or for Ahom:
> Ahom has no independent vowels, but instead uses U+11712 AHOM LETTER A followed by the corresponding dependent vowel sign (or signs).

However, other scripts do the same thing and have their A as an Independent_Vowel instead:
> The vowel _a_ is represented by U+11A50 SOYOMBO LETTER A.
When it occurs with a vowel sign, SOYOMBO LETTER A serves as a vowel-carrier, indicating
an independent vowel.

> The Zanabazar Square script has one vowel letter, nine dependent vowel marks, and one vowel length mark. The letter a vowel, U+11A00 ZANABAZAR SQUARE LETTER A, has the value /a/ when it occurs independently. It can also assume the
value of a combined vowel sign.

It sounds like we should decide what we want to do for those kinds of As, fix the ones that do the other thing (that will need a UTC decision, likely based on a PAG recommendation), and improve the documentation of InSC to mention that.

eggrobin:
Filed https://github.com/unicode-org/properties/issues/203, let’s discuss it there.

markusicu:
Is this still waiting on getting https://github.com/unicode-org/properties/issues/203 resolved?

eggrobin:
I think so.

markusicu:
FYI, pending change still waiting for resolution of https://github.com/unicode-org/properties/issues/203 which got punted beyond Unicode 16.

eggrobin:
We are not doing this, instead UTC decided to
[UTC-181-C44] Consensus: Change the Indic_Syllabic_Category of U+11A50 SOYOMBO LETTER A and U+11A00 ZANABAZAR SQUARE LETTER A from Vowel_Independent to Consonant, and that of U+1900 LIMBU VOWEL-CARRIER LETTER from Consonant_Placeholder to Consonant, for Unicode 17.0 [Ref: 4.6 in L2/24-228]

I am therefore closing this PR.

PR #609, opened 2023-11-18
U+19DA is DISALLOWED because XV8

markusicu:
Fixes a derived-data bug caused by a tooling error which had set the idna2008derived status of U+19DA to be PVALID.

In IdnaMappingTable.txt U+19DA has an IDNA2008 Status of XV8: https://www.unicode.org/reports/tr46/#Table_Data_File_Fields
“XV8 is present when the character is excluded by IDNA2008 for the current version of Unicode.”

Therefore, in the idna2008derived files it needs to be DISALLOWED.

----
All files copied verbatim today from https://www.unicode.org/Public/idna/idna2008derived/
except for restoring the template placeholders in the ReadMe.


Ken-Whistler (approved):
As expected.

PR #610, opened 2023-11-18
NamesList 16.0 d3

markusicu:
from @Ken-Whistler 2023-nov-03


Ken-Whistler (approved):
Changes look correct.

PR #611, opened 2023-11-18
Unihan 16.0 nov06

markusicu:
Unihan data verbatim from @kenlunde oct21 & nov06


kenlunde:
BTW, the first line of the header of `USourceData.txt` will be corrected in the next drop to reflect Version 16.0.0, so for the purposes of this PR, I think that this difference can be ignored.

Manishearth (approved):

PR #613, opened 2023-11-24
Update ScriptExtensions.txt per L2/23-280

roozbehp:
To see what can break if we add Script_Extensions for characters in the 0300 block.


roozbehp:
Messed this one up. Retrying with https://github.com/unicode-org/unicodetools/pull/614

PR #614, opened 2023-11-24
Update ScriptExtensions.txt per L2/23-280 except for U+0300 and U+0308

roozbehp:
[[178-C39](https://www.unicode.org/cgi-bin/GetL2Ref.pl?178-C39)] Consensus: Add 51 entries to ScriptExtensions.txt as proposed in [L2/23-280](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/23-280) for a future version of the Unicode Standard. [Ref. Section 18 of [L2/24-013R](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/24-013R)]

[[178-A99](https://www.unicode.org/cgi-bin/GetL2Ref.pl?178-A99)] Action Item for Roozbeh Pournader, UTC: Add 51 entries to ScriptExtensions.txt as proposed in document [L2/23-280](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/23-280) for a future version of the Unicode Standard. [Ref. Section 18 of [L2/24-013R](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/24-013R)]

All changes proposed in L2/23-280 are incorporated in this pull request, except for U+0300 and U+0308.


macchiati:
I'm concerned about this. The following for example narrows the scope of the characters drastically.

```
02C7          ; Bopo Latn # Lm       CARON
02C9..02CB    ; Bopo Latn # Lm   [3] MODIFIER LETTER MACRON..MODIFIER LETTER GRAVE ACCENT
02D9          ; Bopo Latn # Sk       DOT ABOVE
```

One would one want to make changes like this is one we're confident that these characters were not used or very rarely used in all other scripts Cyrillic, Greek, etc.

roozbehp:
@macchiati FWIW, this whole effort was based on your document L2/17-430. I got action items to complete that data, it took a few years, and we're finally here with almost-complete information.

> I'm concerned about this. The following for example narrows the scope of the characters drastically.
> 
> [...]

If you are concerned about specific characters in the list, we can always remove them. I intentionally went inclusive so we can figure out what types of things may cause alarms and what kind of thing are not controversial. Please take a look at the rest of the data and tell me what else concerns you.

> One would one want to make changes like this is one we're confident that these characters were not used or very rarely used in all other scripts Cyrillic, Greek, etc.

To the best of my effort, I did the work to make sure about this. I combed all of the Core Spec and a ton of L2 documents for this. I believe I have used the same standard used for the existing characters in ScriptExtensions, which is mentioned in UAX#24: "The Script_Extensions property is primarily targeted at customary modern use of characters, and does not encompass technical usage such as phonetic transcriptional systems or mathematics."

macchiati:
I would not (at all) depend on L2 docs and the core spec for determining when to eliminate scripts from Common symbol, punctuation, mark characters that are not closely associated with a given script. The right approach, I think, would be to look at crawls of large corpora of particular scripts. If a search of all Cyrillic text in Wikipedia pages, for example, yields no instance of DOT ABOVE (or very very few as a percentage), then it would be clearly safe to drop Cyrillic; same for other scripts.

macchiati:
Note: one thing to look for as you are working on Script Extensions is cases where the scx value is a single script, and the sc value is Common or Inherited.
Example:
᳷ 	[U+1CF7](https://util.unicode.org/UnicodeJsps/character.jsp?a=1CF7)	VEDIC SIGN ATIKRAMA

markusicu:
I discussed this yesterday with a few straggler UTCers: https://github.com/unicode-org/sah/issues/375#issuecomment-1912595585

markusicu:
@roozbehp do you have a record of what unit test failures you got with the full set of scx in your doc? What are the failing strings, and what are the errors?

macchiati:
We just need to find a reasonable balance. If we give scx values to a
character (that was formerly Common or Inherited) we are effectively saying
that we don't believe it is in common usage in any other script other than
the scx values. For text that contains that character, it will tend to
improve the rendering and processing of that text when it is in one of the
scx scripts, and worsen the rendering and processing of that text when it
is in other scripts.

NOTE: if the character is a Combining Mark, then the impact is *far* less
important — either way — because the character inherits the script of what
it is applied to. https://unicode.org/reports/tr24/#Nonspacing_Marks. So it
doesn't matter much for those characters except in degenerate cases (where
the character is applied to a base character that doesn't itself have an
explicit script or scx value.

So we neither want to be too onerous in requirements, nor too strict (after
all, we can adjust later on).

markusicu:
I think it would be better to keep most of this discussion in one place, probably in the SAH issue.

> NOTE: if the character is a Combining Mark, then the impact is *far* less important — either way — because the character inherits the script of what it is applied to. https://unicode.org/reports/tr24/#Nonspacing_Marks. So it doesn't matter much for those characters except in degenerate cases (where the character is applied to a base character that doesn't itself have an explicit script or scx value.

True for UAX24, but by my read of UTS39 mixed-script detection, the General_Category is ignored there, and missing Recommended Scripts from scx would be inconvenient for those.

> We just need to find a reasonable balance. ...

> So we neither want to be too onerous in requirements, nor too strict (after all, we can adjust later on).

+1

markusicu:
Discussed in PAG meeting 20240222: Let's finish this one and then follow with a pull request for U+0300 and U+0308 and let's review the details of the test failures. Probably we can just modify either the sample strings or the expected results.

eggrobin (approved):

markusicu:
I find it hard to figure out what's changing for which character, for those that already had scx, because we group them by scx value which is a set of sc values -- so they move far in the file when the first listed script changes.

I suspect that this file would be easier to read and maintain if we gave up on grouping by value, and instead simply listed by code points and ranges.

@macchiati @roozbehp @eggrobin

eggrobin:
> I suspect that this file would be easier to read and maintain if we gave up on grouping by value, and instead simply listed by code points and ranges.

I tend to agree. I will poke around the options in MakeUnicodeFiles to see what can easily be done, and then we can discuss that in PAG once we have something concrete.

macchiati:
I think that's reasonable. Sometimes it is convenient to see the groupings, but that can be done elsewhere.

PR #616, opened 2023-11-29
un-break UCA tool: Adlam new FracUCA byte after Garay

markusicu:
In the CLDR/ICU FractionalUCA.txt, the addition of the new script Garay between Medefaidrin and Adlam pushed Adlam across a primary-weight lead byte boundary, which does not work for a primary-compressible script. I needed to split the sequence of scripts from Vai to Adlam across two lead bytes and decided to simply start a lead byte with Adlam.

Some of these scripts are not used widely but use a fair bit of primary weight space because they are cased, and thus stored with two-byte primary weights, so that their collation elements easily fit into 32 bits, instead of with three-byte primaries.

The CollationTest file diffs are large and hard to look at. I don't expect much of a review there.


markusicu:
... and the “Check UCA data” workflow passes for the first time in a while!
Let's keep it green now for Unicode 16!

eggrobin (approved):

macchiati (approved):
Looks good. I usually look over the generated Rules because I can make sense of those but not the plain numbers, but I guess those were unchanged for this PR.

markusicu:
> I usually look over the generated Rules because I can make sense of those but not the plain numbers, but I guess those were unchanged for this PR.

I was looking at the generated FractionalUCA.txt while debugging and fixing, but that file is not checked into this repo, and it's too early to take Unicode 16 data into CLDR.

I am thinking that at some point I should modify the tool to directly generate a nearly complete CLDR FractionalUCA.txt file but with the weights “[blanked](https://github.com/unicode-org/cldr/blob/main/tools/scripts/uca/blankweights.sed)”, and to check that into this repo, so that we see more directly what changes.

macchiati:
good idea

PR #618, opened 2023-11-29
Fix remaining issues with multivalued

macchiati:
There were two further problems with multivalued support

1. \p{exemplar} (has no exemplar value) returned the reverse of what was expected
2. \p{scx=Arab} didn't work right.

The solution to these was to (1) allow for "" properly as a string value, and (2) allow script values that were not present in ScriptExtensions.txt as single values.

UnicodeJsps/src/main/java/org/unicode/jsp/ScriptTester.java
- Supply the Script values in addition to the Script_Extension values as valid alternate aliases.

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
- Expand an empty value to be "No" only for binary, enumerated, and catalogue properties

UnicodeJsps/src/main/java/org/unicode/jsp/XPropertyFactory.java
- main change is to allow "" as the default value in the scx map

UnicodeJsps/src/test/java/org/unicode/jsptest/TestMultivalued.java
- Add some tests for the problems seen above

unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
- add a mask for  BINARY_OR_ENUMERATED_OR_CATALOG_MASK
- if a property is multivalued, don't skip over string values.

Notes:

1. As before, Eclipse helpfully adds missing @Override annotations
2. The exemplars really should be treated like enumerations (over locales), but that was a bit too complicated for a first run.

### DIFF
--- a/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
+++ b/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
@@ -150,6 +150,7 @@ public class UnicodeSetUtilities {
         //      return null;
         //    }
 
+        @Override
         public boolean applyPropertyAlias(
                 String propertyName, String propertyValue, UnicodeSet result) {
             boolean status = false;
@@ -201,9 +202,11 @@ public class UnicodeSetUtilities {
                     }
                     ;
                     if (!status) {
-                        try {
-                            status = applyPropertyAlias0(prop, "No", result, !invert);
-                        } catch (Exception e) {
+                        if (prop.isType(UnicodeProperty.BINARY_OR_ENUMERATED_OR_CATALOG_MASK)) {
+                            try {
+                                status = applyPropertyAlias0(prop, "No", result, !invert);
+                            } catch (Exception e) {
+                            }
                         }
                         ;
                         if (!status) {
@@ -336,6 +339,7 @@ public class UnicodeSetUtilities {
             this.pattern = pattern;
         }
 
+        @Override
         public boolean test(String value) {
             int comp = comparator.compare(pattern, value.toString());
             switch (relation) {
@@ -352,6 +356,7 @@ public class UnicodeSetUtilities {
             }
         }
 
+        @Override
         public PatternMatcher set(String pattern) {
             this.pattern = pattern;
             return this;
--- a/UnicodeJsps/src/main/java/org/unicode/jsp/XPropertyFactory.java
+++ b/UnicodeJsps/src/main/java/org/unicode/jsp/XPropertyFactory.java
@@ -269,15 +269,19 @@ public class XPropertyFactory extends UnicodeProperty.Factory {
 
         // set up the special script property
         UnicodeProperty scriptProp = base.getProperty("sc");
+
+        // Compose the function and add
         UnicodeMap<String> specialMap = new UnicodeMap<String>();
-        specialMap.putAll(scriptProp.getUnicodeMap());
+        specialMap.putAll(
+                scriptProp.getUnicodeMap()); // if there is no value, use the script property
         specialMap.putAll(ScriptTester.getScriptSpecialsNames());
         add(
                 new UnicodeProperty.UnicodeMapProperty()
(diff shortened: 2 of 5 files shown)


UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
```diff
@@ -201,9 +202,11 @@ public boolean applyPropertyAlias(
                     }
                     ;
                     if (!status) {
-                        try {
-                            status = applyPropertyAlias0(prop, "No", result, !invert);
-                        } catch (Exception e) {
+                        if (prop.isType(UnicodeProperty.BINARY_OR_ENUMERATED_OR_CATALOG_MASK)) {
+                            try {
+                                status = applyPropertyAlias0(prop, "No", result, !invert);
```
markusicu, line 207: I understand that binary properties default to "No", but each enumerated/catalog property has its own `@missing` value, so why should those default to "No" rather than their actual `@missing` value?

UnicodeJsps/src/main/java/org/unicode/jsp/XPropertyFactory.java
```diff
@@ -269,15 +269,19 @@ public String transform(Integer source) {
...
+        specialMap.putAll(
+                scriptProp.getUnicodeMap()); // if there is no value, use the script property
         specialMap.putAll(ScriptTester.getScriptSpecialsNames());
         add(
                 new UnicodeProperty.UnicodeMapProperty()
                         .set(specialMap)
                         .setMain("Script_Extensions", "scx", UnicodeProperty.ENUMERATED, "1.1")
                         .addValueAliases(
-                                ScriptTester.getScriptSpecialsAlternates(),
+                                // valueArray,
```
markusicu, line 283: intentional addition?

UnicodeJsps/src/test/java/org/unicode/jsptest/TestMultivalued.java
```diff
@@ -3,11 +3,19 @@
...
 
 public class TestMultivalued extends TestFmwkMinusMinus {
+
+    UnicodeProperty exemplarProp = XPropertyFactory.make().getProperty("exemplar");
+    UnicodeProperty scxProp = XPropertyFactory.make().getProperty("scx");
+
     @Test
     public void TestScx1Script() {
+        String x = scxProp.getValue('।');
+
```
markusicu, line 18: unused, please remove

UnicodeJsps/src/main/java/org/unicode/jsp/XPropertyFactory.java
macchiati: fixed

UnicodeJsps/src/test/java/org/unicode/jsptest/TestMultivalued.java
```diff
@@ -48,4 +72,40 @@ public void TestExemplars() {
...
+        assertEquals(unicodeSetString2, "el", firstValue2);
+
+        //        UnicodeSet mustContain = new UnicodeSet("[æ]");
+        //        assertTrue(unicodeSetString + " contains " + mustContain,
+        // parsed.containsAll(mustContain));
+        //
+        //        UnicodeSet mustNotContain = new UnicodeSet("[ç]");
+        //        assertFalse(
+        //                unicodeSetString + " !contains " + mustNotContain,
+        //                parsed.containsAll(mustNotContain));
```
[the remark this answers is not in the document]
macchiati: will remove

UnicodeJsps/src/test/java/org/unicode/jsptest/TestMultivalued.java
```diff
@@ -37,6 +59,8 @@ public void TestScxMulti() {
 
     @Test
     public void TestExemplars() {
+        String x = exemplarProp.getValue('æ');
+
```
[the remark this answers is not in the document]
macchiati: fixed

markusicu (approved):

PR #626, opened 2023-12-14
UTC-176-C32 Adding lowercase double thorn/wynn

macchiati:
Adding 2 characters (listed in the doc as the following but only the capitals are new)
```
A7D2;LATIN CAPITAL LETTER DOUBLE THORN;Ll;0;L;;;;;N;;;;A7D3;
A7D3;LATIN SMALL LETTER DOUBLE THORN;Ll;0;L;;;;;N;;;A7D2;;A7D2
A7D4;LATIN CAPITAL LETTER DOUBLE WYNN;Ll;0;L;;;;;N;;;;A7D5;
A7D5;LATIN SMALL LETTER DOUBLE WYNN;Ll;0;L;;;;;N;;;A7D4;;A7D4
```

PR #628, opened 2023-12-15
Revert "UTC-176-C32 Adding lowercase double thorn/wynn"

markusicu:
Reverts unicode-org/unicodetools#626 because these characters are only provisionally assigned. They are not approved for Unicode 16.0, so we cannot have their data on the unicodetools main branch. See https://www.unicode.org/alloc/Pipeline.html and RMG https://github.com/unicode-org/utc-release-management/issues/54


echeran (approved):
rslgtm

Issue #627, opened 2023-12-14
Modify MakeUnicodeFiles


macchiati:
Currently all the scripts require clearing out various directories, and copying certain files at the end. This could be built into MakeUnicodeFiles for ease of use.

https://github.com/unicode-org/unicodetools/blob/main/docs/pipeline.md#scripts

PR #636, opened 2023-12-20
Add MakeUnicodeFiles.java option

macchiati:
https://github.com/unicode-org/unicodetools/issues/627

Adds option to clean directory before building, then copy relevant files after building, as per issue

### DIFF
--- a/unicodetools/src/main/java/org/unicode/text/UCD/MakeUnicodeFiles.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/MakeUnicodeFiles.java
@@ -60,23 +60,22 @@ public class MakeUnicodeFiles {
 
     public static void main(String[] args) throws IOException {
 
-        boolean clean = Arrays.asList(args).contains("-c");
+        boolean cleanAndCopy =
+                Arrays.asList(args).contains("-c"); // clean Bin & copy changed output
 
-        if (clean) {
+        if (cleanAndCopy) {
 
             // Remove the bin file so that changes to the dev directory aren't ignored
             // The index unicode properties (eg Alphabetic.bin) don't need to be removed
             // because they are rebuilt if the source files change
 
             File binFile =
-                    new File(
-                            Settings.Output.GEN_DIR + "BIN",
-                            "UCD_Data" + Settings.latestVersion + ".bin");
+                    new File(Settings.Output.BIN_DIR, "UCD_Data" + Settings.latestVersion + ".bin");
             binFile.delete();
 
-            // remove the old files in the output directory
+            // Remove the old files in the output directory
 
-            Files.walk(Path.of(Settings.Output.GEN_DIR + "UCD"))
+            Files.walk(Path.of(Settings.Output.GEN_UCD_DIR))
                     .sorted(Comparator.reverseOrder())
                     .map(Path::toFile)
                     .forEach(File::delete);
@@ -84,7 +83,7 @@ public class MakeUnicodeFiles {
 
         generateFile();
 
-        if (clean) {
+        if (cleanAndCopy) {
 
             // Copy the important changed files to dev directory
 
@@ -108,7 +107,12 @@ public class MakeUnicodeFiles {
                     .forEach(
                             x -> {
                                 Path targetDir = targetBase.resolve(sourceBase.relativize(x));
-                                System.out.println(++fileCount.value + ") Moving:\t" + x + "\tto\t" + targetDir);
+                                System.out.println(
+                                        ++fileCount.value
+                                                + ") Moving:\t"
+                                                + x
+                                                + "\tto\t"
+                                                + targetDir);
                                 try {
                                     Files.move(x, targetDir, StandardCopyOption.REPLACE_EXISTING);
                                 } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


unicodetools/src/main/java/org/unicode/text/UCD/MakeUnicodeFiles.java
```diff
@@ -53,7 +59,66 @@ public class MakeUnicodeFiles {
...
         generateFile();
+
+        if (clean) {
+
+            // Copy the important changed files to dev directory
+
+            Path sourceBase = Path.of(Settings.Output.GEN_DIR + "UCD/" + Settings.latestVersion);
+            Path targetBase =
+                    Settings.UnicodeTools.DataDir.UCD.asPath(Settings.LATEST_VERSION_INFO);
+            // Path.of(Settings.SRC_UCD_DIR + Settings.NEXT_VERSION_FOLDER_NAME);
```
macchiati, line 94: Hmmm. This is left over, should remove.

unicodetools/src/main/java/org/unicode/text/UCD/MakeUnicodeFiles.java
```diff
@@ -53,7 +59,64 @@ public class MakeUnicodeFiles {
...
+
+        if (clean) {
+
+            // Remove the bin file so that changes to the dev directory aren't ignored
+            // The index unicode properties (eg Alphabetic.bin) don't need to be removed
+            // because they are rebuilt if the source files change
+
+            File binFile =
+                    new File(
+                            Settings.Output.GEN_DIR + "BIN",
```
markusicu, line 73:
```suggestion
                            Settings.Output.BIN_DIR,
```

unicodetools/src/main/java/org/unicode/text/UCD/MakeUnicodeFiles.java
```diff
@@ -53,7 +59,64 @@ public class MakeUnicodeFiles {
...
+                            "UCD_Data" + Settings.latestVersion + ".bin");
+            binFile.delete();
+
+            // remove the old files in the output directory
+
+            Files.walk(Path.of(Settings.Output.GEN_DIR + "UCD"))
```
markusicu, line 79:
```suggestion
            Files.walk(Path.of(Settings.Output.GEN_UCD_DIR))
```

unicodetools/src/main/java/org/unicode/text/UCD/MakeUnicodeFiles.java
```diff
@@ -53,7 +59,64 @@ public class MakeUnicodeFiles {
...
+                    .forEach(File::delete);
+        }
+
         generateFile();
+
+        if (clean) {
+
+            // Copy the important changed files to dev directory
+
+            Path sourceBase = Path.of(Settings.Output.GEN_DIR + "UCD/" + Settings.latestVersion);
```
markusicu, line 91:
```suggestion
            Path sourceBase = Path.of(Settings.Output.GEN_UCD_DIR + Settings.latestVersion);
```

unicodetools/src/main/java/org/unicode/text/UCD/MakeUnicodeFiles.java
```diff
@@ -53,7 +59,64 @@ public class MakeUnicodeFiles {
     static boolean DEBUG = false;
 
     public static void main(String[] args) throws IOException {
+
+        boolean clean = Arrays.asList(args).contains("-c");
```
markusicu, line 63: How about a more self-explanatory option string, something like "--cleanBin" ?

unicodetools/src/main/java/org/unicode/text/UCD/MakeUnicodeFiles.java
```diff
@@ -53,7 +59,64 @@ public class MakeUnicodeFiles {
...
+
+            Files.walk(Path.of(Settings.Output.GEN_DIR + "UCD"))
+                    .sorted(Comparator.reverseOrder())
+                    .map(Path::toFile)
+                    .forEach(File::delete);
+        }
+
         generateFile();
+
+        if (clean) {
```
markusicu, line 87: I find it surprising/confusing that the "clean" option should also drive copying the files back to the input.

It would be better to make another new option like "--changesToInput" .

markusicu (commented):
nice functionality!

unicodetools/src/main/java/org/unicode/text/UCD/MakeUnicodeFiles.java
markusicu: ping -- longer option names make reading and documenting command lines much easier

unicodetools/src/main/java/org/unicode/text/UCD/MakeUnicodeFiles.java
markusicu: ping -- I really think it would be better to make a second option, with another readable option name (not just renaming the Java variable)

josh-hadley (approved):

unicodetools/src/main/java/org/unicode/text/UCD/MakeUnicodeFiles.java
josh-hadley: I don't disagree with @markusicu but from the standpoint of practicality, do we see ourselves or anyone else using one option without the other?

macchiati:
Markus, I'm going to go ahead and merged. It it comes to the point where someone needs 2 options (or more options need to be added) I can do additional work then.

PR #645, opened 2024-01-10
Partially test consistency of grapheme cluster segmentation with canonical equivalence, and fix it for LGCs

eggrobin:
Towards #522.

This would have caught the issue reported by Charlotte Buff in https://www.unicode.org/L2/L2023/23159-pubrev.html#ID20230502072316, as well as unicode-org/properties#229.

In fact, it caught the corresponding issue for Tulu-Tigalari (which only affects _legacy_ grapheme clusters). See the failures in https://github.com/unicode-org/unicodetools/pull/645/commits/85eca4e4760cb9b369b992fc2cbdee10aa719a8d.

PR #646, opened 2024-01-10
Show invariant test results in the diff

### LINKED ISSUE #612: Show invariant test failures on UnicodeInvariantTest.txt
Having to run them locally to figure out what went wrong gets annoying (see https://github.com/unicode-org/unicodetools/pull/607#issuecomment-1808998728 but I have also been repeatedly annoyed) and takes time.

This can be done by making the GitHub workflow emit a `::error` line for each failure.

eggrobin:
See it in action [here](https://github.com/unicode-org/unicodetools/pull/646/commits/03b74f241ad1b7978773ca4d357c6a09c48ccd39#diff-c2f2f053bbaa14f43053e3fb81a8afd67add46a4be1377d3e411f0500b40dabf), I obviously don’t want to check in invariant test failures.

Split out from #645 since we will probably not merge that one before UTC-178. Fix #612.


unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -435,16 +437,26 @@ private static void equivalencesLine(String line, ParsePosition pp) throws Parse
             ++testFailureCount;
             printErrorLine("Test Failure", Side.START, testFailureCount);
         }
+        final List<String> errorMessage = new ArrayList<>();
```
markusicu, line 440: Nit/optional: It seems like a container should have a plural name. Maybe errorMessageLines?

markusicu (approved):
looks plausible

markusicu (approved):
tnx

Issue #502, opened 2023-06-15
Use the UCD properties in the JSPs


eggrobin:
Instead of those from ICU.

Repeatedly discussed, blocks #432.

From Mark in PAG:
> Minor: [:Script=Ethi:]&[:L:]-[ሀ-ሆለ-ቆቈቊ-ቍበ-ኆኈኊ-ኍነ-ኮኰኲ-ኵኸ-ኾወ-ዎዐ-ዖዘ-ዮደ-ዷጀ-ጎጐጒ-ጕጠ-ፆፈ-ፗ]-[:Changes_When_NFKC_Casefolded:]-[:Identifier_Status=Restricted:]
Used to work and doesn’t now. The last set needs to be [:Identifier_Statusβ=Restricted:]
Identifier_Status is one of the few UCD properties that ICU doesn’t support, so it needs to use the UCD properties. However, we’re talking about just converting to always use UCD props, so we shouldn’t spend time on it.

eggrobin:
Interesting discovery while working on this: The pile of .bin files in https://github.com/unicode-org/unicodetools/tree/main/UnicodeJsps/src/main/resources/org/unicode/jsp/props is *not* at Unicode 15.1, it has stuff from 16.0 (such as Myanmar Extended-C), because that was merged already when the .bin files were built.

But that is OK, because we don’t actually use the cached props at all, the ICU props get there first.

For the sake of my own sanity I will proceed on this issue (and #432) ignoring this attempt at caching entirely. We’ll see if we need to do caching (hopefully correct caching that we actually use) once we get there.

eggrobin:
Well, we have caching in IndexUnicodeProperties already, so that covers that.

eggrobin:
> But that is OK, because we don’t actually use the cached props at all, the ICU props get there first.

Actually we do use it in some random places, so for instance it leaks here. What a mess.

https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%5Cx%7B116D0%7D%5D&g=&i=

PR #647, opened 2024-01-13
Move VersionedProperty to its own file

eggrobin:
Towards #432 and #502.

I want to do #432 (versioned property support in the JSPs) together with #502 (using unicodetools properties rather than ICU properties and properties badly reverse-engineered from ICU in the JSPs), otherwise we would lose the capability to have β properties.

That means splitting out the VersionedProperty class which is currently in the invariants test to someplace more reasonable for use in the JSPs.

I am trying to make this a « no functional change » PR, so while I am touching weird usages in UTS39-related generators that were reaching into the tests, I am not going to try to improve those.
The changes to VersionedProperty are minimal to get things to compile. They are quite hard to review because of the move, so I want to make any meaningful changes (and I will need to make changes for #432) in a separate PR; again I am trying very hard not to improve anything here. See https://goomics.net/119/.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/text/UCD/VersionedProperty.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/VersionedProperty.java
@@ -1,0 +1,141 @@
+package org.unicode.text.UCD;
+
+import com.ibm.icu.dev.util.UnicodeMap;
+import com.ibm.icu.text.SymbolTable;
+import com.ibm.icu.text.UnicodeSet;
+import java.text.ParsePosition;
+import java.util.Map;
+import java.util.Set;
+import org.unicode.props.IndexUnicodeProperties;
+import org.unicode.props.UnicodeProperty;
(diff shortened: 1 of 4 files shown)


unicodetools/src/main/java/org/unicode/text/UCD/VersionedProperty.java
```diff
@@ -0,0 +1,141 @@
...
+import org.unicode.text.utility.Settings;
+
+public class VersionedProperty {
+    private String propertyName;
+    private String version;
+    private UnicodeProperty.Factory propSource;
+    private UnicodeProperty property;
+    private final transient PatternMatcher matcher = new UnicodeProperty.RegexMatcher();
+
+    private static final Set<String> TOOL_ONLY_PROPERTIES =
```
macchiati, line 22: BTW, just checked in changes to a PR so that the normalization code can use the IndexUnicodeProperties.

macchiati (approved):

PR #648, opened 2024-01-13
Fix the handling of multivalued properties in IndexUnicodeProperties and UnicodeProperty

eggrobin:
DRAFT: failing test only


macchiati (dismissed):
Looks good. Lots of tech debt there. (Making some progress on removing the old UCD.java. Once that is done I think we can simplify/update lots of code.

macchiati (approved):
Good fixes, thanks.

PR #651, opened 2024-01-16
Emoji 16.0 alpha data

nedley:
Rather than going through the traditional provisional -> draft -> final candidate stages, especially since the new characters have already been added to UCD, I’m speedrunning directly to final candidate for 16.0. First set the VM argument `-Demoji-beta=true` as usual to grab the keywords and proposals for the candidates. For Sark, add the sequence alone to emoji-sequences.txt (the other fields are populated automatically) and the proposal to proposalData.txt. For the rest, add to candidateData.txt and emojiOrdering.txt as usual for final. Finally, run GenerateEmoji.java a couple of times until the age of the new characters is correct. I may have forgotten an intermediate stage due to debugging, but I believe that’s it.


unicodetools/src/main/java/org/unicode/tools/emoji/EmojiAnnotations.java
```diff
@@ -97,18 +97,7 @@ public EmojiAnnotations(
             } else if (outShortName.value.contains(Annotations.BAD_MARKER)) {
                 status = Status.gender;
             }
-            // workaround: 15.1 sequences have commas instead of pipes
```
nedley, line 100: No longer needed since CLDR has the fixed data.

markusicu (approved):
looks plausible, but I don't understand every bit; maybe wait a little for Mark to review as well?

unicodetools/src/main/resources/org/unicode/tools/emoji/proposalData.txt
```diff
@@ -1870,6 +1870,17 @@
...
+1FA89; L2/23-256; harp
+1FA8F; L2/23-259; shovel
+1FABE; L2/23-255; leafless tree
+1FAC6; L2/23-258; fingerprint
+1FADC; L2/23-257; root vegetable
+1FADF; L2/23-261; splatter
+1FAE9; L2/23-260; face with bags under eyes
+
+# https://www.unicode.org/L2/L2023/23157.htm#176-A6
+1F1E8 1F1F6; MISSING; Sark
```
nedley, line 1882: Jennifer would rather I not link to the ESC report as I was doing before, so until I can add action items to the DocRegistry class this will have to do.

unicodetools/src/main/java/org/unicode/tools/emoji/ProposalData.java
```diff
@@ -122,7 +122,7 @@ public Set<String> getProposals(String source) {
 
     public String formatProposalsForHtml(Collection<String> proposals) {
         if (proposals == null || proposals.isEmpty()) {
-            return "MISSING";
+            return "&mdash;";
```
nedley, line 125: As I believe Tolstoy wrote, every missing proposal is missing in its own way.

unicodetools/src/main/resources/org/unicode/tools/emoji/proposalData.txt
macchiati: Rather than 'MISSING', you could put AUTO or something to indicate that it is required by policy

macchiati (approved):

PR #657, opened 2024-01-18
Fix the IndexUnicodeProperties Joining_Type

eggrobin:
Treat column 2 of ArabicShaping.txt as an unofficial contributory property and Joining_Type as a derived property in IndexUnicodePropertySource. When retiring ToolUnicodePropertySource, we should switch to using that derivation.

See [the test failure prior to the fix](https://github.com/unicode-org/unicodetools/pull/657/commits/616a9484cd1ad45308dce5e728eff44b5f928df6#diff-c2f2f053bbaa14f43053e3fb81a8afd67add46a4be1377d3e411f0500b40dabf).

### DIFF
--- a/unicodetools/src/main/java/org/unicode/props/UcdProperty.java
+++ b/unicodetools/src/main/java/org/unicode/props/UcdProperty.java
@@ -29,6 +29,7 @@ import org.unicode.props.UcdPropertyValues.NFD_Quick_Check_Values;
 import org.unicode.props.UcdPropertyValues.NFKC_Quick_Check_Values;
 import org.unicode.props.UcdPropertyValues.NFKD_Quick_Check_Values;
 import org.unicode.props.UcdPropertyValues.Numeric_Type_Values;
+import org.unicode.props.UcdPropertyValues.Other_Joining_Type_Values;
 import org.unicode.props.UcdPropertyValues.Script_Values;
 import org.unicode.props.UcdPropertyValues.Sentence_Break_Values;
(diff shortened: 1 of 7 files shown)


macchiati (changes requested):
clever approach. A few items:
The UcdProperty and UcdProperty values are generated by GenerateEnums, so that needs to be modified to produce the right values. Otherwise the next time it is run, the changes will be overridden.

Short term, I think making the derived property "occupy" field 1 is a clever hack. Longer term I think we need to add a better mechanism for derived IndexUnicodeProperties.

eggrobin:
> The UcdProperty and UcdProperty values are generated by GenerateEnums, so that needs to be modified to produce the right values. Otherwise the next time it is run, the changes will be overridden.

I ran GenerateEnums already (a CI check tests that it does not change anything).

> I think we need to add a better mechanism for derived IndexUnicodeProperties.

We have plenty of those in IndexUnicodeProperties.txt already, see e.g. https://github.com/unicode-org/unicodetools/blob/77e3ce601ff2970b1a9b8bc7cf5d9445fd4554e5/unicodetools/src/main/resources/org/unicode/props/IndexUnicodeProperties.txt#L124-L143

eggrobin:
> making the derived property "occupy" field 1 is a clever hack

You may have misread that line of IndexUnicodeProperties.txt, it is

**DerivedJoiningType**; Joining_Type; 1

that is, we are getting it from the derived file (like the other derived properties), where it actually does reside in field 1.

macchiati (approved):

markusicu (approved):

PR #660, opened 2024-01-19
Another script from the future

eggrobin:
The same procedure as #443.

See L2/24-004 and L2/24-013.


unicodetools/src/main/java/org/unicode/text/UCD/UCD_Types.java
```diff
@@ -607,12 +607,13 @@ public interface UCD_Types {
             Sunuwar = 170,
             Todhri = 171,
             Tulu_Tigalari = 172,
-            // Provisionally assigned
+            // Provisionally assigned and proposed
```
markusicu, line 610: nit/optional
```suggestion
            // Provisionally assigned or proposed
```

markusicu (dismissed):

markusicu (approved):

PR #667, opened 2024-01-20
There can only be one MySymbolTable

eggrobin:
Inspired by #66, and by the fact that I wasted time while working on #659 editing a class MySymbolTable that is only used in a broken test.


unicodetools/src/test/java/org/unicode/propstest/TestXUnicodeSet.java
eggrobin, line 1: Actually there are no tests left in this file, so I should delete it.

srl295 (approved):

PR #668, opened 2024-01-21
Warn about DerivedAge.txt for characters not yet in the pipeline

eggrobin:
See the discussion in https://github.com/unicode-org/sah/issues/371#issuecomment-1902678425.

See the warning [here](https://github.com/unicode-org/unicodetools/pull/668/commits/635ce777800c95fb158ffd7f30376bff3b65b0bb#diff-11c75da046f2286ea1987e46a0bd47d3116bdbac8787ae2b80d9ea130651a751).

### DIFF
--- a/py/pipeline-workflow/check-draft-status.py
+++ b/py/pipeline-workflow/check-draft-status.py
@@ -7,11 +7,33 @@ pipeline_label = os.environ['PIPELINE_LABEL']
 with open("unicodetools/data/ucd/dev/DerivedAge.txt", 'r') as f:
     version = f.readline().strip().replace("# DerivedAge-", "").replace(".0.txt", "")
 
+PROVISIONALLY_ASSIGNED_AGE_NOTICE = f"""\f
+While the Unicode Technical Committee has provisionally assigned these
+characters, they have not been accepted for Unicode {version}, nor for any
+specific version of Unicode.
+
+The values of the Age property in this file are a placeholder.
+""".replace("\n", "%0A")
+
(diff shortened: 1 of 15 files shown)


py/pipeline-workflow/check-draft-status.py
```diff
@@ -7,11 +7,35 @@
...
+specific version of Unicode.
+
+The values of the Age property in this file are a placeholder.
+""".replace("\n", "%0A")
+
+UNREVIEWED_AGE_NOTICE = f"""\f
+These characters are neither accepted for Unicode {version}, nor for any
+specific version of Unicode, nor are they provisionally assigned.
+
+The values of the Age property in this file are a placeholder.
```
markusicu, line 22: Singular/plural mismatch.

And maybe better:
The Age property values for new characters are likely incorrect right now. They will be recomputed after the UTC accepts their encoding and this pull request is updated for the target version.

markusicu (commented):
pro: makes it clear that the Age is a lie until the new characters are accepted for a version and we open the gates for that version

con: clutter for PAG members who know that...

pro: still helps PAG members who are new to reviewing these PRs or rarely do it

eggrobin:
> pro: makes it clear that the Age is a lie until the new characters are accepted for a version and we open the gates for that version
>
> con: clutter for PAG members who know that...

Yes, I had thought about that…

> pro: still helps PAG members who are new to reviewing these PRs or rarely do it

But I think this is the decisive point. Even among the cognoscenti this can get a bit confusing.

markusicu (approved):
looks good to Libby

Issue #656, opened 2024-01-18
Test decomposition that contain more than one starter and ends with a non-starter followed by low-CCC non-starter


hsivonen:
ICU4X had a bug in the normalizer for the case where a character whose decomposition contains more than one starter and ends with a non-starter is followed by a non-starter whose CCC is lower than the final non-starter of the decomposition.

I had thought about this case, but I had failed to type a `!` for logical negation of a condition and the test suite didn't catch this.

NormalizationTest.txt should include cases like this. Known examples:
* NFKD of `01C4 0323`
* NFD of `0DDD 0334`

hsivonen:
CC @eggrobin

PR #669, opened 2024-01-22
Ensure we have comprehensive coverage for the ICU4X bug

eggrobin:
See #658, #656.
@hsivonen pointed out to me that they have a special code path for the non-BMP case that is currently unused. This ensure that, should we ever encode a decomposition to which it applies, the tests would exercise that.


unicodetools/src/main/java/org/unicode/text/UCD/GenerateData.java
```diff
@@ -811,6 +814,57 @@ public static void writeNormalizerTestSuite(String directory, String fileName)
...
+        // cases, because there are not many of them, and the edge case feels a
+        // tad too weird to describe in the title of a new part.
+        final org.unicode.props.UnicodeProperty sc =
+                IndexUnicodeProperties.make().getProperty(UcdProperty.Script);
+        final org.unicode.props.UnicodeProperty ccc =
+                IndexUnicodeProperties.make().getProperty(UcdProperty.Canonical_Combining_Class);
+        loopOverCodePoints:
+        for (final String cp : UnicodeSet.ALL_CODE_POINTS) {
+            final String[] decompositions =
+                    new String[] {Default.nfd().normalize(cp), Default.nfkd().normalize(cp)};
```
markusicu, line 836: Nit: Should we cache things like Default.nfd(), Default.nfkd(), Default.ucd(), IUP.make(), ... in variables in this function, or in instance fields? That should be easy, might save a bit of time on class synchronization, and make these call sites shorter (more readable). WDYT?

---
I *do not* like to cache this kind of stuff in *static* class fields, because if something goes wrong in the static initializer it's cumbersome to debug.

unicodetools/src/main/java/org/unicode/text/UCD/GenerateData.java
```diff
@@ -811,6 +814,57 @@ public static void writeNormalizerTestSuite(String directory, String fileName)
...
+            for (final String decomposition : decompositions) {
+                final int lastCCC =
+                        Default.ucd()
+                                .getCombiningClass(
```
markusicu, line 840: Nit: Why don't you use ccc instead of Default.ucd().getCombiningClass()? It seems like you would want to use the object for the property if you already got that.

unicodetools/src/main/java/org/unicode/text/UCD/GenerateData.java
```diff
@@ -811,6 +814,57 @@ public static void writeNormalizerTestSuite(String directory, String fileName)
...
+                        // ARABIC SUBSCRIPT ALEF.
+                        writeLine(cp + "\u0656", log, false);
+                    } else if (lastCCC > 220) {
+                        // COMBINING DOT BELOW.
+                        writeLine(cp + "\u0323", log, false);
+                    } else {
+                        // COMBINING TILDE OVERLAY.
+                        writeLine(cp + "\u0334", log, false);
+                    }
+                    continue loopOverCodePoints;
```
markusicu, line 864: Nit: A simple `break` would suffice here, and eliminate the outer-loop label.

unicodetools/src/main/java/org/unicode/text/UCD/GenerateData.java
eggrobin: > WDYT?

Yes, but in a subsequent cleanup PR, because a cursory search for .nfd() finds 18 results of which 7 are in commented-out code, and with the file in that state I do not want to waste further time figuring out which ones are in the same function; nor do I want to burn the commented-out code to the ground in this unrelated change.

unicodetools/src/main/java/org/unicode/text/UCD/GenerateData.java
eggrobin: I got that, but I am not actually using it, and while `IndexUnicodeProperties` is the more modern thing, it is inconvenient here, as I want an integer, not a string, for the comparisons below. Removed the declaration of `ccc`.
(Someday we bring stronger typing to `IndexUnicodeProperties`.)

unicodetools/src/main/java/org/unicode/text/UCD/GenerateData.java
eggrobin: Done.

markusicu (approved):

PR #673, opened 2024-01-30
Invariants for Equivalent_Unified_Ideograph

eggrobin:
UTC-151-A143 **Action Item for** <del>Mark Davis, Laurențiu Iancu: </del><ins>Robin Leroy, Josh Hadley, PAG: </ins>Add invariant tests for new property and property values from PRI-344.

### DIFF
--- a/unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
+++ b/unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
@@ -945,6 +945,15 @@ Let $ideohack = [〆 〇 〡-〩]
 # TODO(macchiati): The kHanyuPinyin UnicodeSet is excruciatingly slow.
 # \p{kHanyuPinyin=/./} - \p{kMandarin=/./} = [\{228F5}\x{2574C}]
 
+Let $strokesAndRadicals = [\p{Name=/^CJK STROKE /}\p{Name=/^(CJK|KANGXI) RADICAL /}]-\p{gc=Cn}
+Let $nonIdeographicRadicals = \N{CJK RADICAL REPEAT}
+# The following set may expand over time, if strokes are added.
+# It can also shrink, if single-stroke ideographs are encoded.
+Let $nonIdeographicStrokes = \p{Name=/^CJK STROKE (T|WG|XG|BXG|SW|HZZ|HP|HZWG|SZWG|HZT|HZZP|HPWG|HZW|HZZZ|PG|Q|HZXG|SZP)$/}
+
+\p{Equivalent_Unified_Ideograph=/./} ⊆ $strokesAndRadicals
+[$strokesAndRadicals - \p{Equivalent_Unified_Ideograph=/./}] = [$nonIdeographicStrokes $nonIdeographicRadicals]
+
 # InPC-InSC-gc invariants
 # See https://www.unicode.org/L2/L2023/23200-category-invariants.pdf.
 \p{InPC=/(Left|Right)/} ⊆ [\p{gc=Mc}\p{gc=Lo}\p{gc=Lm}]


unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
```diff
@@ -945,6 +945,18 @@ Let $ideohack = [〆 〇 〡-〩]
...
+Let $nonIdeographicRadicals = \N{CJK RADICAL REPEAT}
+# The following set may expand over time, if strokes are added.
+# It can also shrink, if single-stroke ideographs are encoded.
+Let $nonIdeographicStrokes = \p{Name=/^CJK STROKE (T|WG|XG|BXG|SW|HZZ|HP|HZWG|SZWG|HZT|HZZP|HPWG|HZW|HZZZ|PG|Q|HZXG|SZP)$/}
+
+\p{Equivalent_Unified_Ideograph=/./} ⊆ $strokesAndRadicals
+[$strokesAndRadicals - \p{Equivalent_Unified_Ideograph=/./}] = [$nonIdeographicStrokes $nonIdeographicRadicals]
+# TODO(egg): NFC_Quick_Check is a stupid way to get a Yes here; we are checking
+# that Equivalent_Unified_Ideograph values are single unified ideographs.
+In \p{Equivalent_Unified_Ideograph=/./} Unified_Ideograph * Equivalent_Unified_Ideograph = NFC_Quick_Check
```
markusicu, line 958: For people who are looking at this PR but don't know chapter & verse of the syntax here, can you please provide a brief translation into English for what this line does?

eggrobin: The syntax is documented at the top of this file:
https://github.com/unicode-org/unicodetools/blob/709e4c4357d1363cb0f6d6e7f6002da645e56e3f/unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt#L52-L75

markusicu: This is pretty obscure stuff...
- One would think that applying a binary property is like a `<unicodeSet>`, that is, applying a filter.
- However, apparently, in order to make this work, a binary property (without explicit `\p{...}` syntax) works like an enumerated one, mapping each character to "Yes" or "No".
- And enumerated properties map to strings, not enum-type values.

So this says
- For each character that has an Equivalent_Unified_Ideograph mapping,
  - (left) apply the mapping, then map each Unified_Ideograph to "Yes" and everything else to "No";
  - (right) apply NFC_Quick_Check which maps the input character to "Yes" or "Maybe" or "No"
  - left & right should be the same

> we are checking that Equivalent_Unified_Ideograph values are single unified ideographs

Are you? It looks like the "In" syntax and its regex *limits the test* to those characters that map to single ones, so it does not seem to perform that part of the stated intention.

I guess that you are just using NFC_Quick_Check because there is no syntax to just say “I expect the result to be "Yes"”? So you could just as well have used Alphabetic or whatever other property happens to be true for characters that have Equivalent_Unified_Ideograph mappings?

eggrobin: > It looks like the "In" syntax and its regex limits the test to those characters that map to single ones

No, it limits the test to nonempty values (`/./`, not `/^.$/`).

> I guess that you are just using NFC_Quick_Check because there is no syntax to just say “I expect the result to be "Yes"”? So you could just as well have used Alphabetic or whatever other property happens to be true for characters that have Equivalent_Unified_Ideograph mappings?

Yes, This is incredibly stupid. I should add a way to have a literal "Yes" there. Maybe not in this PR.

markusicu (approved):

PR #676, opened 2024-01-31
UTC-178-A13: jg long alias Teh_Marbuta_Goal

markusicu:
UTC-178-A13 Action Item for Mark Davis, PAG: Make the proposed change to PropertyValueAliases.txt to fix the long value alias for jg=Teh_Marbuta_Goal. For Unicode 16.0. See L2/24-009 item 2.6.

The code (with a function called "add unique") had inadvertently put the third alias into the slot for the long=second one because that happened to be the same as the first one.

### DIFF
--- a/unicodetools/data/ucd/dev/PropertyValueAliases.txt
+++ b/unicodetools/data/ucd/dev/PropertyValueAliases.txt
@@ -1,5 +1,5 @@
 # PropertyValueAliases-16.0.0.txt
-# Date: 2024-01-23, 01:51:24 GMT
+# Date: 2024-01-31
 # © 2023 Unicode®, Inc.
 # Unicode and the Unicode Logo are registered trademarks of Unicode, Inc. in the U.S. and other countries.
 # For terms of use, see https://www.unicode.org/terms_of_use.html
@@ -1085,7 +1085,7 @@ jg ; Syriac_Waw                       ; Syriac_Waw
 jg ; Tah                              ; Tah
 jg ; Taw                              ; Taw
 jg ; Teh_Marbuta                      ; Teh_Marbuta
-jg ; Teh_Marbuta_Goal                 ; Hamza_On_Heh_Goal
+jg ; Teh_Marbuta_Goal                 ; Teh_Marbuta_Goal                 ; Hamza_On_Heh_Goal
 jg ; Teth                             ; Teth
 jg ; Thin_Yeh                         ; Thin_Yeh
 jg ; Vertical_Tail                    ; Vertical_Tail
--- a/unicodetools/src/main/java/org/unicode/props/UcdPropertyValues.java
+++ b/unicodetools/src/main/java/org/unicode/props/UcdPropertyValues.java
@@ -1305,7 +1305,7 @@ public class UcdPropertyValues {
         Tah("Tah"),
         Taw("Taw"),
         Teh_Marbuta("Teh_Marbuta"),
-        Hamza_On_Heh_Goal("Teh_Marbuta_Goal"),
+        Teh_Marbuta_Goal("Teh_Marbuta_Goal", "Hamza_On_Heh_Goal"),
         Teth("Teth"),
         Thin_Yeh("Thin_Yeh"),
         Vertical_Tail("Vertical_Tail"),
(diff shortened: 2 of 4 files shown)


roozbehp (dismissed):
LGTM

Manishearth (approved):

Ken-Whistler (changes requested):
You should also update the copyright line to 2024, unless the intent is that gets handled by a separate PR on a group of files later. Otherwise the change for the jg alias looks fine.

markusicu:
> You should also update the copyright line to 2024, unless the intent is that gets handled by a separate PR on a group of files later.

Good catch, but we should do that in a separate change for all of the generated files.

markusicu:
@eggrobin @macchiati when I regenerate the UCD files, the intended change here is reverted :-( Why??

Manishearth (dismissed):

eggrobin:
> when I regenerate the UCD files, the intended change here is reverted :-( Why??

PVA is not a source file if I recall correctly, instead it is expelled from the guts of ToolUnicodePropertySource, UCD_Name, UCD_Types, and friends; but I cannot immediately spot the culprit there.

eggrobin:
My uneducated guess is that it comes from here:

https://github.com/unicode-org/unicodetools/blob/22efd1fe65d90df1b23d2eb0ee23f38fe0f3bbff/unicodetools/src/main/java/org/unicode/text/UCD/ToolUnicodePropertySource.java#L1853-L1856

Which is used here:

https://github.com/unicode-org/unicodetools/blob/22efd1fe65d90df1b23d2eb0ee23f38fe0f3bbff/unicodetools/src/main/java/org/unicode/text/UCD/ToolUnicodePropertySource.java#L2076-L2082

There may be a bug in `lookup` when the main = aux and aux2 is present.

macchiati (dismissed):

eggrobin:
Obviously pulling on that thread brings various bits of MakeUnicodeFiles crashing down.

I think I see how to fix that, I will push a commit to this branch tomorrow.

macchiati:
Hmmm. When you run GenerateEnums it should make the necessary changes in
the PVA data within the props directory. So if GenerateEnums was not run,
that could be the problem. On the other hand, the change to UCDProperties
seems to indicate that that was done.

eggrobin:
> When you run GenerateEnums it should make the necessary changes in
the PVA data within the props directory. So if GenerateEnums was not run,
that could be the problem.

That is not how it works, neither PVA nor the generated UCDProperties are actually authoritative. The source of truth is the bad old ToolsUnicodePropertySource and UCD_Names, and MakeUnicodeFiles generates PVA based on that and nothing else.

This would be relatively straightforward to fix for Joining_Type, were it not for the fact that it exposes all of the grievous hackery around CCC between ToolUnicodePropertySource and MakeUnicodeFiles.

markusicu:
> My uneducated guess is that it comes from here

Thanks for the tip!
That code actually basically looks fine. The problem is that `lookup()` adds aliases using `UnicodeProperty.addUnique(some-alias, result);`

It faithfully adds the short name Teh_Marbuta_Goal, then the long name Teh_Marbuta_Goal, then the alias Hamza_On_Heh_Goal. But *because it "adds unique" aliases, it drops the long name*.

eggrobin:
> It faithfully adds the short name Teh_Marbuta_Goal, then the long name Teh_Marbuta_Goal, then the alias Hamza_On_Heh_Goal. But because it "adds unique" aliases, it drops the long name.

Yes, and removing the uniques fixes that. But then CCC gets all messed up when you regenerate the UCD. (and DT, but that one is an easy fix.)

eggrobin:
𒌝𒈠 𒀀𒈾𒆪𒈠
> I will push a commit to this branch tomorrow (or rather today, my time zone).

Done; and with that I will get back to those Akkadian exercises I was working on…

eggrobin (approved):
Approving so things do not get blocked, but to be honest I am approving my own code here, so someone else should really look at this.

markusicu:
> Done

Thank you very much!!
I was trying to always add the long name to the list, but ran into the ccc and dt problems -- and tried to hack around those.
Your fix reduces the hackery, so that's much better!

Anyone else opposed?

PR #681, opened 2024-02-02
Miscellaneous fixes to IUP

eggrobin:
Resolve `#` in Name as well as `<code point>`, and return a correct UnicodeMap instead of one with unresolved defaults.
Factor out some other paths that had some other resolution logic.

Remove the corresponding shim, as well as the shims for the properties fixed in in #657 and #676.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java
+++ b/unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java
@@ -686,7 +686,8 @@ public class IndexUnicodeProperties extends UnicodeProperty.Factory {
                 for (UnicodeMap.EntryRange<String> range : raw.entryRanges()) {
                     if (range.codepoint == -1) {
                         newMap.put(range.string, range.value);
-                    } else if (DefaultValueType.forString(range.value) == DefaultValueType.CODE_POINT
+                    } else if (DefaultValueType.forString(range.value)
+                                    == DefaultValueType.CODE_POINT
                             || (prop == UcdProperty.Name && range.value.endsWith("#"))) {
                         for (int c = range.codepoint; c <= range.codepointEnd; ++c) {
                             newMap.put(c, resolveValue(range.value, c));
--- a/unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java
+++ b/unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java
@@ -688,7 +688,7 @@ public class IndexUnicodeProperties extends UnicodeProperty.Factory {
                     if (DefaultValueType.forString(range.value) == DefaultValueType.CODE_POINT
                             || (prop == UcdProperty.Name && range.value.endsWith("#"))) {
                         for (int c = range.codepoint; c <= range.codepointEnd; ++c) {
-                            newMap.put(c, _getValue(c));
+                            newMap.put(c, resolveValue(range.value, c));
                         }
                     } else {
                         newMap.putAll(range.codepoint, range.codepointEnd, range.value);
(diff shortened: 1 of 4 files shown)


unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java
```diff
@@ -10,7 +10,6 @@
 import com.ibm.icu.text.Normalizer2;
```
macchiati, line 10: The main reason for having special values for code point and name was to reduce the size of the data in memory. ( I think the .bin files are also zipped so at least the name part probably isn't so bad on disk). 

That probably isn't a big issue having (eg) 80000 extra name strings at this point, so I'm ok with these changes.

unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java
```diff
@@ -535,28 +534,12 @@ private UnicodeMap<String> getCachedMap(UcdProperty prop2, String sourceFileName
...
-        if (value == null && props != null) {
-            if (getResolvedDefaultValueType(prop) == DefaultValueType.CODE_POINT) {
-                return codepoint;
-            }
-        }
-        if (prop == UcdProperty.Name && value != null && value.endsWith("-#")) {
-            return value.substring(0, value.length() - 1) + Utility.hex(codepoint);
-        }
-        return value;
+        return props.getProperty(prop).getValue(codepoint.codePointAt(0));
```
macchiati, line 537: The purpose of this method is to counteract the 'compression' by using # in names and <code point>. If that isn't done any more, then these methods could be refactored away.

unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java
```diff
@@ -693,17 +676,48 @@ class IndexUnicodeProperty extends UnicodeProperty.BaseProperty {
             }
         }
 
+        @Override
         protected UnicodeMap<String> _getUnicodeMap() {
+            var raw = _getRawUnicodeMap();
```
macchiati, line 681: Ah, looks like you are leaving them in the internal map, but then refashioning the map. Would be better to do this once in the internal map instead, probably.

macchiati (changes requested):
Mostly on track; had a few questions.

unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java
eggrobin: I think the size of the data in memory remains the same; getUnicodeMap only transiently creates the full map.

unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java
eggrobin: But then we are back to high memory usage.

eggrobin: For additional context: I started looking at that while trying to make the versioned JSPs use less memory by storing only differences for earlier versions; right now the staging JSPs have a container with more memory than they used to so they can hold the whole UCD history.

eggrobin: (And in fact that container eventually runs out of memory while loading Unicode Version 5.0. The fact that it takes it about twenty hours to reach Unicode 5.0 when on any of my machines it loads all history in 15ish minutes raises questions as to the cultivar of potato we run the JSPs on, but there is room for performance improvement here both in precomputation and memory usage.)

unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java
macchiati: Right. It is a different issue (performance) if the rebuilding is repeated a lot during normal usage.

unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java
macchiati: Yup, looks good

macchiati:
So looking good; let me know when you want to move from Draft.

eggrobin:
Hm, we lose four minutes on the « build and test » part of CI, so this definitely requires more careful thought.

eggrobin:
I found the performance issue (`isTrivial` in VersionedProperty, where we were constructing hashsets containing the entire code space).

There are many more performance improvements to be done by hoisting various things into UnicodeProperty (for instance, the JSPs have a `getAvailableValues().size()` which does not seem like a good idea when there are more than a million available values), but let’s deal with that in subsequent PRs.

macchiati (approved):

PR #685, opened 2024-02-05
Fewer regices

eggrobin:
Fix `getSet` so it correctly handles matchers that match `null`.
Make the `NULL_MATCHER` strict now that we distinguish `<none>` from empty strings.
Make the `NULL_MATCHER` public and special-case it for performance.
Make the syntax for the null matcher more consistent with already-sort-of-reserved syntax in UTS18 and UAX44 instead of making ∅ special. See unicode-org/properties#240.
Instead of `\p{Meow=/./}` for the set of code points for which `Meow` has a nonempty value, use `\P{Meow=@none@}` (the set of code points for which `Meow` has a value. This would include empty values, if there were any, though that is not relevant here.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
+++ b/unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
@@ -450,7 +450,7 @@ public abstract class UnicodeProperty extends UnicodeLabel {
                     usi.next(); ) { // int i = 0; i <= 0x10FFFF; ++i
                 int i = usi.codepoint;
                 String value = getValue(i);
-                if (value != null && matcher.test(value)) {
+                if (matcher.test(value)) {
                     result.add(i);
                 }
             }
@@ -459,6 +459,19 @@ public abstract class UnicodeProperty extends UnicodeLabel {
         List<String> valueAliases = new ArrayList<>(1); // to avoid reallocating...
         List<String> partAliases = new ArrayList<>(1);
         UnicodeMap<String> um = getUnicodeMap_internal();
+        if (matcher.test(null)) {
+            int previousNullStart = 0;
+            for (var range : um.entryRanges()) {
+                result.addAll(previousNullStart, range.codepoint - 1);
+                previousNullStart = range.codepointEnd + 1;
+            }
+            result.addAll(previousNullStart, 0x10FFFF);
+        }
+        if (matcher == UnicodeProperty.NULL_MATCHER) {
+            // Optimization: The null matcher matches only null, no need to look
+            // at the non-null values.
+            return result;
+        }
         Iterator<String> it = um.getAvailableValues(null).iterator();
         main:
         while (it.hasNext()) {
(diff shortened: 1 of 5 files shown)


unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
```diff
@@ -459,6 +459,19 @@ public UnicodeSet getSet(PatternMatcher matcher, UnicodeSet result) {
...
         UnicodeMap<String> um = getUnicodeMap_internal();
+        if (matcher.test(null)) {
+            int previousNullStart = 0;
+            for (var range : um.entryRanges()) {
+                result.addAll(previousNullStart, range.codepoint - 1);
+                previousNullStart = range.codepointEnd + 1;
+            }
+            result.addAll(previousNullStart, 0x10FFFF);
+        }
+        if (matcher == UnicodeProperty.NULL_MATCHER) {
```
macchiati, line 470: We could put this inside the previous block, since it will only test = if the matcher.test(null), right?

unicodetools/src/main/java/org/unicode/text/UCD/VersionedProperty.java
```diff
@@ -147,8 +147,8 @@ public UnicodeSet getSet(
             }
             matcher.set(body);
             set = property.getSet(matcher);
-        } else if (propertyValue.equals("∅")) {
-            set = property.getSet(NULL_MATCHER, null);
+        } else if (propertyValue.equals("@none@")) {
```
macchiati, line 150: Yes, good move

macchiati (approved):
You may not be done yet, but looking good so far. Need to update the help page also.

macchiati:
> You may not be done yet, but looking good so far. Need to update the help page also.

About the help page. https://unicode-org.github.io/unicodetools/help/list-unicodeset appear kinda flakely; falls back to unicode.org for me right now.

unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
eggrobin: Done.

eggrobin:
Re the help page, we were not documenting ∅ (and it did not work anywhere), but I think given where I am implementing `@none@`, it only works in the invariants, not in the JSPs; so I will update that once I make it work in the JSPs.

macchiati:
https://github.com/unicode-org/unicodetools/blob/main/docs/help/list-unicodeset.md looks fine. Not sure why the github io isn't showing it.

BTW, I noticed that \p{idna2003==@uts46@} (an example on that page) doesn't work anymore. If you are touching that page you could fix, otherwise we could have a different issue for that.

macchiati:
> Re the help page, we were not documenting ∅ (and it did not work anywhere), but I think given where I am implementing `@none@`, it only works in the invariants, not in the JSPs; so I will update that once I make it work in the JSPs.

makes sense

macchiati (approved):

PR #686, opened 2024-02-06
Make the invariant tests a bit more understandable, test and fix IUP Jamo_Short_Name

eggrobin:
Fix Jamo_Short_Name, which was incorrectly null instead of empty for U+110B HANGUL CHOSEONG IEUNG.
See [the test failure before the fix](https://github.com/unicode-org/unicodetools/pull/686/commits/6b234062b7a38f5b41c5fd207f2a03257169954f#diff-c2f2f053bbaa14f43053e3fb81a8afd67add46a4be1377d3e411f0500b40dabfR994).

Also rework the DSL a bit:
* Do not implicitly make the application of a non-string property to a string mean the _set_ of values, as this has misled Markus and me (we thought a binary property applied to a two-character string would yield YesYes or somesuch, see https://github.com/unicode-org/unicodetools/pull/673#discussion_r1473392606), and can readily lead to bugs since toNFWhatever was not treated as a string for these purposes! Instead make those return a sequence.
* Introduce some operations to deal with those sequences.
* Syntax things:
  * Require a separator between In ⟨domain⟩ and ⟨<i>left\_</i>props⟩ = ⟨<i>right\_</i>props⟩, as suggested by Markus, since this can otherwise be hard to read if the left-hand-side is a composition, especially if it includes set filters.
    * Markus had suggested a colon, but the colon is already used in the syntax for version prefixes (U-1:General_Category or U5.0.0:General_Category), so a comma seemed better. (I also think it reads more naturally, but that may be a French thing.)
  * Change the EquivalencesOf/partition refinement syntax which I was never satisfied with; if we can require a comma, we can require words that make it make some semblance of sense. The new syntax is the more explicit
    > OnPairsOf ⟨domain⟩, EqualityOf ⟨<i>left\_</i>props⟩ ⟨relation⟩ EqualityOf ⟨<i>right\_</i>props⟩

### DIFF
--- a/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
@@ -144,6 +144,7 @@ public class TestUnicodeInvariants {
      * @throws IOException
      */
     public static int testInvariants(String inputFile, boolean doRange) throws IOException {
+        TestUnicodeInvariants.doRange = doRange;
         parseErrorCount = 0;
         testFailureCount = 0;
         boolean showScript = false;
--- a/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
@@ -241,7 +241,7 @@ public class TestUnicodeInvariants {
                                 showMapLine(line, pp);
                             } else if (line.startsWith("Show")) {
                                 showLine(line, pp);
-                            } else if (line.startsWith("EquivalencesOf")) {
+                            } else if (line.startsWith("OnPairsOf")) {
                                 equivalencesLine(line, pp, lineNumber);
                             } else {
                                 testLine(line, pp, lineNumber);
@@ -278,12 +278,16 @@ public class TestUnicodeInvariants {
 
     private static void equivalencesLine(String line, ParsePosition pp, int lineNumber)
             throws ParseException {
-        pp.setIndex("EquivalencesOf".length());
+        pp.setIndex("OnPairsOf".length());
         final UnicodeSet domain = new UnicodeSet(line, pp, symbolTable);
+        expectToken(",", pp, line);
+        expectToken("EqualityOf", pp, line);
         final var leftProperty = CompoundProperty.of(LATEST_PROPS, line, pp);
         scan(PATTERN_WHITE_SPACE, line, pp, true);
         char relationOperator = line.charAt(pp.getIndex());
         pp.setIndex(pp.getIndex() + 1);
+        scan(PATTERN_WHITE_SPACE, line, pp, true);
+        expectToken("EqualityOf", pp, line);
         final var rightProperty = CompoundProperty.of(LATEST_PROPS, line, pp);
 
         boolean leftShouldImplyRight = false;
@@ -518,11 +522,22 @@ public class TestUnicodeInvariants {
         }
     }
 
+    private static void expectToken(String token, ParsePosition pp, String line)
+            throws ParseException {
+        scan(PATTERN_WHITE_SPACE, line, pp, true);
+        if (!line.substring(pp.getIndex()).startsWith(token)) {
+            throw new ParseException("Expected " + token, pp.getIndex());
+        }
+        pp.setIndex(pp.getIndex() + token.length());
+        scan(PATTERN_WHITE_SPACE, line, pp, true);
+    }
+
     private static PropertyComparison getPropertyComparison(ParsePosition pp, String line)
             throws ParseException {
         final PropertyComparison propertyComparison = new PropertyComparison();
 
         propertyComparison.valueSet = new UnicodeSet(line, pp, symbolTable);
+        expectToken(",", pp, line);
         propertyComparison.property1 = CompoundProperty.of(LATEST_PROPS, line, pp);
         final int cp = line.codePointAt(pp.getIndex());
         if (cp != '=' && cp != '≠') {
--- a/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
@@ -17,10 +17,13 @@ import java.text.ParsePosition;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashMap;
-import java.util.LinkedHashSet;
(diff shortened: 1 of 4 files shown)


eggrobin:
Closing and reopening since somehow CI did not pick this up.

unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
```diff
@@ -53,25 +53,37 @@
...
 # is (=|≠) the result of applying the right <props>.
-#   <props> is of the form (<unicodeSet> | <prop>) ("*" (<unicodeSet> | <prop>))?
+#   <props> is of the form (<unicodeSet> | <prop>) ("*" (<unicodeSet> | <prop> | <sequenceTransformation>))?
 #   It is the functional composition of the properties applied to strings, whereby
 #   <unicodeSet> is used to filter the result.
 #   <prop> for a string property is applied to each character, and the result concatenated
 #       That is, cf("A1") is cf("A")+cf("1") = "a1"
-#   <prop> for an enumerated property, is applied to each character, and the result is a concatenated set.
-#       That is, gc("A1") is gc("A")+gc("1") = "Uppercase_LetterDecimal_Number"
+#   <prop> for an enumerated property is applied to each character, and the result is sequence of
```
markusicu, line 61: insert "a": ... is a sequence...

unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
```diff
@@ -53,25 +53,37 @@
...
+#   strings.
+#       That is, gc("A1") is [gc("A"), gc("1")] = ["Uppercase_Letter", "Decimal_Number"]
+#   <sequenceTransformation> may be applied to a seqence or strings or to a string.  On a string it
```
markusicu, line 64: seqence or --> sequence of

unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
```diff
@@ -53,25 +53,37 @@
...
 #   Example: for <props> of bc * \P{bc=NSM} * cf * dm, the result applied to Å (angstrom sign) are:
 #       bc * \P{bc=NSM} * cf * dm ("Å")
 #       bc * \P{bc=NSM} * cf ("A" + umlaut)
 #       bc * \P{bc=NSM} ("a" + umlaut)
 #       bc ("a")
-#       "Left"
+#       "Left_To_Right"
 #
-#   Example: In \p{dt=canonical} bc * \P{bc=NSM} * dm = bc * \P{bc=NSM}
+#   Example: In \p{dt=canonical} (delete-adjacent-duplicates) * bc * \P{bc=NSM} * dm = bc * \P{bc=NSM}
```
markusicu, line 82: A separator after the "In set" would be nice

unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
```diff
@@ -53,25 +53,37 @@
...
 #       This examines only those characters that have canonical compositions. For each such character X
-#       it gets the decomposition mapping of X, then filters out all NSM characters, then gets the Bidi_Class.
-#       It then tests that against the result of filtering out NSM characters from X, then getting the BIDI_Class.
+#       it gets the decomposition mapping of X, then filters out all NSM characters, then gets the Bidi_Class,
+#       then deduplicates runs of the same Bidi_Class.
+#       It then compares that with the result of filtering out NSM characters from X, then getting the Bidi_Class.
```
markusicu, line 86: What happens if filtering out NSMs from x yields an empty string? What does applying bc to that do?

unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
```diff
@@ -963,9 +975,7 @@ Let $nonIdeographicStrokes = \p{Name=/^CJK STROKE (T|WG|XG|BXG|SW|HZZ|HP|HZWG|SZ
 
 \P{Equivalent_Unified_Ideograph=@none@} ⊆ $strokesAndRadicals
 [$strokesAndRadicals - \P{Equivalent_Unified_Ideograph=@none@}] = [$nonIdeographicStrokes $nonIdeographicRadicals]
-# TODO(egg): NFC_Quick_Check is a stupid way to get a Yes here; we are checking
-# that Equivalent_Unified_Ideograph values are single unified ideographs.
-In \P{Equivalent_Unified_Ideograph=@none@} Unified_Ideograph * Equivalent_Unified_Ideograph = NFC_Quick_Check
+In \P{Equivalent_Unified_Ideograph=@none@} Unified_Ideograph * Equivalent_Unified_Ideograph = (constant Yes)
```
markusicu, line 978: "(constant Yes)" not yet documented?

... but a nice addition!

unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
eggrobin: It returns the sequence of bc for every code point in the empty string; that is the empty sequence.

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -571,6 +593,87 @@ static UnicodeProperty of(
                     propOrFilter.filter = parseUnicodeSet(line, pp);
                     propOrFilter.type = FilterOrProp.Type.filter;
                     result.propOrFilters.add(propOrFilter);
+                } else if (line.charAt(pp.getIndex()) == '(') {
+                    final FilterOrProp propOrFilter = new FilterOrProp();
+                    final var matcher =
+                            Pattern.compile("(\\( *([^ )]+)(?: +([^)]+))? *\\)).*")
+                                    .matcher(line.substring(pp.getIndex()));
+                    if (matcher.matches()) {
```
macchiati, line 601: Minor, but I find the code more readable if the failure condition is first, rather than disconnected by being at the end of a bunch of code.

if (!matcher.matches()) {
  // handle error
} else {
 //many lines of code
}

macchiati: But your choice

macchiati (approved):

macchiati (approved):

PR #688, opened 2024-02-07
168-A63

eggrobin:
[[168-A63](https://www.unicode.org/cgi-bin/GetL2Ref.pl?168-A63)] Action Item for Mark Davis: Add an invariant test to make sure that the three properties, kTotalStrokes, kRSUnicode, and at least one value for an IRG source, are defined for all CJK ideographs.


markusicu (approved):
nice!

PR #689, opened 2024-02-08
Invariants for some more Decomposition_Mapping stability policies

eggrobin:


unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
```diff
@@ -446,22 +446,44 @@ $Unicode_6_1_Decompositions ⊆ [\p{Name=/^CHAKMA VOWEL SIGN /}]
...
+# always limited either to a single value or to a pair. The second character in
+# the pair cannot itself have a canonical mapping.
 [ \p{Decomposition_Type=Canonical} & \p{Decomposition_Mapping=/.../} ] = []
-# TODO: The second character in the pair cannot itself have a canonical mapping.
-
-# Stability: Canonical mappings (Decomposition_Mapping property values) are always limited so that no string when normalized to NFC expands to more than 3× in length (measured in code units). 
+[ \p{Decomposition_Type=Canonical} & \p{Decomposition_Mapping=} ] = []
+Let $expandingCanonicalDecompositions = [ \p{Decomposition_Type=Canonical} & \p{Decomposition_Mapping=/../} ]
+In $expandingCanonicalDecompositions, Decomposition_Type * (drop 1) * Decomposition_Mapping ≠ (constant Canonical)
+# In fact the second character does not have a decomposition mapping at all,
```
macchiati, line 463: I think we need to make clear in the comments what are real failures, and what happens to be the case — so we want to be notified if it breaks — but is not a true invariant failure.

macchiati (dismissed):

unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
eggrobin: Reworded to clarify.

macchiati (approved):

PR #692, opened 2024-02-09
NamesList-16.0.0d13.txt

eggrobin:
@Ken-Whistler wrote:
> Another alpha review feedback item: I updated the subhead for 16100 in Gurung Khema.


Ken-Whistler (approved):
As expected.

PR #699, opened 2024-02-14
Store only changes from the next version for older copies of the UCD

eggrobin:
Store a magic value in the map if the value is the same as in the subsequent version (for the dev version, if it is the same as the latest released version); instead of running out of 4 GB of memory on the container, this means we take ~1.1 GB. The .BIN files are part of the image, so loading them takes only a couple of minutes (so long as the container is configured to actually have its CPU); the JSPs start being served even faster, within ~20 s, as soon as the current UCD is loaded (requests for history will return only as much history as is loaded).

This compression is disabled outside the JSPs: I got the invariant tests to pass, but lots of things (especially in emoji) are accessing the maps directly (which was already bad, as they are getting literal `<code point>` and the like and treating that as a property value) so I want to deal with that separately in #716.

See it live at https://unicode-jsps-staging-o2ookmn2oq-uc.a.run.app/UnicodeJsps/character.jsp?a=3400&history=assigned&showDevProperties=1.


eggrobin:
> A question. I presume that the last release version and the current development version are stored completely, with references to any other versions.

Only the current dev version is stored completely; the last release version is stored like the older versions (but the older the version the fewer pointers to the next version need to be followed, so the last release version is still very fast).

macchiati:
So the last release version needs to be rebuilt any time the dev version is
rebuilt?

eggrobin:
> So the last release version needs to be rebuilt any time the dev version is rebuilt?

You mean in the Generated directory while working on UCD changes?
I suppose so, but in general it is a bad idea not to rebuild that; I have wasted some time on false invariant test failures resulting from having an inconsistent Generated directory after switching branches with our new process where the pipeline is a hydra of draft pull requests.
Indeed the standard operating procedures for UCD development start by clearing it: https://github.com/unicode-org/unicodetools/blob/main/docs/pipeline.md#merge.

Come to think of it, that probably means I probably ought to devise a mechanism to detect an inconsistent Generated directory. But that is another another can of worms for another day.

macchiati:
I think we should be able to have the generated directory *for released
versions* to be immutable, except when there is a change in the format for
the bin files.

eggrobin:
> I think we should be able to have the generated directory *for released
versions* to be immutable, except when there is a change in the format for
the bin files.

Well, at the very least it will be mutated whenever a new version comes out (since the last released version, no longer being the last released version, will become a diff to the newly released version).
And I am not sure we track the changes to the format of bin files well enough to be able to take advantage of that.
But OK, it is not too onerous to change things to make the last released version the reference and dev be a diff from that (backward rather than forward for the historical versions), so I will do that.

eggrobin:
> But OK, it is not too onerous to change things to make the last released version the reference and dev be a diff from that (backward rather than forward for the historical versions), so I will do that.

Done, rebuilt, and deployed to staging.

macchiati (approved):

PR #701, opened 2024-02-16
Add Unihan properties from Unicode 14.0, 15.0, 15.1, and 16.0

eggrobin:
While playing with the staging JSPs which have Unihan properties since #659, @Manishearth had lamented the lack of kStrange.


Manishearth (approved):

macchiati:
Btw I was thinking of a test that every .txt file is either a source for
props or is on an exclusion list. That would catch new files

PR #707, opened 2024-02-17
14 letters with palatal hook

eggrobin:
L2/24-(TODO forthcoming)
See https://github.com/unicode-org/sah/issues/350.


unicodetools/data/ucd/dev/DerivedCoreProperties.txt
```diff
@@ -1371,6 +1371,7 @@ FFDA..FFDC    ; Alphabetic # Lo   [3] HALFWIDTH HANGUL LETTER EU..HALFWIDTH HANG
 1DF0A         ; Alphabetic # Lo       LATIN LETTER RETROFLEX CLICK WITH RETROFLEX HOOK
 1DF0B..1DF1E  ; Alphabetic # L&  [20] LATIN SMALL LETTER ESH WITH DOUBLE BAR..LATIN SMALL LETTER S WITH CURL
 1DF25..1DF2A  ; Alphabetic # L&   [6] LATIN SMALL LETTER D WITH MID-HEIGHT LEFT HOOK..LATIN SMALL LETTER T WITH MID-HEIGHT LEFT HOOK
+1DF2D..1DF3A  ; Alphabetic # L&  [14] LATIN SMALL LETTER D WITH HOOK AND PALATAL HOOK..LATIN LETTER PHARYNGEAL VOICED FRICATIVE WITH PALATAL HOOK
```
nedley, line 1374: I find it curious that even a range consisting of only `Ll` gets `L&` but this is by no means new.

unicodetools/data/ucd/dev/UnicodeData.txt
```diff
@@ -35658,6 +35658,20 @@ FFFD;REPLACEMENT CHARACTER;So;0;ON;;;;;N;;;;;
 1DF28;LATIN SMALL LETTER R WITH MID-HEIGHT LEFT HOOK;Ll;0;L;;;;;N;;;;;
 1DF29;LATIN SMALL LETTER S WITH MID-HEIGHT LEFT HOOK;Ll;0;L;;;;;N;;;;;
 1DF2A;LATIN SMALL LETTER T WITH MID-HEIGHT LEFT HOOK;Ll;0;L;;;;;N;;;;;
+1DF2D;LATIN SMALL LETTER D WITH HOOK AND PALATAL HOOK;Ll;0;L;;;;;N;;;;;
+1DF2E;LATIN SMALL LETTER DZ DIGRAPH WITH PALATAL HOOK;Ll;0;L;;;;;N;;;;;
+1DF2F;LATIN SMALL LETTER ETH WITH PALATAL HOOK;Ll;0;L;;;;;N;;;;;
+1DF30;LATIN LETTER SMALL CAPITAL G WITH STROKE AND PALATAL HOOK;Ll;0;L;;;;;N;;;;;
```
nedley, line 35664: Has this name gained an extra `WITH STROKE`?

eggrobin: Hmm, it seems it lost it between the document that was seen in SAH (the one dated February 2 on the SAH issue) and the one posted to the registry.

Great that Kirk corrected his mistake in his UnicodeData lines (the name was correct elsewhere in the proposal, this is really a typo), but it would be nice of him to call that out…

unicodetools/data/ucd/dev/DerivedCoreProperties.txt
eggrobin: Yes, this makes me sad. The Kenfiles (EastAsianWidth, LineBreak, VerticalOrientation) do that right, I wrote some code for that when I made the tools capable of regenerating them; but the derived files were doing it in a simpler and uglier way.

nedley: I figured as much. This was really just a drive-by observation and was not meant to be part of my review.

roozbehp (changes requested):
Please add the DoNotEmit data too.

eggrobin:
> Please add the DoNotEmit data too.

Done. If we are going to get much more of these kinds of proposals, we need to find a way to incorporate validation and regeneration of DoNotEmit.txt in the tools; sorting these entries by hand is error-prone, and having data that is not validated is a good way to have bugs (I could easily have missed https://github.com/unicode-org/sah/issues/350#issuecomment-2018043770). 

(The comments also want to be generated; I did not check them, and we should not have to check them.)

roozbehp:
> If we are going to get much more of these kinds of proposals, we need to find a way to incorporate validation and regeneration of DoNotEmit.txt in the tools; sorting these entries by hand is error-prone, and having data that is not validated is a good way to have bugs (I could easily have missed [unicode-org/sah#350 (comment)](https://github.com/unicode-org/sah/issues/350#issuecomment-2018043770)).

I agree on principle, but I don't understand unicodetools enough to contribute to tooling.

Code reviewers are supposed to catch these issues.

> (The comments also want to be generated; I did not check them, and we should not have to check them by hand.)

I have a Python script that deletes the character name comments and regenerates them. I can rerun it after these are merged, to double check.

PR #713, opened 2024-02-20
16.0 hieroglyphs in the UCA

eggrobin:
From Ken, 2024-02-17T04:26+01:00:

> unidata-16.0.0d14.txt
> allkeys-16.0.0d14.txt
> This adds the 3995 hieroglyphs for 16.0.


markusicu (approved):
rslgtm tnx
@Ken-Whistler FYI

eggrobin:
> rslgtm

does not look good to CI though. Investigating.

eggrobin:
> does not look good to CI though. Investigating.
@markusicu: I have no idea what is going on, so all yours to investigate further…
I did see one message
> reordering group {Egyp Merc Mero Hluw Tang Nshu Kits} marked for compression but uses more than one lead byte 7F..80

which seems to be fixed by setting .minimalGap3() in PrimariesToFractional, but that does not help with the various mysterious errors in the validity log.

Ken-Whistler:
This has to do with the addition of 3995 new hieroglyphs, which apparently broke the bank for some ICU assumptions about how it could compress keys. allkeys-16.0.0d12.txt Egyptian through Anatolian primary weights 5C79..6310 (0x597 values). allkeys-16.0.0d15.txt Egyptian through Anatolian primary weights 5C85..72B7 (0x1532 values).

eggrobin:
> that does not help with the various mysterious errors in the validity log.

Markus tells we are ignoring those errors, which are not new, and indeed they do not break the build.
Errors are warnings, ignorance is strength, etc.

So minimalGap3 is enough to unbreak the ICU bank.

markusicu (approved):
minimal gap lgtm as discussed in unicodetools meeting

We might be able to do a little more optimization like this and save a lead byte again, but that will be for another PR.

PR #718, opened 2024-02-26
Ideographic invariants

eggrobin:
This would have caught https://github.com/unicode-org/properties/issues/250 (PRI-497#ID20240216140104, which was fixed by #715).


markusicu (approved):
lgtm
This looks like a PR where it makes sense to squash the commits.

eggrobin:
> This looks like a PR where it makes sense to squash the commits.

Squash-and-merge is the default here, so that is what I do anyway.

Issue #650, opened 2024-01-16
VersionedProperty is garbled when reading some old files


eggrobin:
Spotted while working on #432, I got character.jsp to tell me that U+0041 went

Identifier_Type | 9.0:Not_Character | 10.0:Recommended
-- | -- | --

Which seems wrong, and indeed is; the default value changed between 9.0 and 10.0, from Recommended to Not_Character.

Likewise for U+4E00:

Block | 2.0..3.0: 9FFF | 3.1..15.1: CJK_Unified_Ideographs
-- | -- | --

(The format of Blocks.txt changed after 3.0, from first; last; `<block>` to first..last; `<block>`.)

eggrobin:
Broadened this issue to cover the wider issue of files changing format. Cleverness will be needed at some point.

macchiati:
Documentation needs a bit of work!

Blocks:
There is provision for variant formats in the data files; look at IndexUnicodeProperties, eg:

# Unicode 13 moves kTotalStrokes to Unihan_IRGSources.txt.
# The line with the new location (i.e., the line without version number)
# must occur in this file before the line with the old location.
Unihan_DictionaryLikeData ; kTotalStrokes ; v12.1

For (unique) radical changes like that, it may be better to special case in the code. Ugg.

Identifier_Type:
We could extend ExtraPropertyValueAliases.txt in the same way; a trailing Age means that age or below.

PR #719, opened 2024-02-27
Old blocks and unborn properties

eggrobin:
Some progress towards #650.

| | Before | After |
|-|---|--|
| A | image | image |
| 㐀 | image | image |

### DIFF
--- a/unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
+++ b/unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
@@ -24,6 +24,7 @@ import org.unicode.cldr.util.RegexUtilities;
 import org.unicode.idna.Regexes;
 import org.unicode.props.IndexUnicodeProperties.DefaultValueType;
 import org.unicode.props.PropertyUtilities.Merge;
+import org.unicode.props.UcdLineParser.IntRange;
 import org.unicode.text.utility.Settings;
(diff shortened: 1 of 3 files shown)


unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
```diff
@@ -1044,6 +1045,25 @@ private static void parseSimpleFieldFile(
                         setPropDefault(propInfo.property, defaultValue, "hardcoded", false);
                     }
                 }
+                if (line.getParts().length != 2) {
```
markusicu, line 1048: optional: test for ` == 3`  ?

markusicu (approved):

unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
eggrobin: Well, I really want to fail if it is anything other than 2 so we don’t miss it.

markusicu: I am ok with what you have, but you could be more specific:
```
if (3 parts and prop is Block) {
    // handle old Blocks.txt
    continue;
}
if (not 2 parts) {
    throw new ...
}
```

markusicu (approved):

PR #722, opened 2024-02-28
Fix U9.0:Identifier_Type defaults

eggrobin:
#650 intensifies.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
+++ b/unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
@@ -17,6 +17,7 @@ import java.util.Map;
 import java.util.MissingResourceException;
 import java.util.Objects;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.regex.Matcher;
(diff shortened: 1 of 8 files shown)


unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
```diff
@@ -50,7 +51,7 @@ enum SpecialProperty {
     public String oldFile;
     public VersionInfo maxOldVersion = MIN_VERSION;
 
-    String defaultValue;
+    TreeMap<VersionInfo, String> defaultValues = new TreeMap<>();
```
markusicu, line 54:
```suggestion
    /**
     * Maps from Unicode versions to default values.
     * A property whose default value depends on the version has more than one entry.
     * A particular default value applies to the Unicode versions after the previous-version entry,
     * up to and including its own version.
     */
    TreeMap<VersionInfo, String> defaultValues = new TreeMap<>();
```

markusicu (dismissed):

markusicu (approved):

Issue #46, opened 2021-03-02
Uploading UnicodeJsps


macchiati:
The generation and local testing of UnicodeJsps is unblocked. Two remaining items are:

- how to debug locally in Eclipse
- how to upload to production

(Both of these have some documentation; I have to verify that that documentation is sufficient.)

macchiati:
@srl295

srl295:
I'll document the debugging in #43, this one will be for automatically uploading.

srl295:
- also see #1007

PR #724, opened 2024-03-01
UTS #46 minimize base exclusion set

markusicu:
Minimize the UTS46 base exclusion set.
Make some characters valid that are currently disallowed because their mappings differ from those in IDNA2003, see https://github.com/unicode-org/properties/issues/243#issuecomment-1972023328

UTC-179-C36 Consensus: In UTS #&zwnj;46, the base exclusion set has been derived from differences between IDNA2003 data and UTS #&zwnj;46 principles. Replace this base exclusion set with a smaller set. This will change the IdnaMappingTable.txt Status for some characters from disallowed to ignored/mapped/valid. For details see the proposal in document L2/24-064 item 6.2. For Unicode Version 16.0.

UTC-179-A120 Action Item for Markus Scherer, PAG: In the UTS #&zwnj;46 tools code, the base exclusion set has been derived from differences between IDNA2003 data and UTS #&zwnj;46 principles. Replace this base exclusion set with a smaller set. This will change the IdnaMappingTable.txt Status for some characters from disallowed to ignored/mapped/valid. For details see the proposal in document L2/24-064 item 6.2. For Unicode Version 16.0.


macchiati:
Just looked over: lgtm

Ken-Whistler (dismissed):
I don't understand all the code, but the documentation improvement for the baseExclusionsSet2 looks fine.

markusicu:
@macchiati I started the experiment with comparing the base exclusion set in the code vs. that in the spec: https://unicode.org/reports/tr46/#TableDerivationStep3

The code computes a baseExclusionSet, and it also hardcodes its expected contents into a baseExclusionSet2. There is code to compare the two, but it was turned off.

I added comments to the hardcoded/expected base exclusions and turned the comparison on.

Please review the following discrepancies and what I propose to do with them.

a)
There are three characters in the expected base exclusion set that are not in the spec:
- U+200B ZERO WIDTH SPACE
- U+2060 WORD JOINER
- U+FEFF ZERO WIDTH NO-BREAK SPACE

These are also not in the computed base exclusion set, so I think we should just remove these from the expected set.
In IdnaMappingTable.txt, these three are "ignored" (mapped to empty string).

b)
The expected set, and the set in the spec, also contains the following characters which are not in the computed set: \u200E\u200F\u202A-\u202E

These are the Bidi_Control characters that the spec lists. In the data file, they are all disallowed, so having them in the base exclusion set seems unnecessary but harmless.

However, these are out of sync, probably since Unicode 6.3: Bidi_Control has since added U+061C ALM and U+2066..U+2069 isolate controls. These are also all disallowed in the data file.

I propose that we remove the out-of-date Bidi_Control characters from the spec base exclusion set. In the Modifications we could note that they continue to be disallowed in the output.

c)
The computed base exclusion set contains one character that is not in the spec and not in the code's expected set: U+180E MONGOLIAN VOWEL SEPARATOR. It is therefore disallowed.

It is in the computed base exclusion set because it is disallowed in IDNA2003 and “every code point in C's base mapping value is in the base valid set” -- but trivially so: Its base mapping value is the empty string.

I guess that we should add U+180E to the spec base exclusion set so that it matches reality.

Note that some gc=Cf characters are "ignored" rather than "disallowed" (e.g., SHY, ZWSP). I don't know if that would be an improvement for the Mongolian Vowel Separator, or if we care.

WDYT?

markusicu:
Dug deeper:

The Bidi_Control characters don't get computed into the base exclusion set because their base mapping is removed (map to themselves) rather than NFKC_Casefold mapping them to empty.

I see now that the spec says “For example, for Unicode 5.2 and 6.0, the base exclusion set consists of the list that follows.”
So it's ok if it does not match the current state. That seems to explain U+180E because in Unicode 6 it did not have an NFKC_CF mapping. And it probably explains the Bidi_Control characters, because in [Unicode 6.0 UTS46](https://www.unicode.org/reports/tr46/tr46-5.html#Mapping_Table_Derivation) their base mapping value was not yet removed.

The code that computes the base exclusion set only looks at characters that were in Unicode 3.2. Step 3.3 does not say that unassigned code points should be skipped. If the code didn't skip them, then I think all post-3.2 letters would end up in the base exclusion set.

I think we should change “According to IDNA2003, C is prohibited” to “According to IDNA2003, C is prohibited and not unassigned” -- or say generally that we only look at Age=3.2 characters.

Ken-Whistler:
It seems where you are going on this is fine. But the text of UTS#46 is extraordinarily confusing, particularly if it takes this kind of analysis by someone who actually maintains the code to figure it out and set things straight. I think it is way past time to ditch all the history in UTS#46 and write a much simpler account of the *current* state of affairs.

BTW, I am also highly skeptical of this kind of standards writing where somebody writes a complicated algorithm, and then tries to turn the algorithm into step-by-step prose. That is a recipe for immediate spec obsolescence any time the code has to be tweaked.

asmusf:
> I think it is way past time to ditch all the history in UTS#46 and write a much simpler account of the _current_ state of affairs.

Agree.

markusicu:
Ken: I hear you, but the question is what we do now. And I don't think any of us has time to completely rewrite this thing.
- In the spec, I think we need to clarify that we compute the base exclusion set only on Age=3.2 characters.
- We should probably either update the “for Unicode 5.2 and 6.0, the base exclusion set consists of the list that follows” part to 16.0, especially if we decide to make deliberate changes, or just remove it.
- In the code, I can align the expected set with current reality.
- Then we need to decide whether we still care about characters whose mappings would differ from those in IDNA2003. They are currently disallowed, so we should be ok making them valid/mapped/ignored.
  - This is particularly attractive for the 5 Han characters with Unicode 4.0 normalization corrections, because they complicate optimized implementations.
  - See [`[:uts46=disallowed:]-[:idna2003=disallowed:]` grouped by NFC_QC](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3Auts46%3Ddisallowed%3A%5D-%5B%3Aidna2003%3Ddisallowed%3A%5D&g=NFC_QC&i=)

markusicu:
Ok, I
- sync'ed the expected & actual sets
- made one commit where I removed the five 4.0 normalization correction characters from the base exclusion set, which changes them from disallowed to mapped -- see fbe61299e5a7df98f667ef46fcf4870b3ade38e9
- made another commit where I ignore all of the IDNA2003 vs. UTS46 mapping changes, which makes many of them mapped, makes some fillers ignored, and makes U+1806 MONGOLIAN TODO SOFT HYPHEN simply valid -- see c455769e364f6928c3146e9b7470469226d620ce

Details of the latter change right here:
```diff
713c713
< 04C0          ; disallowed                             # 1.1  CYRILLIC LETTER PALOCHKA
---
> 04C0          ; mapped                 ; 04CF          # 1.1  CYRILLIC LETTER PALOCHKA
1495c1495,1532
< 10A0..10C5    ; disallowed                             # 1.1  GEORGIAN CAPITAL LETTER AN..GEORGIAN CAPITAL LETTER HOE
---
> 10A0          ; mapped                 ; 2D00          # 1.1  GEORGIAN CAPITAL LETTER AN
...
> 10C5          ; mapped                 ; 2D25          # 1.1  GEORGIAN CAPITAL LETTER HOE
1509c1546
< 115F..1160    ; disallowed                             # 1.1  HANGUL CHOSEONG FILLER..HANGUL JUNGSEONG FILLER
---
> 115F..1160    ; ignored                                # 1.1  HANGUL CHOSEONG FILLER..HANGUL JUNGSEONG FILLER
1613c1650
< 17B4..17B5    ; disallowed                             # 3.0  KHMER VOWEL INHERENT AQ..KHMER VOWEL INHERENT AA
---
> 17B4..17B5    ; ignored                                # 3.0  KHMER VOWEL INHERENT AQ..KHMER VOWEL INHERENT AA
1625,1627c1662
< 1800..1805    ; valid                  ;      ; NV8    # 3.0  MONGOLIAN BIRGA..MONGOLIAN FOUR DOTS
< 1806          ; disallowed                             # 3.0  MONGOLIAN TODO SOFT HYPHEN
< 1807..180A    ; valid                  ;      ; NV8    # 3.0  MONGOLIAN SIBE SYLLABLE BOUNDARY MARKER..MONGOLIAN NIRUGU
---
> 1800..180A    ; valid                  ;      ; NV8    # 3.0  MONGOLIAN BIRGA..MONGOLIAN NIRUGU
2486c2521
< 2132          ; disallowed                             # 1.1  TURNED CAPITAL F
---
> 2132          ; mapped                 ; 214E          # 1.1  TURNED CAPITAL F
2559c2594
< 2183          ; disallowed                             # 3.0  ROMAN NUMERAL REVERSED ONE HUNDRED
---
> 2183          ; mapped                 ; 2184          # 3.0  ROMAN NUMERAL REVERSED ONE HUNDRED
3367c3402
< 3164          ; disallowed                             # 1.1  HANGUL FILLER
---
> 3164          ; ignored                                # 1.1  HANGUL FILLER
5730c5765
< FFA0          ; disallowed                             # 1.1  HALFWIDTH HANGUL FILLER
---
> FFA0          ; ignored                                # 1.1  HALFWIDTH HANGUL FILLER
```

Ken-Whistler:
Let me know when this is all merged and you have regenerated IdnaMappingTable.txt. With all the changes, I'll need to run the IdnaDerived.txt derivation over that data again, to ensure that it ends up with exactly the same output as before.

markusicu:
> Let me know when this is all merged and you have regenerated IdnaMappingTable.txt.

I have been regenerating the data file to show what's changing when.
I can't merge this until we decide which way to go.

Ken-Whistler:
@markusicu I've not suggesting you merge prematurely. I'm just requesting that once this PR is complete and merged, that you let me know, because separately, I need to verify the regeneration for Idna2008.txt. I'm not intending to even attempt to do that as part of this PR, which is already headbangingly complicated.

markusicu:
> ... once this PR is complete and merged, that you let me know, because separately, I need to verify the regeneration for Idna2008.txt.

Yes you and I will sync on this as usual.

> I'm not intending to even attempt to do that as part of this PR, which is already headbangingly complicated.

It's getting a lot simpler, especially if y'all agree with my latest version :-)

My latest experiment: Hardcode a nearly empty base exclusion set, ignore what IDNA2003 used to do. I only hardcoded the following which I thought should really remain disallowed (and in ICU I completely rely on U+FFFD being disallowed):
- U+FFFC OBJECT REPLACEMENT CHARACTER
- U+FFFD REPLACEMENT CHARACTER
- U+E00xx tag characters

With this fixed base exclusion set, 17 characters go from disallowed to ignored:
```diff
1663,1664c1663
< 180B..180D    ; ignored                                # 3.0  MONGOLIAN FREE VARIATION SELECTOR ONE..MONGOLIAN FREE VARIATION SELECTOR THREE
< 180E          ; disallowed                             # 3.0  MONGOLIAN VOWEL SEPARATOR
---
> 180B..180E    ; ignored                                # 3.0  MONGOLIAN FREE VARIATION SELECTOR ONE..MONGOLIAN VOWEL SEPARATOR
2407,2408c2406
< 2060          ; ignored                                # 3.2  WORD JOINER
< 2061..2063    ; disallowed                             # 3.2  FUNCTION APPLICATION..INVISIBLE SEPARATOR
---
> 2060..2063    ; ignored                                # 3.2  WORD JOINER..INVISIBLE SEPARATOR
2412c2410
< 206A..206F    ; disallowed                             # 1.1  INHIBIT SYMMETRIC SWAPPING..NOMINAL DIGIT SHAPES
---
> 206A..206F    ; ignored                                # 1.1  INHIBIT SYMMETRIC SWAPPING..NOMINAL DIGIT SHAPES
6839c6837
< 1D173..1D17A  ; disallowed                             # 3.1  MUSICAL SYMBOL BEGIN BEAM..MUSICAL SYMBOL END PHRASE
---
> 1D173..1D17A  ; ignored                                # 3.1  MUSICAL SYMBOL BEGIN BEAM..MUSICAL SYMBOL END PHRASE
```

Concerns?

markusicu:
From Ken about my latest data file:
> I re-ran the derivation of Idna2008.txt over that new file and have verified that the output is identical as before, except for the timestamp on the file. So I think we are good, and Idna2008-16.0.0.txt does not need updating.

asmusf:
> I re-ran the derivation of Idna2008.txt over that new file and have verified that the output is identical as before, except for the timestamp on the file. So I think we are good, and Idna2008-16.0.0.txt does not need updating.

That is good, because changing any characters in Idna2008-16.0.0.txt that are not newly assigned would have been very troubling.

markusicu:
Experimentation done, UTC action item in hand, cleaned up now-obsolete code that used to compute the base exclusion set -- in favor of hardcoding the minimal set agreed in PAG & UTC. Please review.

markusicu:
PS: It no longer helps to review one commit at a time. We decided what to do, and I intend to squash-and-merge.

markusicu:
Test failure in TestIdnaTest.testBackwardsCompatibility, not totally surprising.
Hold your horses, I will fix that.

markusicu:
All tests pass. I think it works as advertised and approved. Someone please review.

nedley (approved):

markusicu:
Thanks Ned!

PR #729, opened 2024-03-06
Use the Last Resort font for Co, Cn, and most Cc

### LINKED ISSUE #725: jsp: E000 in characters.jsp shows as FFFD
https://util.unicode.org/UnicodeJsps/character.jsp?a=E000
PUA, but the glyph shown and copied is FFFD

eggrobin:
Sadly not Cs, even though it has really pretty glyphs for those, since that is not interchangeable, so that I cannot get them to show up.

Fix #725.


macchiati:
We would probably want to use this (if it works) for the default-ignorable characters, or at least the Cf characters. But for those, I would change the text color so people don't think that the characters have those glyphs for real.

eggrobin:
> We would probably want to use this (if it works) for the default-ignorable characters, or at least the Cf characters.

Sadly the rendering engine got in the way when I experimented with that, and they mostly got a glyph saying that they are ASCII (probably because they are playing with a preceding ASCII character, or something like that).

macchiati (approved):

PR #730, opened 2024-03-07
Ensure we have 16.0 before serving

### LINKED ISSUE #728: The JSPs are flaky
Since I pushed the new JSPs incorporating #432 and #502, they have tended to 500 on list-unicodeset.

(Originally filed as https://github.com/unicode-org/properties/issues/258.)

eggrobin:
Fix #728. Pushed this to prod (staging did not have the issue at all), it seems to have fixed things:
image


markusicu (dismissed):

eggrobin:
I had screwed up initially and the JSPs had thus rewedged themselves ; but I think with this third commit it should be fine ; I pushed that on Friday and there has been no such 500ing over the weekend.

markusicu (approved):

PR #732, opened 2024-03-10
Tolong Siki

josh-hadley:
See UTC-174-C22: _Provisionally assign 54 code points U+11DB0..U+11DE9 in a new Tolong Siki block at U+11DB0..U+11DEF, for 54 Tolong Siki characters as described in [L2/23-024](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/23-024) and section 6 of [L2/23-012](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/23-012)._

https://github.com/unicode-org/sah/issues/225
https://github.com/unicode-org/utc-release-management/issues/21

[UTC-181-C61] Consensus: UTC accepts for encoding in Unicode 17.0 the following 228 characters for new scripts for which code points have previously been provisionally assigned: 50 characters for Beria Erfe script, 40 characters for Chisoi script, 29 characters for Sidetic script, 55 characters for Tai Yo script, and 54 characters for Tolong Siki script.


unicodetools/data/ucd/dev/UnicodeData.txt
```diff
@@ -21870,6 +21870,60 @@ FFFD;REPLACEMENT CHARACTER;So;0;ON;;;;;N;;;;;
...
+11DD0;TOLONG SIKI LETTER R;Lo;0;L;;;;;N;;;;;
+11DD1;TOLONG SIKI LETTER L;Lo;0;L;;;;;N;;;;;
+11DD2;TOLONG SIKI LETTER V;Lo;0;L;;;;;N;;;;;
+11DD3;TOLONG SIKI LETTER NNY;Lo;0;L;;;;;N;;;;;
+11DD4;TOLONG SIKI LETTER S;Lo;0;L;;;;;N;;;;;
+11DD5;TOLONG SIKI LETTER H;Lo;0;L;;;;;N;;;;;
+11DD6;TOLONG SIKI LETTER X;Lo;0;L;;;;;N;;;;;
+11DD7;TOLONG SIKI LETTER RR;Lo;0;L;;;;;N;;;;;
+11DD8;TOLONG SIKI LETTER RRH;Lo;0;L;;;;;N;;;;;
+11DD9;TOLONG SIKI SIGN SELA;Lo;0;L;;;;;N;;;;;
```
eggrobin, line 21914: This is a vowel length mark, see p. 5 and the chart on p. 13 of the proposal L2/23-024.

I think this should be gc=Lm, and in PropList.txt, should have the Diacritic and Extender properties (like U+02D0).

The proposal bundles U+11DDA with it. For that noe I don’t know whether it is more appropriate to treat it as a letter that is a glottal stop or as modifier that does mysterious syllable coda things. @Ken-Whistler, what do you think?

Ken-Whistler: I don't think it (hecakā) is a vowel length mark, as contrasted with selā, which is. The glottal stop seems to be a full segment, although perhaps in restricted distribution. See the discussion of Kurukh and Brahui phonology on Wikipedia, for instance, which shows length distinctions (and in some cases nasalization distinctions) for vowels, which when marked in the orthography can be considered vowel modifiers and/or extenders, etc. But I think it is safest to just leave the glottal stop as Lo and treat it as another letter.

josh-hadley: Does this change also require changes to LineBreak.txt?

markusicu (approved):

PR #748, opened 2024-03-22
U+0300 and U+0308 from L2/23-280

eggrobin:
And fix the bad tests. Also add some tests to cover the combinations no longer covered by the combining marks.

See https://github.com/unicode-org/properties/issues/256.

[179-C] Consensus: Add all 53 entries to ScriptExtensions.txt as proposed in document L2/23-280, including U+0300 and U+0308, for Unicode Version 16.0. See document L2/24-064 item 2.3. This decision supersedes consensus UTC-178-C39.


unicodetools/src/test/java/org/unicode/test/TestSecurity.java
```diff
@@ -453,15 +453,17 @@ public void TestScriptDetection() {
         Set<Set<Script_Values>> expected = new HashSet<>();
         String[][] tests = {
             {"℮", "Common"},
+            {"1ℓ ℮", "Common"},
```
roozbehp, line 456: This appears unrelated. Move to another pull request?

unicodetools/src/test/java/org/unicode/test/TestSecurity.java
```diff
@@ -453,15 +453,17 @@ public void TestScriptDetection() {
...
+            {"75 cl ℮", "Latin"},
```
roozbehp, line 457: Same here

markusicu, line 457: I got email notifications for comments on these from Roozbeh but don't see the comments here...??

Anyway, I don't mind adding a couple of new test strings.

unicodetools/src/test/java/org/unicode/test/TestSecurity.java
```diff
@@ -524,9 +526,9 @@ public void TestWholeScripts() {
             {"⼒", Status.SAME}, // KANGXI RADICAL POWER
             {"力", Status.SAME}, // CJK UNIFIED IDEOGRAPH-529B
             {"!", Status.SAME, Status.OTHER},
-            {"\u0300", Status.SAME},
+            {"\u0300", Status.COMMON},
             {"a\u0300", Status.SAME, Status.COMMON, Status.OTHER},
-            {"ä", Status.SAME, Status.COMMON, Status.OTHER},
+            {"ä", Status.SAME, Status.OTHER},
```
markusicu, line 531: I haven't looked at what these expected values are, but (a) I assume these do naturally fall out from the data changes and (b) I believe that whole-script confusables haven been put on ice until someone can figure out real use cases, so this probably does not matter a lot.

FYI @macchiati

markusicu (commented):
lgtm

unicodetools/src/test/java/org/unicode/test/TestSecurity.java
eggrobin: As mentioned in the PR description, the point is that they cover cases that were previously covered by the test cases with U+0300 (A pile of Common characters is Common, and a pile of common characters with some Latin is Latin).

roozbehp: > I got email notifications for comments on these from Roozbeh but don't see the comments here...??

I deleted them. I saw @eggrobin's comment and understood the nature of the additional tests.

markusicu (approved):

PR #749, opened 2024-03-26
String property fixes

eggrobin:
Fix versioned property (which was broken because isTrivial was broken) and getSet when using incremental properties (which we do on the JSPs).

### DIFF
--- a/unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java
+++ b/unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java
@@ -804,6 +804,14 @@ public class IndexUnicodeProperties extends UnicodeProperty.Factory {
                 System.out.println(
                         "Long getSet for U" + ucdVersion + ":" + prop + " (" + Δt_in_ms + " ms)");
             }
+            // We only do the delta thing for code points; for strings, we need to do the lookup
+            // directly (and clean whatever was added by walking through history).
+            if (baseVersionProperties != null
+                    && (result.hasStrings()
+                            || (_getRawUnicodeMap().stringKeys() != null
+                                    && !_getRawUnicodeMap().stringKeys().isEmpty()))) {
+                result.removeAllStrings().addAll(super.getSet(matcher, new UnicodeSet()).strings());
+            }
             return result;
         }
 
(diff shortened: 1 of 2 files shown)


unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java
```diff
@@ -722,9 +722,11 @@ class IndexUnicodeProperty extends UnicodeProperty.BaseProperty {
         @Override
         public boolean isTrivial() {
             return _getRawUnicodeMap().isEmpty()
-                    || _getRawUnicodeMap()
-                            .keySet(_getRawUnicodeMap().getValue(0))
-                            .equals(UnicodeSet.ALL_CODE_POINTS);
+                    || ((_getRawUnicodeMap().stringKeys() == null
+                                    || _getRawUnicodeMap().stringKeys().isEmpty())
```
markusicu, line 726: FYI: a UnicodeMap.hasStrings() function would be nice...

unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
```diff
@@ -884,6 +884,17 @@ Let $HairComponents = [\U0001F9B0-\U0001F9B3]
...
 [\p{Emoji_Presentation} - \p{RI}] ⊂ \p{ea=W}
 
+\p{RGI_Emoji_Zwj_Sequence} ⊃ []
+\p{RGI_Emoji_Zwj_Sequence} ⊇ \p{U-1:RGI_Emoji_Zwj_Sequence}
+# Unicode Version 15.1 had no new emoji code points, but had ZWJ sequences.
+\p{U15.1:RGI_Emoji_Zwj_Sequence} ⊃ \p{U15:RGI_Emoji_Zwj_Sequence}
+\p{U15.1:Emoji} = \p{U15:Emoji}
+[\p{U15.1:RGI_Emoji_Zwj_Sequence} - \p{U15:RGI_Emoji_Zwj_Sequence}] ⊃ [{🐦‍🔥}]
+# Likewise Emoji Version 13.1 (which was not actually a Unicode version, so we
+# cannot look at U13.1:Emoji.
```
markusicu, line 894: missing paren:
```suggestion
# Likewise Emoji Version 13.1
# (which was not actually a Unicode version, so we cannot look at U13.1:Emoji).
```

unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java
eggrobin: Yes.

markusicu (approved):

PR #752, opened 2024-03-29
make 113CF TULU-TIGALARI SIGN LOOPED VIRAMA a tertiary variant of 113…

markusicu:
…CE TULU-TIGALARI SIGN VIRAMA

For https://github.com/unicode-org/properties/issues/192 "UCA 16: virama variants primary vs. tertiary"

Also, I fixed the UCA tool again: It failed with the group of scripts from Sinhala to before Thai overflowing a single compressible lead byte. I don't know why, since we have not added characters to that group since last November, but I didn't try to investigate this at length. I randomly started a new lead byte for Newa.

From Ken:

UCA 16.0 delta 16

I've made the single point fix to DUCET to make 113CF TULU-TIGALARI SIGN LOOPED VIRAMA a tertiary variant of 113CE TULU-TIGALARI SIGN VIRAMA.

The bad news is that I went looking for consistency in the other virama characters in other scripts, and it turns out that will be very hard to make completely consistent if the criterion is based on same or different functions in IndicSyllabicCategory.txt.

Problem areas for Robin to muse over:

Malayalam: 0D4D is a virama, 0D3B, 0D3C are pure_killer. Right now 0D3B and 0D3C are tertiary variants of 0D4D, which would not be consistent if we are trying to give separate primary weights to viramas versus pure_killer.

Tagalog has two pure killers: 1714, 1715. Currently they have separate primary weights, which is inconsistent with the principle, but it seems wrong to treat these two as presentation variants of each other.

Batak also has two pure killers: 1BF2, 1BF3. I can't figure out from the documentation whether these are presentation variants of each other, or not.

Thai has two pure killers: 0E3A, 0E4E. The second is yamakkan, but the corresponding Lao yamakkan is treated as a syllable_modifier, not a pure_killer, so I'm not sure which way to go on that one.

The Kirat Rai virama and saat sound like they should be separate, but both are classified as pure_killer, so it isn't clear what the right answer is there, either.

Basically, I think looking for complete consistency here is going to hurt everybody's brains, with minimal ROI. I suggest we just sweep the rest under the rug, take the Tulu-Tigalari case as a win, and go home.


Ken-Whistler (approved):
Input change looks correct. No comment on the change required in the tooling for first byte ranges.

PR #761, opened 2024-04-02
Fix a bad regex

eggrobin:
Follow-up on https://github.com/unicode-org/unicodetools/pull/751, as foretold in the tools meeting today 🙃


josh-hadley (approved):
😸

PR #776, opened 2024-04-12
minor terms of use update from Anne

markusicu:
- Minor changes to "terms of use" language in Java code that generates data, and in readme template files.
- Regenerated UCD files.
- Manually updated other UCD files.

*Best to review one commit at a time.*

The generated files (readmes, and data files outside of the UCD) that result from these programs will be updated in future pull requests as we regenerate and publish data for Unicode 16 beta.


unicodetools/data/security/dev/data/ReadMe.txt
markusicu, line 1: FYI -- This file does not get posted on the web site.

unicodetools/data/ucdxml/dev/ucdxml.readme.txt
markusicu, line 1: FYI -- I changed this UCDXML readme into a template similar to the other readme templates.
When we turn the crank for 16 beta, I will need to work with John on adding this file to the publication scripts that handle the readme templates. I would also like to remove the hardcoded file sizes, to avoid manual updates here. TBD

unicodetools/src/main/java/org/unicode/tools/emoji/GenerateEmoji.java
```diff
@@ -1533,9 +1533,9 @@ private static void showOrdering(
 
             outPlain.println(
                     "# labels.txt\n"
-                            + "# Copyright © 1991-2016 Unicode, Inc.\n"
+                            + "# Copyright © 1991-Present Unicode, Inc.\n"
```
markusicu, line 1536: FYI -- I don't know which files this generates; I am surprised that this was stuck on 2016, and I am not aware of where such files might live. @macchiati @nedley might know.

nedley: labels.txt is generated in the `internal` charts directory, presumably for ingestion by CLDR. I’m honestly not sure why it needs to state a copyright when another file used by CLDR (emoji-ordering-rules.txt) doesn’t have one.

unicodetools/src/main/java/org/unicode/tools/Segmenter.java
```diff
@@ -1311,8 +1311,8 @@ public static void main(String[] args) throws IOException {
                             + "<!DOCTYPE ldml SYSTEM \"../../common/dtd/ldml.dtd\">\n"
                             + "<!--\n"
                             + "Copyright © 1991-2015 Unicode, Inc.\n"
```
srl295, line 1313: ^ have a 2015 here

unicodetools/src/main/java/org/unicode/tools/emoji/GenerateEmoji.java
srl295: why `Present`? was that @annebright approved? 

`CldrUtility.getCopyrightShort()` will return

```java
"Copyright \u00A9 1991-"
                        + Calendar.getInstance().get(Calendar.YEAR)
                        + " Unicode, Inc.";
```

srl295: and actually, `"<!--\n"+getCopyrightString()+"\n-->\n"` from that same class is what CLDR uses to output its whole file header

markusicu: As @nedley wrote, this one doesn't really matter, because it just generates internal files.

markusicu: changed to use `Default.getYear()` like elsewhere in these tools

unicodetools/src/main/java/org/unicode/tools/Segmenter.java
markusicu: changed to use `Default.getYear()` like elsewhere in these tools

srl295 (approved):
I'd use the CLDR utillity, but ok..

srl295 (dismissed):

eggrobin:
Still conflicting apparently?

eggrobin (approved):

markusicu:
Sorry, conflict resolution not done yet, and running into trouble with that...

eggrobin:
> Sorry, conflict resolution not done yet, and running into trouble with that...

I ran https://github.com/unicode-org/unicodetools/blob/main/docs/pipeline.md#merge on this branch and it Just Worked, so I took the liberty of pushing it in case you are stuck. Have a good International Workers’ Day!

eggrobin (approved):
Rubberstamping my own merge commit, but I won’t be the one pressing the squash-and-merge button so it’s OK.

markusicu:
I guess I force-pushed my version at about the same time...

PR #780, opened 2024-04-23
Armenian modifiers

eggrobin:
See L2/24-107, Proposal to encode three Armenian superscript characters, and https://github.com/unicode-org/sah/issues/432.


eggrobin:
The invariant that annoyed me in https://github.com/unicode-org/unicodetools/pull/743#issuecomment-2009829136 is annoying me again. I need to do something about it…

markusicu:
PAG says looks fine

markusicu (approved):

markusicu (approved):

PR #799, opened 2024-05-02
Correct the section number for Tangut in TangutSources.txt

eggrobin:
See email from @dwanders-A.


markusicu (approved):
indeed: [18.11 Tangut](http://www.unicode.org/versions/Unicode15.0.0/ch18.pdf#G43635)

PR #798, opened 2024-05-02
Unicode 16 line breaking changes

eggrobin:
[[179-A103](https://www.unicode.org/cgi-bin/GetL2Ref.pl?179-A103)] Action Item for Robin Leroy, PAG: In LineBreakTest.txt, Change rule LB19 and add LB19a as described in [L2/24-064](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/24-064) item 5.5, and add realistic handwritten test cases. For Unicode Version 16.0. See document [L2/24-064](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/24-064) item 5.5.

[[179-A112](https://www.unicode.org/cgi-bin/GetL2Ref.pl?179-A112)] Action Item for Robin Leroy, PAG: In LineBreakTest.txt, add rule LB20a as described in document [L2/24-064](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/24-064) item 5.13, and add realistic handwritten test cases. For Unicode Version 16.0. See document [L2/24-064](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/24-064) item 5.13.

[[179-A117](https://www.unicode.org/cgi-bin/GetL2Ref.pl?179-A117)] Action Item for Robin Leroy, PAG: In LineBreakTest.txt, Change rules LB13 and LB25, and add rules LB15c and LB15d, and make changes to the regular expression for numbers and to the examples, as described in [L2/24-064](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/24-064) item 5.15. Add test cases that exercise the rule changes. For Unicode Version 16.0. See document [L2/24-064](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/24-064) item 5.15.

Note that I am _not_ updating SegmenterCldr.txt, because we want to burn it down anyway. See discussion in https://github.com/unicode-org/unicodetools/issues/492#issuecomment-2023947121.

PR #801, opened 2024-05-03
More Unicode 16 linebreaking changes

eggrobin:
I had missed this one in #798 because we were missing a LineBreakTest.txt action item for it:

[[179-C25](https://www.unicode.org/cgi-bin/GetL2Ref.pl?179-C25)] Consensus: In UAX #14, change rule LB21a from HL (HY | BA) × to HL (HY | BA) × [^HL] (do not break after the hyphen in Hebrew-hyphen-non-Hebrew) to match the rationale for adding LB21a given in [L2/11-141R](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/11-141R). For Unicode Version 16.0. See document [L2/24-064](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/24-064) item 5.2.


markusicu (approved):

eggrobin:
Note: running the ICU rbbi/RBBITest/TestUnicodeFiles on the current unicodetools LineBreakTest.txt (patched to replace U+1B50 with U+1BC0 since ICU does not yet have its LB change UTC-177-C46) on my uax14-integration branch failed with
```
         Test file "LineBreakTest.txt", line 16664, unexpected break found at position 2
         Expected : ×א×-÷א÷
         Actual   : ×א÷-÷א÷
```
So our conformance tests would have caught that blunder during ICU integration.

With the LineBreakTest.txt from this branch (similarly patched) the ICU test now passes.

Note that this all passes after removing the exemptions for ICU-21097, so this one can be closed soon (with the resolution that Andy had predicted four years ago).

PR #802, opened 2024-05-03
Unihan 16 for UTC-179

markusicu:
@kenlunde prepared this version of the data for UTC-179, and in that meeting, all of these changes were approved.

First approver wins!


kenlunde (approved):

Manishearth (approved):

markusicu:
Sorry, there was a test failure. @eggrobin had added an "invariant" test with one exception, submitted feedback about that exception, and this version of the data fixes the data value. So I had to remove the exception from the test to make it work again.

eggrobin (approved):

PR #811, opened 2024-05-08
Remove a redundant rule

eggrobin:
This rule, approved by UTC-179-C35, does nothing, because LB13 already did whatever it would do.

I will be putting a review note about that in the proposed update (see https://github.com/unicode-org/unicode-reports/pull/141), this obviously has no effect and will have been called out during review so I do not think it warrants a decision separate from the decision to approve UAX14 for final publication.

Also: add a missing space.


macchiati (approved):
In particular, it is redundant because of:

```
13.04) × $SY
```

PR #816, opened 2024-05-10
SpecialCasing context/locale only historical

markusicu:
[[176-A68](https://www.unicode.org/cgi-bin/GetL2Ref.pl?176-A68)] Action Item for Markus Scherer, Asmus Freytag, PAG: In the core spec, near Table 3-17 “Context Specification for Casing” and in chapter 5.18 “Case Mappings”, document that additional language-specific and orthography-specific case mapping behavior is or will be documented in CLDR; see also [L2/23-160](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/23-160) item 2.3; for Unicode 16.0.


Ken-Whistler (approved):
Changes look fine

roozbehp (approved):
LGTM too.

PR #824, opened 2024-05-17
UTS46 test data fixes for 175-A88

markusicu:
For https://github.com/unicode-org/properties/issues/162

which contains discussion for
[175-A88] Action Item for Mark Davis, PAG: Review IdnaTestV2.txt and ensure that the documentation and output of status codes conforms to the specification; see L2/23-079 item 6.3 for a report of mismatches; for Unicode Version 15.1.

Review one commit at a time.

Changes, in the order of the bug report items:
- P4: fixed header comment
- in-browser-test-friendly: avoid the last label consisting of only ASCII digits (second commit)
- U1 not V7 if disallowed_STD3_* (third commit)
- NV8: Updated header comment, copied from issue with minor changes
- A3: Added a test case that triggers it (unpaired surrogate), and a header comment that it can be filtered out
- X3:
  - The implementation "knows about it" but it's unreachable code
  - Added a code comment
  - Updated header comment, copied from issue but simplified it
- "CheckBidi is not V8" was already fixed in an earlier PR


josh-hadley (approved):

markusicu:
Thanks Josh! @macchiati ok to merge?

macchiati (approved):

PR #828, opened 2024-05-22
update: spotless 2.43 & google-java-style 1.22

markusicu:
See the version numbers in the root pom.xml file.
The old versions no longer worked with a new JDK on my machine.


echeran (approved):
LGTM

PR #830, opened 2024-05-23
idnatest Punycode with uppercase ASCII

### LINKED ISSUE #747: IdnaTestV2.txt should test for valid input with upper case in the ASCII part of Punycode
IdnaTestV2.txt missed a bug in the UTS 46 implementation that I'm writing due to not testing an upper-case letter in the ASCII part of Punycode when there are no errors.

I suggest adding this test case:

Input: `xn--A-1ga`
Expected ToASCII: `xn--a-1ga`
Expected ToUnicode: `aö`

markusicu:
Fixes #747

@hsivonen wrote:

IdnaTestV2.txt missed a bug in the UTS 46 implementation that I'm writing due to not testing an upper-case letter in the ASCII part of Punycode when there are no errors.

I suggest adding this test case.


echeran (approved):
LGTM

PR #831, opened 2024-05-23
Make the invariant related Cased to Case_Ignorable more maintainable

eggrobin:
See https://github.com/unicode-org/unicodetools/pull/780#issuecomment-2072335732 and https://github.com/unicode-org/unicodetools/pull/743#issuecomment-2009829136.


markusicu (approved):

macchiati (commented):
Much better

PR #835, opened 2024-05-24
Do not use ICU property values in the JSP TestProperties

eggrobin:
Follow-up on https://github.com/unicode-org/unicodetools/pull/810#issuecomment-2101660806.


UnicodeJsps/src/test/java/org/unicode/jsptest/TestProperties.java
```diff
@@ -294,7 +296,10 @@ public void TestCCC() {
         for (int i = 0; i < 256; ++i) {
             String s = String.valueOf(i);
             if (s.contains("3")) {
-                expected.addAll(new UnicodeSet("[:ccc=" + s + ":]"));
+                expected.addAll(
+                        ToolUnicodePropertySource.make(Settings.latestVersion)
```
markusicu, line 300: Please move this line out of the loop, keeping the TUPS in a variable. I understand that these are cached, but it will make it a little more readable too, I think.

PR #836, opened 2024-05-24
Hani CJK strokes

eggrobin:
For https://www.unicode.org/review/pri502/feedback.html#ID20240523095709 (https://github.com/unicode-org/properties/issues/296).


roozbehp (commented):
LGTM

PR #839, opened 2024-05-30
Test consistency of grapheme cluster segmentation with canonical equivalence some more

eggrobin:
… and fix it.

UTC-180-CXXX Consensus: Assign the Other_Grapheme_Extend property to U+0CC0, U+0CC7, U+0CC8, and U+0CCA, U+0CCB (Kannada vowel signs II, EE, AI, O, and OO), as well as U+1B3B, U+1B3D, and U+1B43 (Balinese vowel signs ra repa tedung, la lenga tedung, and pepet tedung), thereby changing their Grapheme_Cluster_Break property from SpacingMark to Extend, ensuring consistency of legacy grapheme clusters with canonical equivalence. For Unicode 16.0. See L2/24-162 item 6.2, PRI-494#ID20240422114157, and PRI-494#ID20240422120150.

See https://github.com/unicode-org/properties/issues/287.

Previously: https://github.com/unicode-org/unicodetools/pull/645.

PR #834, opened 2024-05-24
Move the security invariants to their own CI check

eggrobin:
See discussion in PAG meeting 2024-05-23.

See it in action in https://github.com/unicode-org/unicodetools/pull/833/files#diff-ae46290feceee5ef92cbfa7106806971b53e015ab3f842e6d1bd9f816dbb9958.

PR #840, opened 2024-05-30
Fix some bugs in the splitting of the invariants

eggrobin:
Among other things, #834 broke the reporting of invariant test failures in the diff.

Cherry-picked from https://github.com/unicode-org/unicodetools/pull/839 which will take longer to merge because it has a property change.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
@@ -124,17 +124,12 @@ public class TestUnicodeInvariants {
     /**
      * Fetch a reader for our input data.
      *
-     * @param inputFile if null, read DEFAULT_FILE from classpath
+     * @param inputFile read from classpath
      * @return BufferedReader
      * @throws IOException
      */
     private static BufferedReader getInputReader(String inputFile) throws IOException {
-        if (inputFile != null) {
-            return FileUtilities.openUTF8Reader(Settings.SRC_UCD_DIR, inputFile);
-        }
-
-        // null: read it from resource data
-        return FileUtilities.openFile(TestUnicodeInvariants.class, DEFAULT_FILE);
+        return FileUtilities.openFile(TestUnicodeInvariants.class, inputFile);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -144,14 +139,22 @@ private static BufferedReader getInputReader(String inputFile) throws IOExceptio
...
         try (final PrintWriter out2 =
                 FileUtilities.openUTF8Writer(
                         Settings.Output.GEN_DIR,
-                        "UnicodeTestResults." + (doHtml ? "html" : "txt"))) {
+                        "UnicodeTestResults"
+                                + (inputFile.equals(DEFAULT_FILE)
+                                        ? ""
+                                        : "-" + inputFile.split("\\.")[0])
+                                + "."
+                                + (doHtml ? "html" : "txt"))) {
```
markusicu, line 157: optional
```suggestion
                        "UnicodeTestResults"
                                + (inputFile.equals(DEFAULT_FILE)
                                        ? ""
                                        : "-" + inputFile.split("\\.")[0])
                                + (doHtml ? ".html" : ".txt"))) {
```

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -144,14 +139,22 @@ private static BufferedReader getInputReader(String inputFile) throws IOExceptio
      * @throws IOException
      */
     public static int testInvariants(String inputFile, boolean doRange) throws IOException {
+        if (inputFile == null) {
+            inputFile = DEFAULT_FILE;
+        }
```
markusicu, line 144: optional:
Here you could set an extra variable with "" or "-security" so that you need not rediscover later which variant it is.

Or, even better,
- add a `variant` parameter to the testInvariants() signature
- change the call sites to pass it in
- consider having a non-empty suffix for the default tests as well

markusicu (commented):
Could you please also update https://github.com/unicode-org/unicodetools/blob/main/docs/build.md#invariant-checking with the split and the input/output file names?

markusicu (approved):
lgtm tnx

Issue #575, opened 2023-10-17
Feature Request: Codepoint property diff


iwsfutcmd:
Sometimes, it's really nice to be able to know how the properties of two codepoints differ. For example, if we at the SAH are trying to recommend using an existing codepoint for some novel usage, it'd be nice to easily compare two potential candidates by their properties.

It seems like it'd be a good addition to https://util.unicode.org/UnicodeJsps/character.jsp, although perhaps it'd be better as its own tool.

eggrobin:
This would also be useful for PAG review; I keep saying that something new is propertywise like some existing thing, but I end up checking that by staring at the files, which is error prone (especially if the new character mistakenly has a default value for some property).

I just wrote a quick and dirty tool hidden in a test, I should make it its own command-line tool for my use (I don’t want to have to build local JSPs to do the PAG review) and also surface a feature like that in the JSPs so others can use it for existing characters.

PR #842, opened 2024-05-31
Propertywise … AreAlike

eggrobin:
Towards #575. Does not fix it; while should address the use case for PAG review of proposals, only convenient for people who know how to wield the tools—and specifically the invariant tests—, and that is a very small number of people indeed. I should build something in a similar vein in the JSPs.

See a sample test failure in https://github.com/unicode-org/unicodetools/pull/842/files/3d97c4a4a9bc806c65dc70722e0d947dec826edd#diff-f86c50974502e26e45f2495660a5406f44604ae6ce4c49329e45a795bd520a9a.


unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -326,6 +332,83 @@ protected String getFailure(int codepoint) {
...
+        final var iup = IndexUnicodeProperties.make(Settings.latestVersion);
+        final List<String> errorMessageLines = new ArrayList<>();
+        properties:
+        for (var p : UcdProperty.values()) {
+            final var property = iup.getProperty(p);
+            for (String alias : property.getNameAliases()) {
+                if (excludedProperties.contains(alias)) {
+                    continue properties;
+                }
+            }
```
markusicu, line 367: Could you do something like
```
if (property.getNameAliases().stream().anyMatch(alias -> excludedProperties.contains(alias))) {
    continue;
}
```
and then, because you no longer have an inner loop, avoid the loop label?

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -326,6 +332,83 @@ protected String getFailure(int codepoint) {
...
+            final int first = set.charAt(0);
+            String p1 = property.getValue(first);
+            for (var range : set.ranges()) {
+                for (int c = range.codepoint; c <= range.codepointEnd; ++c) {
+                    String p2 = property.getValue(c);
```
markusicu, line 372: Aren't you comparing first with itself as well?
-->
```suggestion
                for (int c = range.codepoint; c <= range.codepointEnd; ++c) {
                    if (c == first) { continue; }
                    String p2 = property.getValue(c);
```

markusicu (approved):
lgtm tnx

PR #843, opened 2024-06-04
NamesList update for UTC-164-A42

eggrobin:
From @Ken-Whistler:
> I've done another generation of NamesList.txt, with the first of the
additions after we started beta review. I'm calling it a post-beta, and
anticipate there will end up being more updates needed as beta review
feedback comes in.
>
> This first change was for an old action item, 164-A42, which was to deal
with Rebecca's feedback in L2/20-158. SAH had taken a look at it, and at
the time, back in 2020, decided it was too much for the names list
annotations, and asked me to look at whether it made sense for text
additions to 22.7 in the core spec, or just as fodder for a UTN. After
thinking about this a bit, it turns out this really is special
information about 3 particular characters. Trying to provide enough
context in an explanation for Section 22.7 would just be more work. And
while a UTN on the entire set of these legacy terminal graphics sets
might make sense, certainly it doesn't make sense for just these 3
oddball characters.
>
> So I went back to the drawing board, and just put in a couple subheads
and then notices that could live with those subheads for the slightly
longer explanations Rebecca wanted, instead of annotations directly on
the characters. I think it works.

### DIFF
--- a/unicodetools/data/ucd/dev/NamesList.txt
+++ b/unicodetools/data/ucd/dev/NamesList.txt
@@ -1,23 +1,11 @@
 ; charset=UTF-8
 @@@	The Unicode Standard 16.0.0
 @@@+	NamesList-16.0.0.txt
-@+	Generation Date: 2024-05-17, 19:24:02 GMT
+@+	Generation Date: 2024-06-03, 12:53:56 GMT
 	Unicode 16.0.0 names list.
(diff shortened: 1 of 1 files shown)


Ken-Whistler (approved):
Looks correct

markusicu (approved):
rslgtm

PR #846, opened 2024-06-04
Modifier ψ and ω

eggrobin:
See https://github.com/unicode-org/sah/issues/485.


eggrobin:
@markusicu Note that we reviewed this with incorrect names (from the non-R version of the document).
Now corrected, and I updated the tests to use a similarly named character as a reference and to explicitly mention its name.
I don’t think there is much more I can reliably do to prevent that kind of error from my side, I added a note to the RMG agenda so that these updates don’t get dropped on the floor.

PR #851, opened 2024-06-05
Allow for UCA and security invariant test failures on repertoire PRs

eggrobin:
https://github.com/unicode-org/unicodetools/blob/main/docs/pipeline.md says 
> The Check UCA data CI check might fail; many character additions need separate handling there, but that is out of scope for the PAG work of preparing data-for-new. This will get resolved later.

But now the security check also _systematically_ fails; being an invariant test, it also brings up error messages in the diff. This can easily lead to missing real invariant check failures—which may point to potential issues with encoding proposals—, as we start expecting proposal PRs to be red.

In order to address that issue, this PR silences the UCA and security invariant failures on PR runs for changes to the répertoire (the build at main will still go red if we merge UCA or security breakages, which we may want to do before the start of the β phase). PRs that do not affect the répertoire will also get the full CI results.

When silenced, the security invariant test failures still show up in the diff, but as _notices_ (in grey) rather than errors (in red).

See it in action:
1. in [9bd67946d9ec4798ce5b02697eaab46608c8ba27](https://github.com/unicode-org/unicodetools/pull/851/commits/9bd67946d9ec4798ce5b02697eaab46608c8ba27), the only two failing CI checks are due to UCD invariant check failures, and the security invariant test failure is reported as a notice rather than an error;
2. in [2c1c296ed470362cc2cc13ef714cb799d521113e](https://github.com/unicode-org/unicodetools/pull/851/commits/2c1c296ed470362cc2cc13ef714cb799d521113e), there are no failing CI checks (all is green), and the security invariant test failure is reported as a notice rather than an error;
3. in [d8f557dcb497f0141223a7ebada2b3a088d42426](https://github.com/unicode-org/unicodetools/pull/851/commits/d8f557dcb497f0141223a7ebada2b3a088d42426), there are no répertoire changes, so the CI checks run normally: both the UCA and the security invariants are broken.


macchiati (approved):
makes sense

PR #855, opened 2024-06-06
Parse Unikemet

eggrobin:


unicodetools/src/main/resources/org/unicode/props/ExtraPropertyValueAliases.txt
```diff
@@ -85,6 +85,10 @@
 # @missing: 0000..10FFFF; Emoji_Component  	; No
 # @missing: 0000..10FFFF; Extended_Pictographic  	; No
 
+# @missing: 0000..10FFFF; kEH_Core  ;  No
+# @missing: 0000..10FFFF; kEH_NoMirror  ;  No
+# @missing: 0000..10FFFF; kEH_NoRotate  ;  No
```
markusicu, line 90: Do we need `@missing` lines for binary properties? They must default to "No" because we don't currently have syntax to have a "No" override.

unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
```diff
@@ -1090,6 +1090,13 @@ Let $japaneseSimplifiedRadicals = \p{Name=/CJK RADICAL J-SIMPLIFIED/}
 In $chineseSimplifiedRadicals, Equivalent_Unified_Ideograph ∈ [\p{kRSUnicode=/^[0-9]+'\.0$/} $radicalsWithUnifiableSimplifications]
 In $japaneseSimplifiedRadicals, Equivalent_Unified_Ideograph ∈ \p{kRSUnicode=/^[0-9]+''\.0$/}
 
+# Egyptian hieroglyph invariants
+
+Let $unikemetScope = [\p{Block=/^Egyptian.Hieroglyphs/} - \p{gc=Cn}]
+$unikemetScope = [ [\p{gc=Lo} & \p{sc=Egyp}] - \p{Name=/^EGYPTIAN HIEROGLYPH (FULL |HALF |TALL |WIDE )?(BLANK|LOST SIGN)$/} ]
```
markusicu, line 1096: just FYI: I always stumble a little over "=" meaning "assign" on a "Let" line but "assertEquals" on a test line (without a statement prefix) ... It would be a disruptive change, but using "==" for "assertEquals" might make this file more obvious.

unicodetools/src/main/resources/org/unicode/props/ExtraPropertyValueAliases.txt
eggrobin: For now we need them, though this annoys me:
https://github.com/unicode-org/unicodetools/blob/5aa08acba04b6e5794ead18e03c355ae1417670a/unicodetools/src/main/resources/org/unicode/props/ExtraPropertyValueAliases.txt#L6

unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
eggrobin: Or := (or ≔?) in `Let` (that would keep the actual invariant statements closer to normal mathematical operators on sets).

Not today, but good idea.

markusicu: Let with ASCII := would be ok

markusicu (approved):
lgtm but we need to remember that if UTC-180 disagrees with the PA/PVA additions, then we will need to move these back to the Extra files

markusicu:
CI failure:
```
Error:  org.unicode.propstest.TestProperties.TestDefaults  Time elapsed: 2.912 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: Null kEH_NoMirror for U+FFFF but property is Binary with default value type LITERAL. Add an @missing line to ExtraPropertyValueAliases.txt if needed. ==> expected: not <null>
	at org.unicode.propstest.TestProperties.TestDefaults(TestProperties.java:614)
```

eggrobin:
> CI failure:

Ah yes, I went too far in my removal in Extra, and there’s an illustration of the answer on needing those lines for the binaries 😩

eggrobin:
> lgtm but we need to remember that if [UTC-180](https://www.unicode.org/cgi-bin/GetL2Ref.pl?180) disagrees with the PA/PVA additions, then we will need to move these back to the Extra files

Or if UTC does not take the decision recommended by https://github.com/unicode-org/sah/issues/496 (I kept Core in Extra in anticipation of that).

markusicu (approved):

PR #858, opened 2024-06-06
Make emoji-ordering-rules.txt an internal data file

### LINKED ISSUE #853: Simpler/robuster update of CLDR emoji (emoji-ordering-rules.txt)
https://github.com/unicode-org/cldr/blob/main/docs/dev/generate-emoji-paths.md describes the files we need for each CLDR release when emoji data changes. We should make this process as simple as possible on both ends, both to speed the process and make it less error-prone.

One of the files, emoji-ordering-rules.txt, is not generated and checked into the unicodetools repo. It would make the process of the ESR updating emoji for CLDR simpler (and less error prone) if that file were also in the repository at unicodetools/data/emoji/dev/internal, like emoji-proposals.txt, etc. are.

The file is current generated during the chart generation, and is at the top level of charts-XX, whereever they are generated.I think it would take a small fix to the chart generation in order to redirect the generation of that file to the repo location.

nedley:
Fix #853


markusicu:
CI failure --> please `mvn spotless:apply`

Will there be changes on the CLDR side as well? Updating the docs, removing this file from there, ...?

nedley:
> Will there be changes on the CLDR side as well? Updating the docs, removing this file from there, ...?

@macchiati will have to answer, I am doing his bidding here.

PR #859, opened 2024-06-06
Ucdxml preview

jowilco:
PR to make it easy to see what changes have been made to support UCDXML.
NOT INTENDED FOR MERGE.


jowilco:
Comment on June 6 is no longer valid - we're now ready for review.

jowilco:
@macchiati @eggrobin @markusicu - Please can you review?

unicodetools/src/main/resources/org/unicode/props/IndexUnicodeProperties.txt
```diff
@@ -310,6 +313,15 @@ Unihan_Variants ; kSpoofingVariant
 Unihan_Variants ; kTraditionalVariant
 Unihan_Variants ; kZVariant
```
eggrobin, line 314: Should there be a line for kZhuang here? (In other words, are you getting any data for kZhuang?)

unicodetools/src/main/resources/org/unicode/props/ExtraPropertyAliases.txt
```diff
@@ -150,6 +151,13 @@ cjkJoyoKanji ; kJoyoKanji
 cjkKoreanEducationHanja ; kKoreanEducationHanja
 cjkKoreanName ; kKoreanName
 cjkTGH ; kTGH
+cjkRSTUnicode ; kRSTUnicode
+cjkReading ; kReading
+cjkSrc_NushuDuben ; kSrc_NushuDuben
+cjkTGT_MergedSrc ; kTGT_MergedSrc
```
eggrobin, line 157: Please revert this change: Tangut and Nüshu are not CJK, they should not have a cjk alias.

The name kReading is unfortunate (since this is really Nüshu-specific), but it is what it is.
I guess you should add the comment that I should have added saying that these are the fields from the Tangut and Nüshu sources files.

unicodetools/src/main/resources/org/unicode/props/ExtraPropertyAliases.txt
eggrobin, line 1: Thanks for catching those stragglers. I had been trying to make this file more comprehensive, but apparently I still missed some.

unicodetools/src/main/resources/org/unicode/props/ExtraPropertyAliases.txt
jowilco: Fixed.

unicodetools/src/main/java/org/unicode/xml/AttributeResolver.java
```diff
@@ -0,0 +1,458 @@
...
+                            extensionBuilder.append(
+                                    UcdPropertyValues.Script_Values.valueOf(extension)
+                                            .getShortName());
+                            extensionBuilder.append(" ");
+                        }
+                        return extensionBuilder.toString().trim();
+                    default:
+                        throw new RuntimeException("Missing Catalog case");
+                }
+            case Enumerated:
```
eggrobin, line 296: This (and the associated pile of `UnicodeMap`s) seems like it is going to be a bit annoying to maintain as we add properties.
Is there a reason why you are not doing something like
```java
final UnicodeProperty property = indexUnicodeProperties.getProperty(prop);
final List<String> valueAliases = property.getValueAliases(property.getValue(codepoint));
return valueAliases.size() == 1 ? valueAliases.get(0) : valueAliases.get(1);
```
for most of them (special-casing Decomposition_Type etc. as needed)?

macchiati:
No good reason. I think the code might predate the formal recognition as
properties.

uax/uax42/Readme.md
markusicu, line 1: Documentation should go into https://github.com/unicode-org/unicodetools/tree/main/docs,
one of
- some existing file that covers UCDXML (not sure if there is one)
- a new ucdxml.md there
- a new index.md in a new ucdxml/ folder there

uax/uax42/fragments/block/block.xml
markusicu, line 1: The Unicode Tools normally work with internal data files stored in https://github.com/unicode-org/unicodetools/tree/main/unicodetools/src/main/resources/org/unicode

Is it necessary to create a new, separate folder structure outside of that?
Why not add a ucdxml/ folder in the usual place?

uax/uax42/fragments/boolean/boolean.xml
markusicu, line 1: There are many little folders with one file each, and one folder with three. Is that necessary or useful? Can we flatten these all into one folder?

Some folders like "properties" are ok, but adding lots of mini folders seems cumbersome.

uax/uax42/fragments/datatypes/code points.xml
markusicu, line 1: Please no spaces in file names

uax/uax42/Readme.md
```diff
@@ -0,0 +1,24 @@
...
+- In UNICODETOOLS_REPO_DIR/uax/uax42/ run `mvn xml:transform`
+
+  index.html and index.rnc will be generated in UNICODETOOLS_REPO_DIR/uax/uax42/output/
+
+## Step 3 - Validate generated UAX XML files
+
+You'll need a [RELAX NG](https://relaxng.org/) schema validator. We'll use [jing-trang](https://github.
+com/relaxng/jing-trang) in this example.
+
+1. Clone and build [jing-trang](https://github.com/relaxng/jing-trang)
```
markusicu, line 18: This does not look workable. Dependencies should be handled by Maven via the pom.xml file which I assume would pull the Trang artifacts from https://mvnrepository.com/artifact/org.relaxng/trang

It should be possible to run one Maven command that builds and runs the tool.

uax/uax42/Readme.md
```diff
@@ -0,0 +1,24 @@
...
+## Step 2 - Generate TR42 index.html and index.rnc 
+
+- In UNICODETOOLS_REPO_DIR/uax/uax42/ run `mvn xml:transform`
+
+  index.html and index.rnc will be generated in UNICODETOOLS_REPO_DIR/uax/uax42/output/
+
+## Step 3 - Validate generated UAX XML files
+
+You'll need a [RELAX NG](https://relaxng.org/) schema validator. We'll use [jing-trang](https://github.
+com/relaxng/jing-trang) in this example.
```
markusicu, line 16: Is this the recommended tool for the job? There has been no release and no commit in 2.5 years. Abandoned?

uax/uax42/Readme.md
```diff
@@ -0,0 +1,24 @@
+# Generating TR42
+
+## Step 1 - Generate property value fragments
+
+- Run org.unicode.xml.GeneratePropertyValues to populate the UNICODETOOLS_REPO_DIR/uax/uax42/fragments/ folder.
```
markusicu, line 5: This will want to be a specific, reproducible `mvn` command line.

uax/uax42/Readme.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Step 2 - Generate TR42 index.html and index.rnc 
+
+- In UNICODETOOLS_REPO_DIR/uax/uax42/ run `mvn xml:transform`
```
markusicu, line 9: mvn command lines should run from the root as usual, see https://github.com/unicode-org/unicodetools/blob/main/docs/build.md for examples.

uax/uax42/Readme.md
```diff
@@ -0,0 +1,24 @@
...
+
+  index.html and index.rnc will be generated in UNICODETOOLS_REPO_DIR/uax/uax42/output/
```
markusicu, line 11: The output should go into a folder under UNICODETOOLS_GEN_DIR as usual, such as UNICODETOOLS_GEN_DIR/ucdxml/17.0.0/

uax/uax42/Readme.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Step 3 - Validate generated UAX XML files
+
+You'll need a [RELAX NG](https://relaxng.org/) schema validator. We'll use [jing-trang](https://github.
+com/relaxng/jing-trang) in this example.
+
+1. Clone and build [jing-trang](https://github.com/relaxng/jing-trang)
+2. Run the following:
+    ```
+   java -jar C:\_git\jing-trang\build\jing.jar -c UNICODETOOLS_REPO_DIR\uax\uax42\output\index.rnc <path to UAX xml file>
```
markusicu, line 21: mvn ...

uax/uax42/Readme.md
```diff
@@ -0,0 +1,24 @@
...
+   ```
+   Note that the UAX xml file has to be saved as NFD as the Unihan syntax regular expressions are expecting NFD.
```
markusicu, line 23: Does the tool generate the data files in NFD?
It seems like the files should come out as needed for the tool chain as well as for publication.

markusicu (commented):
Hi John, I took a peek -- not yet at the new .java files... -- and we discussed some high-level things in today's Unicode Tools meeting.

Some comments below.
For publication, we are thinking that we continue to copy two of the grouped files into the repo, but run the tool as part of a publication step, so that we don't check in all of the large, highly redundant files.

unicodetools/src/main/java/org/unicode/xml/AttributeResolver.java
jowilco: I think that I was assuming that there were going to be more special cases, but I agree that as there are not, your solution is better. Implemented.

uax/uax42/Readme.md
jowilco: Moved to docs/ucdxml.md

uax/uax42/fragments/block/block.xml
jowilco: Moved to resources/org/unicode/uax42/

uax/uax42/fragments/boolean/boolean.xml
jowilco: Done

uax/uax42/fragments/datatypes/code points.xml
jowilco: Done

uax/uax42/Readme.md
jowilco: Done, but please check to see if this is what you were thinking.

uax/uax42/Readme.md
jowilco: I think I have what you want, but...

uax/uax42/Readme.md
jowilco: The data files are not NFD by default, and I'm not sure that we should change the format for publication.
The rest of your question ties back to my previous comment; we could add a step to create an NFD version of the UCDXML files as part of an end-to-end process. However, is that in scope?

uax/uax42/Readme.md
jowilco: I think that we should probably discuss what is in-scope and what is out-of-scope for the process and the utility. The deliverables are:
1. The UCDXML files
2. The TR42 HTML file
3. The RNC file

The fragment files are generated, but are then consumed by the TR42 HTML file and the RNC file. I *think* that these should be in the repo, especially as some of them are created manually.

Should any of the deliverables be stored in unicodetools?
If we are planning to validate the UCDXML files using the RNC files as part of the build, we'll need a process that incorporates all steps. Should that be considered a "unit test"?

uax/uax42/Readme.md
jowilco: I think I have what you want, but...

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
```diff
@@ -0,0 +1,825 @@
...
+                    ucdVersion = VersionInfo.getInstance(options[UCDVERSION].value);
+                } catch (Exception e) {
+                    throw new IllegalArgumentException(
+                            "Could not convert "
+                                    + options[UCDVERSION].value
+                                    + " to a valid UCD version");
+                }
+            } else {
+                throw new IllegalArgumentException(
+                        "Missing command line option: --ucdversion (or -v)");
```
markusicu, line 92: how about we default to Settings.latestVersion?

and, other than maybe right now for comparison with unicode 16, do we even need a version parameter?

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
```diff
@@ -0,0 +1,825 @@
...
+                            throw new IOException();
+                        }
+                    }
+                } catch (Exception e) {
+                    throw new IllegalArgumentException(
+                            "Could not find or create " + options[OUTPUTFOLDER].value);
+                }
+            } else {
+                throw new IllegalArgumentException(
+                        "Missing command line option: --outputfolder (or -f)");
```
markusicu, line 140: optional: as long as the output folder is in the "resources", it should be possible to get a File or Path to there via the class/ClassLoader APIs

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
```diff
@@ -0,0 +1,825 @@
...
+    }
+
+    private static void buildUcdXMLFile(
+            VersionInfo ucdVersion,
+            File destinationFolder,
+            UCDXMLOUTPUTRANGE outputRange,
+            UCDXMLOUTPUTTYPE outputType)
+            throws IOException, TransformerConfigurationException, SAXException {
+        int lowCodepoint = 0x0;
+        int highCodepoint = 0x10FFFF;
```
markusicu, line 177: nit: "code point" is two words, so it would be nice to change "...odepoint" to "...odePoint" / in this whole new package

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
```diff
@@ -0,0 +1,825 @@
...
+            return Range.NONCHARACTER;
+        }
+        return Range.RESERVED;
+    }
+
+    private static int getLastCodepointInGroup(
+            AttributeResolver attributeResolver, int lowCodepoint, int highCodepoint) {
+        String blk = attributeResolver.getAttributeValue(UcdProperty.Block, lowCodepoint);
+        for (int codepoint = lowCodepoint; codepoint <= highCodepoint; codepoint++) {
+            if (!blk.equals(attributeResolver.getAttributeValue(UcdProperty.Block, codepoint))) {
```
markusicu, line 615: nit: it should be easy to get the last code point of the blk from the IUP, and do that once outside the loop

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
```diff
@@ -0,0 +1,825 @@
...
+                return codepoint - 1;
+            }
+            if (codepoint == 0x20 - 1 // put the C0 controls in their own group
+                    || codepoint == 0xa0 - 1 // put the C0 controls in their own group
```
markusicu, line 619:
```suggestion
                    || codepoint == 0xa0 - 1 // put the C1 controls in their own group
```

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
```diff
@@ -0,0 +1,825 @@
...
+                                attrValue,
+                                attributeResolver.isUnihanAttributeRange(codepoint),
+                                propDetail,
+                                prop,
+                                outputRange);
+                if (isAttributeIncluded) {
+                    String propName = prop.getShortName();
+                    if (propName.startsWith("cjk")) {
+                        propName = propName.substring(2);
+                    }
```
markusicu, line 656: FYI: that substring should be the same as the prop.getLongName()

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
```diff
@@ -0,0 +1,825 @@
...
+                                    bestAttrValue,
+                                    attributeResolver.isUnihanAttributeRange(lowCodepoint),
+                                    propDetail,
+                                    prop,
+                                    outputRange);
+                    if (isAttributeIncluded) {
+                        String propName = prop.getShortName();
+                        if (propName.startsWith("cjk")) {
+                            propName = propName.substring(2);
+                        }
```
markusicu, line 725: FYI ditto

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
```diff
@@ -0,0 +1,825 @@
...
+                                propDetail.getUcdProperty(), range.get(0));
+
+                attributes.addAttribute(
+                        NAMESPACE, prop.getShortName(), prop.getShortName(), "CDATA", attrValue);
+            }
+        }
+        return attributes;
+    }
+
+    private static String getVersionString(VersionInfo version, int maxDigits) {
```
markusicu, line 805: I know that [VersionInfo.getVersionString()](https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/com/ibm/icu/util/VersionInfo.html#getVersionString-int-int-) is marked `@internal` but I suggest you call it anyway, rather than reinvent it.

unicodetools/src/main/java/org/unicode/xml/XMLProperties.java
```diff
@@ -0,0 +1,482 @@
+package org.unicode.xml;
+
+import com.ibm.icu.dev.util.UnicodeMap;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.util.*;
```
markusicu, line 7: Please don't import with "*"

unicodetools/src/main/java/org/unicode/xml/XMLProperties.java
```diff
@@ -0,0 +1,482 @@
...
+import java.util.Map.Entry;
```
markusicu, line 8: Please don't import nested classes; call sites are easier to understand if they say `Map.Entry` rather than `Entry`.

unicodetools/src/main/java/org/unicode/xml/UcdPropertyDetail.java
markusicu, line 1: FYI: Wow, this is long and tedious!
Is this by chance generated (if so, how?) -- or else generatable?

unicodetools/src/main/java/org/unicode/xml/UCDXMLWriter.java
```diff
@@ -0,0 +1,74 @@
...
+        transformer.setOutputProperty("{http://xml.apache.org/xalan}indent-amount", "3");
+        transformerHandler.setResult(new StreamResult(f));
+    }
+
+    public void startFile() throws SAXException {
+        transformerHandler.startDocument();
+        char[] c = "\n".toCharArray();
+        transformerHandler.characters(c, 0, c.length);
+        // TODO: JRW change hardcoded 2023 to current year.
+        c = " \u00A9 2023 Unicode\u00AE, Inc. ".toCharArray();
```
markusicu, line 43: yes, and currently 2025 :-)

unicodetools/src/main/java/org/unicode/xml/UCDDataResolver.java
markusicu, line 1: FYI: I am severely glossing over the large amount of Java code. I trust the author based on the output...

unicodetools/src/main/java/org/unicode/xml/GeneratePropertyValues.java
```diff
@@ -0,0 +1,1749 @@
...
+import java.net.URISyntaxException;
+import java.nio.charset.StandardCharsets;
+import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.unicode.props.PropertyParsingInfo;
+import org.unicode.props.UcdProperty;
+import org.unicode.props.UcdPropertyValues.*;
+
+public class GeneratePropertyValues {
```
markusicu, line 16: Another very large file... :-/
Could you please add a `/** ... */` comment to each top-level Java class here (one per .java file) to say, in broad strokes, what each class does and how it connects with the other top-level classes?

docs/ucdxml.md
```diff
@@ -0,0 +1,22 @@
...
+- mvn compile exec:java '-Dexec.mainClass="org.unicode.xml.GeneratePropertyValues"' '-Dexec.args="--ucdversion 16.0.0 -f $(cd ./unicodetools/src/main/resources/org/unicode/uax42/fragments; pwd)"' -DCLDR_DIR=$(cd ../cldr ; pwd) -DUNICODETOOLS_GEN_DIR=$(cd ../Generated ; pwd) -DUNICODETOOLS_REPO_DIR=$(pwd)
+
+## Step 2 - Generate TR42 index.html and index.rnc 
+
+- mvn xml:transform -f $(cd ./unicodetools/src/main/resources/org/unicode/uax42/fragments; pwd) -Doutputdir=../Generated/uax42/
+
+## Step 3 - Validate generated UAX XML files
+
+You'll need a [RELAX NG](https://relaxng.org/) schema validator. We'll use [jing-trang](https://github.
+com/relaxng/jing-trang) in this example.
```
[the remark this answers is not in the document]
jowilco: Fixed

unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
```diff
@@ -760,10 +760,20 @@ private static void parsePropertyValueFile(
...
+                            value = "No";
+                        } else {
+                            value = "Yes";
+                        }
+                    } else {
+                        value = "Yes";
+                    }
+                } else {
+                    value = line.getParts()[2];
+                }
```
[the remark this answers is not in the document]
jowilco: Longer story, which I'll email you about. Basically, tried to rebase, broke the branch on which this PR is based on, spent *way* too long trying to fix said branch, then created a new one based on the current HEAD of main. This change has now been picked up.

unicodetools/src/main/resources/org/unicode/props/ExtraPropertyAliases.txt
```diff
@@ -67,6 +67,7 @@ CJKR ; CJK_Radical
 EDCM ; Emoji_DCM
 EKDDI ; Emoji_KDDI
 ESB ; Emoji_SB
+EVS ; emoji_variation_sequence
```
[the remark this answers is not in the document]
jowilco: I'm not quite sure how to answer your question.

Yes, the variation sequences for emoji have been displayed in the UCDXML files. For example,
`      <standardized-variant cps="0023 FE0E" desc="text style" when=""/>`
`      <standardized-variant cps="0023 FE0F" desc="emoji style" when=""/>`

Neither EKDDI nor Emoji_KDDI are defined in TR51. Nor are they defined in EmojiSources.txt. I'm not sure where else these names would be defined. I just tried to come up with an alias + name to support the values that are in ucd/emoji/emoji-variation-sequences.txt.

unicodetools/src/main/resources/org/unicode/props/ExtraPropertyAliases.txt
```diff
@@ -150,6 +151,13 @@ cjkJoyoKanji ; kJoyoKanji
 cjkKoreanEducationHanja ; kKoreanEducationHanja
 cjkKoreanName ; kKoreanName
 cjkTGH ; kTGH
+kRSTUnicode ; kRSTUnicode
+kReading ; kReading
+kSrc_NushuDuben ; kSrc_NushuDuben
+kTGT_MergedSrc ; kTGT_MergedSrc
+ncCorrected ; NC_Corrected
+ncOriginal ; NC_Original
+ncVersion ; NC_Version
```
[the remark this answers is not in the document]
jowilco: Similar to your question about EVS/emoji_variation_sequence, these were added to support a small section of Normalization Corrections (https://www.unicode.org/reports/tr42/#d1e4038). For example,
`<normalization-correction cp="F951" old="96FB" new="964B" version="3.2.0"/>`

The source for these elements is NormalizationCorrections.txt, and that is where the names come from (Original (erroneous) decomposition, Corrected decomposition, Version of Unicode for...)

Again, I accept that I'm creating new aliases, and perhaps this is the single use-case for them.

unicodetools/src/main/resources/org/unicode/uax42/fragments/block.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--Manual-->
+<ucdxml:block xmlns:ucdxml="http://unicode.org/ns/2001/ucdxml" title="blocks" id='schema.block'>
+  ucd.content &amp;=
+    element blocks {
+      element block {
+        attribute first-cp { single-code-point },
+        attribute last-cp { single-code-point },
+        attribute name { text } }+ }?
```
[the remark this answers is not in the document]
jowilco: Yeah, not sure why he did that.
`blk` is an attribute on a single cp or a cp range.
`block` is an element that specifies the range and the **name** of the block.

I assume that you are suggesting that line 9 should be:
`attribute name { blk } }+ }?`

However, note that the `blk` attribute values don't match the `name` attribute values of `block`, for example:

- blk = ASCII
- block = Basic Latin

unicodetools/src/main/resources/org/unicode/uax42/index.xml
```diff
@@ -0,0 +1,1353 @@
...
+>
+
+    <title>Unicode Character Database in XML</title>
+
+    <articleinfo>
+        <unicode:tr number='42' class='uax' version='16.0.0'
+                    stage='proposed-update'
+                    schema='rnc' prevrev='34'/>
+        <copyright>
+            <year>2024</year>
```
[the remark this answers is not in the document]
jowilco: Done

unicodetools/src/main/resources/org/unicode/uax42/index.xml
```diff
@@ -0,0 +1,1353 @@
...
+        </copyright>
+        <authors>
+            <author>
+                <surname>Wilcock</surname>
+                <firstname>John</firstname>
+            </author>
+        </authors>
+        <revhistory>
+            <revision revnumber="36" date="2024-08-15">
```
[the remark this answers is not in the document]
jowilco: Potentially the copyright year.
For the `revision` date, I'll need to understand if this is the date that the revision was created, the date that the revision was last modified, or an intended publication date. Depending on that answer, we might be able to do something.

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
jowilco: Let me think about this for a bit. I see your point as the assumption is that we'll always be running on the latest version, but I'm not sure if that is a valid assumption.

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
jowilco: Are you expecting that the output would be in \unicodetools\data\ucdxml\ ?

This comes back to the discussion that we had in the unicodetools meeting - do we expect/require that ucdxml always is in sync with \data\ucd\ ? I think that we concluded that ucdxml will only be updated intermittently, and in which case I'd prefer to generate interim versions outside of the \unicodetools\ tree.

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
jowilco: Done

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
jowilco: Done

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
jowilco: There is no getLongName() in UcdProperty.java.

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
jowilco: Ditto! ;-)

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
jowilco: It's marked as `@Deprecated` not `@internal`. Is your recommendation still the same?

unicodetools/src/main/java/org/unicode/xml/XMLProperties.java
jowilco: Fixed.

unicodetools/src/main/java/org/unicode/xml/XMLProperties.java
jowilco: I think that I fixed all of these.

unicodetools/src/main/java/org/unicode/xml/UcdPropertyDetail.java
eggrobin: This does seem a bit worrisome from a maintenance standpoint: we add properties reasonably often (especially provisional Unihan ones), and we will likely forget about this.
Some of it seems derivable: the cjkness of a property is tested passim is the tools by seeing if the short alias starts with "cjk" (Yes, the short alias. "cjk" is short for "k". We're all mad here.).
Some of it seems independent, but ideally it would be moved into one of the data files (IndexUnicodeProperties?).

jowilco: It wasn't exactly generated, but I automated it somewhat via a spreadsheet. There was some research needed to determine the correct version number.

I'm hoping that we'll now only need to append, not regenerate.

unicodetools/src/main/java/org/unicode/xml/UCDXMLWriter.java
jowilco: Done

unicodetools/src/main/java/org/unicode/xml/UCDDataResolver.java
jowilco: That's brave...

unicodetools/src/main/java/org/unicode/xml/UcdPropertyDetail.java
jowilco: @eggrobin 
>>Some of it seems independent, but ideally it would be moved

I was somewhat surprised that the version when each property was added was not defined elsewhere.

The rest of the values are to a certain extent convenience, but isCJKShowIfEmpty is to handle somewhat arbitrary decisions that Eric made when deciding whether an empty cjk property should be shown.

eggrobin: > I was somewhat surprised that the version when each property was added was not defined elsewhere.

Yes, I have been annoyed by that. VersionedProperty checks whether the property assignment is the same for all code points as a proxy for the property not actually being defined... https://github.com/unicode-org/unicodetools/blob/f068b26207961c4bcb0a49e16f4b5ea38c00145f/unicodetools/src/main/java/org/unicode/text/UCD/VersionedProperty.java#L132

unicodetools/src/main/java/org/unicode/xml/GeneratePropertyValues.java
jowilco: Done

unicodetools/src/main/resources/org/unicode/props/ExtraPropertyAliases.txt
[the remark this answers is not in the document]
markusicu: Most important: We should not publish names that are not approved by the UTC. Or at least not more than what UAX42 published already.

Second: If we can **avoid** parsing files like EmojiSources.txt and NormalizationCorrections.txt, and especially inventing properties for them, then let's do so.

unicodetools/src/main/resources/org/unicode/props/ExtraPropertyAliases.txt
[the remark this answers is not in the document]
markusicu: As Asmus has been pointing out, we have a history of publishing data files with "data" but without formally defining corresponding properties. We should skip them if possible. We should definitely not publish unsanctioned property and value names.

I understand that this file is internal to unicodetools. If we don't need these novelties for UCDXML, nor for other Unicode Tools work, then I prefer we don't add them even here.

unicodetools/src/main/resources/org/unicode/uax42/fragments/block.xml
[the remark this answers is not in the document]
markusicu: weird. 🤷

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
markusicu: That's fine.
But if we output into the "resources" path, then we should be able to get that path from Java APIs, rather than having the user tell us where it is.

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
markusicu: maybe it's called something else? getName()? seems like there should be a function for each one of the two main aliases.

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
markusicu: yes

unicodetools/src/main/resources/org/unicode/props/IndexUnicodeProperties.txt
jowilco: Just FYI - because I rebased to main, UCDXML now supports kZhuang.

unicodetools/src/main/resources/org/unicode/props/ExtraPropertyAliases.txt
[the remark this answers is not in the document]
jowilco: >>>Or at least not more than what UAX42 published already.
Both of these (EmojiSources.txt -> variation sequence, NormalizationCorrections.txt) are already part of what Eric had defined in UAX42. Not sure what you want me to do at this point.

unicodetools/src/main/resources/org/unicode/props/ExtraPropertyAliases.txt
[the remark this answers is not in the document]
jowilco: They are needed for UAX42. See https://www.unicode.org/reports/tr42/#d1e4038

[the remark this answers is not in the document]
eggrobin: > If we don't need these novelties for UCDXML, 

As John points out, this ship has sailed, but
> nor for other Unicode Tools work,

in any case I have been adding such things pretty systematically because it is useful to have them in the tools in _some_ form, even if that form is unstable and not fully designed.

I am not sure we need to invent a short alias (e.g., EVS, ncWhatever) though, and if we do not need it, I think we should avoid it; just make it the same as the long name (so, in this file, `Something ; Something` rather than `sth ; Something`).

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
jowilco: Changed to:
`                        if (propName.startsWith("cjk")) {
                            propName = prop.getNames().getAllNames().get(1);
                        }
`

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
jowilco: Ditto (#3)

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
jowilco: Gone. Replaced with com.ibm.icu.util.VersionInfo#getVersionString

unicodetools/src/main/resources/org/unicode/props/ExtraPropertyAliases.txt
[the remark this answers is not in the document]
jowilco: @eggrobin That seems reasonable to me.
so:
`emoji_variation_sequence ; emoji_variation_sequence`

and
`normalization_correction_original ; normalization_correction_original`
`normalization_correction_corrected ; normalization_correction_corrected`
`normalization_correction_version ; normalization_correction_version`

I'll wait for @markusicu to agree before making the change.

uax/uax42/Readme.md
jowilco: Discussed on 2/18

unicodetools/src/main/resources/org/unicode/props/ExtraPropertyAliases.txt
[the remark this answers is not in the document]
jowilco: See https://github.com/unicode-org/unicodetools/issues/1049

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
jowilco: Discussed on 2/18 and implemented.

unicodetools/src/main/java/org/unicode/xml/UcdXML.java
jowilco: Discussed on 2/18 and implemented using Settings.Output.GEN_DIR

unicodetools/src/main/resources/org/unicode/props/ExtraPropertyAliases.txt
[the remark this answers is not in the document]
jowilco: Fixed in https://github.com/unicode-org/unicodetools/pull/1030

PR #863, opened 2024-06-10
Require a link to the RMG issue for pipeline PRs

eggrobin:


markusicu:
Are you sure you want to require this? It would keep us from drafting properties before there is an RMG issue.
@josh-hadley @macchiati @PeterConstable FYI

markusicu:
PS: I have so far used the indirection from RMG issue to SAH/SEW issue and looked there for the unicodetools PR.

eggrobin:
> It would keep us from drafting properties before there is an RMG issue.

Well, we are requiring a posted L2 doc, and I have been drafting properties before the document gets posted.
In general we want to wait until we have a posted document before PAG looks at the draft properties (that way we expect things to be green in PAG review), and by that time the RMG issue is filed; so I don’t think this would pose a problem we don’t already have.

eggrobin:
> I have so far used the indirection from RMG issue to SAH/SEW issue and looked there for the unicodetools PR.

Yes, and I have done the reverse; and with [nearly 50](https://github.com/unicode-org/unicodetools/pulls?q=is%3Apr+is%3Aopen+label%3Adata-for-new) of these PRs open I am getting a bit tired of the indirections.

markusicu (approved):

PR #865, opened 2024-06-26
Allow invariant test statements to extend over multiple lines

eggrobin:
No semicolons for now; the current grammar almost didn’t need them.
The compound properties on the right-hand side of OnPairsOf and In lines were too greedy and started parsing the keyword at the beginning of the next test; but disallowing newlines in property names solves the problem and does not seem like a big problem anyway.

Maybe at some point we will want to add semicolons just for readability (and better parse errors).

Note that I still plan to do [the `:=` thing](https://github.com/unicode-org/unicodetools/pull/855#discussion_r1630379027), but I will do it separately.


macchiati (approved):
Nice!

PR #866, opened 2024-06-26
Use := for definitions in the invariant tests

eggrobin:
See https://github.com/unicode-org/unicodetools/pull/855#discussion_r1629990475 sqq.

Since Markus wanted := rather than ≔ this takes a bit of restructuring of the parser; I wrote it in a way that should avoid running into the C++03 >> problem in the future.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
@@ -694,8 +694,8 @@ public class TestUnicodeInvariants {
     // 1. words: runs of [^\p{Pattern_White_Space}\p{Pattern_Syntax}];
     // 2. simple operators: sequences of the form \p{Pattern_Syntax} \p{Mn}*;
     // 3. explicitly expected sequences of words and simple operators without intervening spaces;
-    //    this allows for contextually accepting operators such as :=, <<, ’s, or .GT.,
-    //    without treating, e.g., every << as atomic.
+    //    this allows for contextually accepting operators such as :=, >>, ’s, or .GT.,
+    //    without treating, e.g., every >> as atomic.
     private static class Lookahead {
         // Advances pp through any pattern white space, then looks ahead one token.
         public static Lookahead oneToken(ParsePosition pp, String text) {
--- a/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
@@ -1113,7 +1113,7 @@ public class TestUnicodeInvariants {
     private static void letLine(ParsePosition pp, String source) throws ParseException {
         expectToken("$", pp, source);
         final String variable = Lookahead.oneTokenNoSpace(pp, source).consume();
-        expectToken("=", pp, source);
+        expectToken(":=", pp, source);
         final int valueStart = pp.getIndex();
         final UnicodeSet valueSet = parseUnicodeSet(source, pp);
         valueSet.complement().complement();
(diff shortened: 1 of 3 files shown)


unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -691,26 +690,69 @@ private static void inLine(
...
     // A one-token lookahead.
-    // Tokens are defined as runs of [^\p{Pattern_White_Space}\p{Pattern_Syntax}],
-    // or single code points in \p{Pattern_Syntax}.
+    // Tokens are defined as:
+    // 1. words: runs of [^\p{Pattern_White_Space}\p{Pattern_Syntax}];
+    // 2. simple operators: sequences of the form \p{Pattern_Syntax} \p{Mn}*;
+    // 3. explicitly expected sequences of words and simple operators without intervening spaces;
+    //    this allows for contextually accepting operators such as :=, >>, ’s, or .GT.,
+    //    without treating, e.g., every >> as atomic.
     private static class Lookahead {
```
markusicu, line 699: Nit: It would be nice to use `/** ... */` doc strings so that these docs show up in an IDE.

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -691,26 +690,69 @@ private static void inLine(
...
+                if (continuation == null) {
+                    break;
+                }
+                candidate =
+                        new Lookahead(candidate.token + continuation.token, pp, continuation.next);
+
+                for (String s : sequences) {
+                    if (candidate.token.equals(s)) {
+                        result = candidate;
+                        break;
```
markusicu, line 736: I think you want to break out of the outer loop here as well. If so, the simplest would be to just `return candidate;` right here.

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -691,26 +690,69 @@ private static void inLine(
...
+                    break;
+                }
+                final Lookahead continuation = oneTokenNoSpace(candidate.next, text);
+                if (continuation == null) {
+                    break;
+                }
+                candidate =
+                        new Lookahead(candidate.token + continuation.token, pp, continuation.next);
+
+                for (String s : sequences) {
```
markusicu, line 733: Nit: If you passed the sequences in as some kind of a Collection, then you could replace this loop with .contains(candidate). (And the previous inner loop could probably use a Stream...anyMatch() construct.)

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
eggrobin: I don’t want to break here; I could be looking for `==` or `===`, and in that case I would want to wait until I get `===`.

unicodetools/src/main/java/org/unicode/text/UCD/GenerateBreakTest.java
```diff
@@ -1735,6 +1735,7 @@ public GenerateLineBreakTest(UCD ucd, Segmenter.Target target) {
                         // Examples for LB21a.
                         " ⁧John ו-Michael⁩;", // No break after ו-‏.
                         "וַֽיְהִי־כֵֽן׃", // Break after maqaf since Unicode 16.
+                        "︷ \uD81B\uDFF1\u302b⸠ᅛ" + "ᆅ",
```
markusicu, line 1738: intentionally part of this PR?

unicodetools/src/main/resources/org/unicode/tools/SegmenterDefault.txt
```diff
@@ -209,7 +209,7 @@ $QU_Pf=($QU_Pf $X)
 $QUmPi=($QUmPi $X)
 $QUmPf=($QUmPf $X)
 
-$NotEastAsian=($NotEastAsian $X)
+$NotEastAsian=($NotEastAsian | [$NotEastAsian - [$SP $BK $CR $LF $NL $ZW]] $X)
```
markusicu, line 212: intentionally part of this PR?

eggrobin: No.

markusicu (approved):

PR #867, opened 2024-06-29
NamesList-16.0.0d25.txt

eggrobin:
From @Ken-Whistler:

> This incorporates a number of subhead fixes for the new Egyptian
Hieroglyphs block, noted by Michel Mariani and verified by Michel
Suignard (so they have a double-Michel imprimatur). I also rolled in a
fix for an out-of-place subhead in Garay noted by Peter C, and a couple
of xrefs requested by Charlotte Buff.
> 
> I may have missed something, but I think this catches the names list
annotations up with recent feedback.


Ken-Whistler (approved):
As intended. Looks good.

PR #868, opened 2024-06-29
Propertywise correspondence

eggrobin:
This should make things more readable when it comes to ignoring properties, and make it possible to write propertywise tests for new cased or decomposable characters.


macchiati (approved):
Lgtm. Am on vacation, so only intermittently reading emai

PR #876, opened 2024-07-05
Update LB10 and LB21a for implementability

eggrobin:
See https://github.com/unicode-org/unicode-reports/pull/158.


macchiati (approved):
Only spot checked a few lines in the generated files though.

PR #880, opened 2024-07-20
Create devcontainer.json

Diamondb32:
Is this right


markusicu:
We would need more information here.
- What does publishing gradle-built packages do for us, for the project, for you?
- What are you trying to achieve?
- What are you trying to use the unicodetools for?
- Why not use the Maven build?
- Note that the unicodetools are not production-level code. They are used for the maintenance of the Unicode Character Database and related things.

PR #892, opened 2024-07-25
Unikemet 16 draft 17: kEH_Core is no longer binary

eggrobin:
From @Ken-Whistler:
> Michel has provided an important update for Unikemet.txt that should be
> pulled into unicodetools.
> 
> This fixes an oversight in the generation of the kEH_Core values in the
> file after that property was changed from binary to an enumerated
> ternary property.
> 
> I have verified that the change doesn't trigger any updates for the
> generated NamesList.txt, so I'm not also posting an update for that.


unicodetools/src/main/resources/org/unicode/props/ExtraPropertyValueAliases.txt
```diff
@@ -205,6 +204,11 @@ Do_Not_Emit_Type ; Preferred_Spelling       ; Preferred_Spelling
 # @missing: 0000..10FFFF; kSrc_NushuDuben ; <none>
 # @missing: 0000..10FFFF; kReading ; <none>
 
+# @missing: 0000..10FFFF; kEH_Core  ;  N
+kEH_Core ; C ; C
+kEH_Core ; L ; L
+kEH_Core ; N ; N
```
markusicu, line 210: should we set the long names to Core / Legacy / None?
https://www.unicode.org/reports/tr57/tr57-2.html#N101BA

eggrobin: Hm. That wording seems a bit squishy (I read it as explaining the origin of the one-letter values rather than formally defining long names), but since this is all provisional and since defining long names here will have no effect on the Standard and associated published data files, and indeed no effect anywhere except in display in the online tools, we might as well do it for readability.

markusicu (approved):

PR #908, opened 2024-08-09
Work around a UnicodeSet bug

eggrobin:
See https://www.unicode.org/reports/tr35/#unicodeset-syntax, `\N{whatever}` is a `quoted`, which is a `char`, which is an `element`, so one should be able to use it in a `range`.

ICU treats it a a synonym for `\p{Name=whatever}`, which is a `unicodeSet`, so that the thing that should be a range is a set difference (equal to its LHS in practice).

That also means that `\N{nonexistent name}` is silently empty.

We should fix it in ICU, but we urgently need to fix it here for UCD development, because _we have already been bitten by it_, with no-one noticing; the following only tests the first letter of the Garay alphabet:
https://github.com/unicode-org/unicodetools/blob/cccbe935932337bc7f87b8a7b151d6e2fcf78a17/unicodetools/src/main/resources/org/unicode/text/UCD/AdditionComparisons.txt#L27-L34

### DIFF
--- a/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
@@ -2036,10 +2036,10 @@ public class TestUnicodeInvariants {
             return icuSet;
         }
         // Simplest way for the lambda function to report errors.
-        // It cannot throw a ParseException, and it cannot modify class fields.
-        // It _can_ modify what this field points to.
+        // It cannot throw a ParseException, and it cannot modify local variables.
+        // It _can_ modify what this local variable points to.
         // Below, we will throw a ParseException for the first bad position.
-        List<Integer> badEscapePositions = new ArrayList<>();
+        final List<Integer> badEscapePositions = new ArrayList<>();
         unicodeSetExpression =
                 matcher.replaceAll(
                         (MatchResult match) -> {
--- a/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
@@ -2035,6 +2035,10 @@ public class TestUnicodeInvariants {
         if (!matcher.find()) {
             return icuSet;
         }
+        // Simplest way for the lambda function to report errors.
+        // It cannot throw a ParseException, and it cannot modify class fields.
+        // It _can_ modify what this field points to.
+        // Below, we will throw a ParseException for the first bad position.
         List<Integer> badEscapePositions = new ArrayList<>();
         unicodeSetExpression =
                 matcher.replaceAll(
--- a/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
@@ -2052,6 +2052,7 @@ public class TestUnicodeInvariants {
                                     ' ');
                         });
         for (int p : badEscapePositions) {
+            // Simplest way to throw an exception for only the first list element.
             throw new ParseException("No character matching \\N escape", initialPosition + p);
         }
         var patchedParsePosition = new ParsePosition(0);
--- a/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
@@ -1,5 +1,6 @@
 package org.unicode.text.UCD;
 
+import com.google.common.base.Strings;
 import com.ibm.icu.dev.tool.UOption;
 import com.ibm.icu.dev.util.UnicodeMap;
 import com.ibm.icu.lang.UCharacter;
@@ -2045,14 +2046,14 @@ public class TestUnicodeInvariants {
                                 badEscapePositions.add(match.start());
                                 return "";
                             }
-                            return String.format(
-                                    "%" + (match.group().length() + 1) + "s",
-                                    "\\\\x{" + Integer.toHexString(character.charAt(0)) + "}");
+                            return Strings.padStart(
+                                    "\\\\x{" + Integer.toHexString(character.charAt(0)) + "}",
+                                    match.group().length() + 1,
+                                    ' ');
                         });
         for (int p : badEscapePositions) {
             throw new ParseException("No character matching \\N escape", initialPosition + p);
         }
-        System.err.println(unicodeSetExpression);
         var patchedParsePosition = new ParsePosition(0);
         try {
             return new UnicodeSet(unicodeSetExpression, patchedParsePosition, symbolTable);
--- a/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
@@ -26,6 +26,7 @@ import java.util.Set;
 import java.util.Stack;
 import java.util.TreeMap;
 import java.util.function.Function;
+import java.util.regex.MatchResult;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import org.unicode.cldr.draft.FileUtilities;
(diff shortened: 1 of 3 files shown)


markusicu:
test failure: https://github.com/unicode-org/unicodetools/actions/runs/10322995005/job/28579439631?pr=908
```
ParseErrorCount=6
TestFailureCount=0
Error:  Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 63.737 s <<< FAILURE! - in org.unicode.text.UCD.TestTestUnicodeInvariants
Error:  org.unicode.text.UCD.TestTestUnicodeInvariants.testUnicodeInvariants  Time elapsed: 63.719 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: TestUnicodeInvariants.testInvariants(default) failed ==> expected: <0> but was: <6>
	at org.unicode.text.UCD.TestTestUnicodeInvariants.testUnicodeInvariants(TestTestUnicodeInvariants.java:39)
```

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -2007,17 +2008,58 @@ public BackwardParseException(String s, int errorOffset) {
...
+                            UnicodeSet character =
+                                    new UnicodeSet(
+                                            match.group(), new ParsePosition(0), symbolTable);
+                            if (character.isEmpty()) {
+                                badEscapePositions.add(match.start());
+                                return "";
+                            }
+                            return String.format(
+                                    "%" + (match.group().length() + 1) + "s",
+                                    "\\\\x{" + Integer.toHexString(character.charAt(0)) + "}");
```
markusicu, line 2050: What does this do, other than return "\\x{code point}"? Pad it to the length of the \N{} substring?

If so, can we use Guava [Strings.padStart(replacement, length, ' ')](https://guava.dev/releases/31.1-jre/api/docs/com/google/common/base/Strings.html#padStart(java.lang.String,int,char))?

CLDR uses Guava. I assume we use it, or could use it, in the unicodetools as well.

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -2007,17 +2008,58 @@ public BackwardParseException(String s, int errorOffset) {
...
+                        });
+        for (int p : badEscapePositions) {
+            throw new ParseException("No character matching \\N escape", initialPosition + p);
+        }
```
markusicu, line 2054: why the List and the loop? the first bad position will throw, and thus exit this function. why not throw right away when you find character.isEmpty()?

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -2007,17 +2008,58 @@ public BackwardParseException(String s, int errorOffset) {
...
+        System.err.println(unicodeSetExpression);
```
markusicu, line 2055: temporary for debugging?

unicodetools/src/test/java/org/unicode/text/UCD/TestTestUnicodeInvariants.java
```diff
@@ -49,4 +53,31 @@ void testSecurityInvariants() throws IOException {
...
+                TestUnicodeInvariants.parseUnicodeSet(
+                                "TEST [\\N{LATIN SMALL LETTER A}-\\N{LATIN SMALL LETTER Z}]",
+                                new ParsePosition(5))
+                        .size());
+        ParseException thrown =
+                assertThrows(
+                        ParseException.class,
+                        () ->
+                                TestUnicodeInvariants.parseUnicodeSet(
+                                        "TEST [\\N{MEOW}]", new ParsePosition(5)));
```
markusicu, line 70: i see a character encoding proposal brewing ;-)

markusicu:
Is there an ICU ticket? Would you be willing to work on it?

markusicu:
@aheninger FYI

markusicu:
The ICU User Guide does not go into details about `\N`: https://unicode-org.github.io/icu/userguide/strings/regexp.html

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
eggrobin: I cannot throw from a lambda.
It is slightly dumb to build up a list, but other options seemed uglier.

eggrobin:
> Is there an ICU ticket?

Not yet. Could you file one?

> Would you be willing to work on it?

Yes, if you file the ticket please assign it to me.

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
eggrobin: Oops. yes.

markusicu:
Actually, at a closer look, while there is some documentation for ICU regex, there is no documentation for `\N` for UnicodeSet:
- https://unicode-org.github.io/icu/userguide/strings/unicodeset.html
- https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/com/ibm/icu/text/UnicodeSet.html

Trying this in an old C++ UnicodeSet demo gives a weird result: https://icu4c-demos.unicode.org/icu-bin/ubrowse?go=FFFF&us=%5B%5CN%7BDIGIT+FOUR%7D%5D&gosetk.x=12&gosetk.y=23

I think this is not an ICU bug because _ICU UnicodeSet does not support \N_.

In the unicodetools, it must be handled by Mark's SymbolTable implementation. And that probably has no way of distinguishing between "set" and "character" results -- looking at https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/com/ibm/icu/text/UnicodeSet.XSymbolTable.html

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
markusicu: how about keeping just one badEscapePosition? make it an int, initialize with -1, set it if it's still <0, later throw if it's >=0

building a List and looping over it when only the first item will be used seems very confusing, and would want comments to say what's going on.

markusicu:
Actually, clicking a different button on the old demo seems to recognize \N: https://icu4c-demos.unicode.org/icu-bin/ubrowse?go=FFF0&us=%5B%5CN%7BDIGIT+FOUR%7D%5D&gosetn.x=17&gosetn.y=17

Could you please try this in vanilla ICU4J UnicodeSet, without the Unicode Tools machinery?

eggrobin:
@markusicu wrote:
> I think this is not an ICU bug because ICU _UnicodeSet does not support \N_.

https://github.com/unicode-org/icu/blob/b5b3e16afac61f9aa9b775aaf497f8cc88ce9481/icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java#L3741-L3751

https://github.com/unicode-org/icu/blob/b5b3e16afac61f9aa9b775aaf497f8cc88ce9481/icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java#L3791

https://github.com/unicode-org/icu/blob/b5b3e16afac61f9aa9b775aaf497f8cc88ce9481/icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java#L3802-L3812

https://github.com/unicode-org/icu/blob/b5b3e16afac61f9aa9b775aaf497f8cc88ce9481/icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java#L3835-L3850

eggrobin:
Support for `\N` was added on 2002-08-28, in https://github.com/unicode-org/icu/commit/681c0468a4139488ec92e7d027c6aeaab819ae01, for ICU-1767.

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
eggrobin: Except it cannot be an int, because it has to be effectively final in order to be poked at by the λ (my first attempt was an `Integer` initialized with `null`, not unlike what you suggest).

markusicu:
Next thing to check in ICU behavior is whether we allow a subtraction like CHARACTER_CLASS '-' LITERAL. (Using UTS18 syntax rule names here.) If we do, then turning \N into a LITERAL would be harmless. If we don't, then we could potentially break existing patterns.

--> ICU-22851 “UnicodeSet should treat \N like a character not like a set”

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -2007,17 +2009,58 @@ public BackwardParseException(String s, int errorOffset) {
...
+        // ICU incorrectly treats \N{X} as a synonym for \p{Name=X}, returning a
+        // set rather than a character, so that it can be empty, and so that
+        // \N{X}-\N{Y} is a set difference (equal to \N{X}) rather than the range \N{X}-\N{Y}.
+        // This should likely be fixed in ICU, but in the meantime we need to work around it in
+        // the invariant before someone gets hurt.
+        var matcher = nameEscape.matcher(unicodeSetExpression);
+        if (!matcher.find()) {
+            return icuSet;
+        }
+        List<Integer> badEscapePositions = new ArrayList<>();
```
markusicu, line 2038:
```suggestion
        // Simplest way for the lambda function to report errors.
        // It cannot throw a ParseException, and it cannot modify class fields.
        // It _can_ modify what this field points to.
        // Below, we will throw a ParseException for the first bad position.
        List<Integer> badEscapePositions = new ArrayList<>();
```

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -2007,17 +2009,58 @@ public BackwardParseException(String s, int errorOffset) {
...
+                                return "";
+                            }
+                            return Strings.padStart(
+                                    "\\\\x{" + Integer.toHexString(character.charAt(0)) + "}",
+                                    match.group().length() + 1,
+                                    ' ');
+                        });
+        for (int p : badEscapePositions) {
+            throw new ParseException("No character matching \\N escape", initialPosition + p);
+        }
```
markusicu, line 2056:
```suggestion
        for (int p : badEscapePositions) {
            // Simplest way to throw an exception for only the first list element.
            throw new ParseException("No character matching \\N escape", initialPosition + p);
        }
```

markusicu (commented):
there are still CI failures

aheninger:
> The ICU User Guide does not go into details about `\N`: https://unicode-org.github.io/icu/userguide/strings/regexp.html

For ICU4C regex, \N{whatever} matches exactly one code point.
The whatever string, delimited by the braces, is dumped into
```UChar32  theChar = u_charFromName(U_UNICODE_CHAR_NAME, name, fStatus);```
to get the character.

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
markusicu: resolved without accepting the additional comment...

eggrobin: … Huh, I am very confused. I had pressed the « Commit suggestion » button, and it added only part of this suggested comment, and in the wrong place at that ? ? ? https://github.com/unicode-org/unicodetools/pull/908/commits/708a46de24629618e1d5fd32f29cde1f6bb336cf

eggrobin: Ah I see, you had suggested two similar-looking comments in different places.

markusicu (approved):
lgtm tnx

PR #911, opened 2024-08-15
Allow strings in propertywise tests

eggrobin:
Needed to test https://github.com/unicode-org/unicodetools/pull/885.
The strings are considered only as property values, their properties (whatever that means) are not looked at.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
@@ -526,11 +526,11 @@ public class TestUnicodeInvariants {
     }
 
     private static String stringAt(UnicodeSet set, int i) {
-        int codePoints = set.size() - set.strings().size();
-        if (i < codePoints) {
+        final int codePointsSize = set.size() - set.strings().size();
+        if (i < codePointsSize) {
             return Character.toString(set.charAt(i));
         } else {
-            return set.strings().stream().skip(i - codePoints).findFirst().get();
+            return set.strings().stream().skip(i - codePointsSize).findFirst().get();
         }
     }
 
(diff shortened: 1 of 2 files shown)


unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -525,6 +525,15 @@ private static void propertywiseAlikeLine(
         }
     }
 
+    private static String stringAt(UnicodeSet set, int i) {
+        int codePoints = set.size() - set.strings().size();
```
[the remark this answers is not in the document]
macchiati: It might be worth suggesting to Mihai, as he is adding iteration, that he add .codePointsSize() as an API for uniformity.

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -554,18 +562,40 @@ private static void propertywiseCorrespondenceLine(
                                 + ")",
                         pp.getIndex());
             }
+            if (set.hasStrings() && set.strings().size() != set.size()) {
```
macchiati, line 565: That would also change this to the easier to read:
```
            if (set.codePointsSize() > 0 && set.strings().size() > 0) {
```

unicodetools/src/main/java/org/unicode/text/UCD/TestUnicodeInvariants.java
```diff
@@ -538,13 +547,12 @@ private static void propertywiseCorrespondenceLine(
         final List<UnicodeSet> sets = new ArrayList<>();
         sets.add(firstSet);
         expectToken(":", pp, source);
+
+        // Index of the first set of multi-character strings (and of the first multi-character
+        // reference string).
+        int m = -1;
```
macchiati, line 553: use more understandable name, eg firstStringIndex

macchiati (commented):
Only minor comments

macchiati (approved):

markusicu (approved):

PR #917, opened 2024-08-19
Unicode 16 final NamesList

markusicu:
From Ken:
I have done the step to clean out all the intermediate change notes and mark NamesList.txt final for 16.0.
No substantive changes to any of the content of the NamesList.txt file that appears in charts.


Ken-Whistler (approved):
As intended. Looks fine.

PR #918, opened 2024-08-19
Always use latest Unicode version when building emoji

nedley:
When building emoji 16.0 charts I remembered/discovered that the latest Unicode version was only used during beta. Remove the `IS_BETA` checks for everything but human-readable text and rename `LATEST` to `PROPS`. I’m leaving most other variables alone in case we want to bring back the ability to build charts for the previous version.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/tools/emoji/Emoji.java
+++ b/unicodetools/src/main/java/org/unicode/tools/emoji/Emoji.java
@@ -44,18 +44,7 @@ import org.unicode.text.utility.Utility;
  *
  * <p>Example: -Demoji-beta
  *
- * <p>With each new version, set
- *
- * <ul>
- *   <li>VERSION_LAST_RELEASED2 = VERSIONxx;
- *   <li>VERSION_LAST_RELEASED = VERSIONyy;
- *   <li>VERSION_BETA = VERSIONzz;
- * </ul>
- *
- * <p>Normally, when emoji and Unicode versions are in sync, we should use
- * Settings.LATEST_VERSION_INFO etc. here.
- *
- * <p>You also need to
+ * <p>With each new version:
  *
  * <ul>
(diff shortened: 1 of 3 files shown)


unicodetools/src/main/java/org/unicode/tools/emoji/Emoji.java
```diff
@@ -956,10 +938,10 @@ public static Iterable<Source> orderedEnum(Source... doFirst) {
         return ordered;
     }
 
-    public static final IndexUnicodeProperties LATEST =
+    public static final IndexUnicodeProperties PROPS =
             IndexUnicodeProperties.make(VERSION_TO_GENERATE_UNICODE);
     public static final IndexUnicodeProperties BETA =
-            IS_BETA ? IndexUnicodeProperties.make(VERSION_BETA_UNICODE) : LATEST;
+            IndexUnicodeProperties.make(VERSION_BETA_UNICODE);
```
markusicu, line 944: This is now equivalent with PROPS. Can we delete this? Any code still using it?

If you want to keep the variable, then maybe for now just assign it PROPS rather than make a new IUP object for the same version?

macchiati (dismissed):

unicodetools/src/main/java/org/unicode/tools/emoji/Emoji.java
nedley: I’m not prepared to fully flatten the distinction just yet, so I’m taking your suggestion of using the same IUP object.

markusicu (approved):
tnx lgtm pse squash

markusicu:
> tnx lgtm pse squash

meaning i suggest using squash-and-merge -- not squashing and then waiting for people again

nedley:
> meaning i suggest using squash-and-merge -- not squashing and then waiting for people again

Understood—I was waiting for the build checks to complete.

PR #919, opened 2024-08-19
Unicode 16 final NamesList.html

markusicu:
From Ken:
It turns out nobody ever updated NamesList.html (at all) for the beta review. The copy you have still claims it is Unicode 15.1.0. I've just done a pro forma update to version 16.0.0.

FYI, related:
- https://github.com/unicode-org/properties/issues/122


Ken-Whistler (approved):
Looks correct.

Issue #922, opened 2024-08-20
Unicode 16 aux Script Charts bug: 2 Todhri characters in charts for several other scripts


markusicu:
@dwanders-A reports:
I’m checking the Auxiliary Charts and wanted to know why 2 Todhri characters appear at the end of Coptic in the script chart: https://www.unicode.org/charts/beta/script/index.html ?
Whoa… Those two Todhri characters 105C9 105E4 show up at the end Tai Le and Tifinagh too, so something is off.

Even more charts are affected. Look for "105C9" on
- https://www.unicode.org/charts/beta/script/chart_Coptic.html
- https://www.unicode.org/charts/beta/script/chart_Tai_Le.html
- https://www.unicode.org/charts/beta/script/chart_Tifinagh.html
- https://www.unicode.org/charts/beta/script/chart_Hebrew.html
- https://www.unicode.org/charts/beta/script/chart_Syriac.html
- https://www.unicode.org/charts/beta/script/chart_Duployan.html
- https://www.unicode.org/charts/beta/script/chart_Old_Permic.html
- maybe more -- but not all of the script charts

@macchiati ideas? Why these two?
```
UnicodeData.txt
105C9;TODHRI LETTER EI;Lo;0;L;105D2 0307;;;;N;;;;;
105E4;TODHRI LETTER U;Lo;0;L;105DA 0307;;;;N;;;;;

Scripts.txt
105C0..105F3  ; Todhri # Lo  [52] TODHRI LETTER A..TODHRI LETTER OO

Nothing in ScriptExtensions.txt
```

Ken-Whistler:
Those are the only two Todhri letters with canonical decompositions, so I suspect an issue with normalization.
105C9 ; [.5237.0020.0002] # TODHRI LETTER EI
105D2 0307 ; [.5237.0020.0002] # TODHRI LETTER EI
105E4 ; [.5252.0020.0002] # TODHRI LETTER U
105DA 0307 ; [.5252.0020.0002] # TODHRI LETTER U

eggrobin:
Markus wrote:
> Even more charts are affected. Look for "105C9" on
> https://www.unicode.org/charts/beta/script/chart_Coptic.html
> https://www.unicode.org/charts/beta/script/chart_Tai_Le.html
> https://www.unicode.org/charts/beta/script/chart_Tifinagh.html
> https://www.unicode.org/charts/beta/script/chart_Hebrew.html
[...]

PR #923, opened 2024-08-21
script charts: each character on exactly one

markusicu:
Fixes
- https://github.com/unicode-org/unicodetools/issues/922

by changing the script chart generator so that it shows each character on exactly one chart, which is the one with its explicit script (which could be from scx if there is exactly one explicit value).

Avoids showing characters on multiple charts because they or their decomposition mappings have multiple Script_Extensions. The per-script charts are now cleaner than before, and there are more per-General_Category charts for characters that don't have one single explicit script.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/text/UCA/WriteCharts.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCA/WriteCharts.java
@@ -19,7 +19,6 @@ import java.text.NumberFormat;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.BitSet;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -540,13 +539,9 @@ public class WriteCharts implements UCD_Types {
     public static void scriptChart() throws IOException {
         HACK_KANA = false;
 
-        final Set set = new TreeSet();
-        final BitSet toReturn = new BitSet();
+        final Set<Pair> set = new TreeSet<>();
 
         for (int i = 0; i <= 0x10FFFF; ++i) {
-            if (i == 0x0342) {
-                System.out.println("?");
-            }
             if (!Default.ucd().isRepresented(i)) {
                 continue;
             }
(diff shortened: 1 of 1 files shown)


markusicu:
Temporary comment to help reviewing:
- 15.1 charts
  - https://www.unicode.org/charts/script/
  - https://www.unicode.org/charts/script/chart_Coptic.html
- previously generated 16.0 charts
  - https://www.unicode.org/charts/beta/script/
  - https://www.unicode.org/charts/beta/script/chart_Coptic.html
- 16.0 charts from this change
  - https://corp.unicode.org/~book/incoming/markus/u16/auxcharts-20240821/script/
  - https://corp.unicode.org/~book/incoming/markus/u16/auxcharts-20240821/script/chart_Coptic.html

rickmcgowan (commented):
Can't comment on most of the code, but the results look good.

unicodetools/src/main/java/org/unicode/text/UCA/WriteCharts.java
```diff
@@ -909,46 +899,52 @@ static short getBestScript(String s) {
...
+    /**
+     * Returns the best explicit Script value for cp or else for decomp. If there is no such
+     * explicit script, returns a negative value. Never returns Common or Inherited.
+     */
+    private static int getBestScript(int cp, String decomp) {
+        int sc = getExplicitScript(cp);
+        if (sc < 0 && !equals(decomp, cp)) {
+            for (int i = 0; sc < 0 && i < decomp.length(); i += Character.charCount(cp)) {
+                cp = decomp.codePointAt(i);
+                sc = getExplicitScript(cp);
```
nedley, line 911: Last one wins…

markusicu: No, the "for" condition includes `sc < 0` which means when we get the first explicit script we stop.

nedley: Ah, there it is! I think it would be cleaner to explicitly check and break after calling the function, but I’m not going to argue with working code.

markusicu: I like it concise like this...

markusicu: I also thought about using the IntStream functionality, but it might take some playing with the syntax and I am not sure that it would be any more readable, or any faster. (More object creation vs. better compiler optimizations.) Something like
```
decomp.codePoints().map(cp to script).filter(non-negative).findFirst().ifPresent(set sc);
```

macchiati:
The one advantage is that it you don't care about order you can use one of
those multi threaded streams. But for short strings it is even worse.

Manishearth (approved):

PR #941, opened 2024-09-23
The most ancient script from the future

eggrobin:
Same procedure as https://github.com/unicode-org/unicodetools/pull/888, as this one already has a script code.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/text/UCD/UCD_Names.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/UCD_Names.java
@@ -431,6 +431,7 @@ public final class UCD_Names implements UCD_Types {
         "Tolong_Siki",
         "Beria_Erfe",
         "Jurchen",
+        "Proto_Cuneiform",
     };
 
     public static final Relation<String, String> EXTRA_SCRIPT =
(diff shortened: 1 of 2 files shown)


roozbehp:
I don't understand why you need this. Proto-cuneiform isn't planned for 17.0, is it?

eggrobin:
> I don't understand why you need this.

Because I am drafting properties for the Archaic Cuneiform Numerals proposal, which introduces sc=Pcun and scx=Xsux Pcun characters, and I do not want to deal with divergent versions of these Java files (it is only in the actual data files that I can do automated conflict resolution).

Adding these to the enumeration several versions ahead has been standard procedure, see https://github.com/unicode-org/unicodetools/pull/888, https://github.com/unicode-org/unicodetools/pull/660, https://github.com/unicode-org/unicodetools/pull/546, and https://github.com/unicode-org/unicodetools/pull/443. See the last one of those for more explanations.

eggrobin:
> I thought these were sc=Xsux.

The proposal has all three of scx=Xsux, scx=Xsux Pcun, and scx=Pcun (where Xsux « wins » for the Script property). See §§1, 2.3.3, and 2.3.4.

markusicu (approved):

Issue #492, opened 2023-06-02
adjust cldr/*BreakTest generation for Unicode 15.1


markusicu:
Since Unicode 15.1 and consensus UTC-175-C26 (PR #477), the default and CLDR _grapheme break_ rules (SegmenterDefault.txt vs. SegmenterCldr.txt) are the same. Therefore, we _might_ want to stop generating UCD/(version)/cldr/GraphemeBreakTest.* files. We could also continue to generate them just to keep exercising the relevant code paths.

However, the _line break_ rules are still different, and we have not been generating UCD/(version)/cldr/LineBreakTest.* files. Why not? Shouldn't we generate those, check them into the [CLDR repo](https://github.com/unicode-org/cldr/tree/main/common/properties/segments), and use them for testing in ICU?

Also, it seems like the CLDR expression `$QU_Pf=($QU_Pi $X)` has a typo.

More generally, are the diffs in lines related to QU intentional?

@eggrobin @macchiati @echeran

Diffs:
```
$ diff -u unicodetools/src/main/resources/org/unicode/tools/SegmenterDefault.txt unicodetools/src/main/resources/org/unicode/tools/SegmenterCldr.txt
```
-->
```
--- unicodetools/src/main/resources/org/unicode/tools/SegmenterDefault.txt	2023-06-01 10:47:21.084375072 -0700
+++ unicodetools/src/main/resources/org/unicode/tools/SegmenterCldr.txt	2023-06-01 10:47:21.080374729 -0700
@@ -108,8 +108,8 @@
 $ZWJ_O=\p{Line_Break=ZWJ}
 $ZWJ=\p{Line_Break=ZWJ}
 
-$QU_Pi=[$QU & \p{gc=Pi}]
-$QU_Pf=[$QU & \p{gc=Pf}]
+$QU_Pi=($QU_Pi $X)
+$QU_Pf=($QU_Pf $X)
 
 $DottedCircle = ◌
 
@@ -147,7 +147,7 @@
 $Spec3a_=[^ $SP $BA $HY $CM]
 $Spec3b_=[^ $BA $HY $CM]
 $Spec4_=[^ $NU $CM]
-##CLDR: $Spec5_=[$BK $CB $CR $LF $NL $SP $ZW]
+$Spec5_=[$BK $CB $CR $LF $NL $SP $ZW]
 
 # SPECIAL EXTENSIONS
 
@@ -195,7 +195,7 @@
 $ZWJ=($ZWJ $X)
 
[...]

PR #947, opened 2024-10-09
Remove SegmenterCldr.txt

eggrobin:
Back [in March](https://github.com/unicode-org/unicodetools/issues/492#issuecomment-2023947121) I had written:
> Right now the 15.1 line breaking rules in SegmenterCldr.txt are used by no-one, and this is a good thing since they are wrong and untested (and we went through a release with these wrong rules!). Let’s get rid of that quasi-copy before someone gets hurt.

And then I forgot to do anything. They are still wrong, and they have not been updated for 16.0. Let’s get rid of those.


macchiati (dismissed):

macchiati (changes requested):
I'm not sure we should remove the whole mechanism, because we may need it back in the future. Once we settle on overrides with "character x behaves like character y", I'd be comfortable removing.

eggrobin:
We discussed that earlier: https://github.com/unicode-org/unicodetools/issues/492#issuecomment-2024020481.
Right now this is not used, and if it had been used it would have produced garbage, most likely without anyone noticing, so it is a footgun: we really should remove SegmenterCldr.txt before someone gets hurt.
When it comes to the code itself, I am removing the minimum needed to make CI pass; if we ever want to bring back a SegmenterCldr this is trivial to revert (that is why we have source control!).

macchiati (approved):
Ok makes sense

PR #949, opened 2024-10-14
Support remap rules in the segmenter

eggrobin:
This will vastly improve the maintainability of that copy of the rules, as they can closely mirror the UAX; note that this removes 84 lines from SegmenterDefault.txt, including some nontrivial things like
```
$NotEastAsian=( $NotEastAsian | [$NotEastAsian - $Spec1_] $X)
```
or 
```
# LB 10  Treat any remaining combining mark as AL and non-$EastAsian.
$AL=($AL | ^ $CM | (?<=$Spec1_) $CM)
$NotEastAsian=( $NotEastAsian | ^ $CM | (?<=$Spec1_) $CM )
```

Further, this should make it possible to generate ICU’s old monkeys, which implement the remap rules as specified (only the production implementation integrates them into the regices).

No changes to [LineBreakTest.txt](https://github.com/unicode-org/unicodetools/pull/949/files#diff-1b1eeb194c1f2b4087afdaf8d965071c366e8e729e9bd61eaca82ea7dbedfa18) except in comments.


macchiati (approved):
Nice!

PR #959, opened 2024-11-05
Only run Maven cache workflow on the upstream repo

echeran:
Mostly a nice-to-have adjustment that prevents the workflow from running in personal forks. When that happens, Github will complain about running a scheduled workflow on a repo with > 60 days of inactivity


echeran:
> We could probably benefit from this in ICU as well for some of our workflows...

I'll duplicate these efforts mechanically in ICU for the corresponding workflow.

PR #967, opened 2024-11-14
Remove Joining_Group numeric constants from UCD_Types

eggrobin:
Recently discussed in https://github.com/unicode-org/unicodetools/pull/756#discussion_r1842413362.

UCD_Types is part of the most ancient layers of these tools:
https://github.com/unicode-org/unicodetools/blob/6da68627c20dafcdc3e16c2f67196fa45d4cf83d/unicodetools/src/main/java/org/unicode/text/UCD/UCD_Types.java#L2-L3

It is only used in very limited contexts by the more recent strata which generate and parse the bulk of the UCD.

In particular, these constants are not used, and they are annoying to maintain (especially when merging independent proposals).


macchiati (commented):
Anything to reduce the use of UCD_Types and UCD_Names is great!

PR #984, opened 2024-12-06
update to CLDR 46 & ICU 76

markusicu:
- CLDR 46: https://github.com/unicode-org/cldr/packages/521051
- ICU 76: https://unicode-org.github.io/icu/download/76.html#download
  - First we updated to a GH Maven artifact created from a temporary branch that picked up the near-final ICU 76, from just before its version was changed to remove the -SNAPSHOT suffix. We need this suffix for publishing to GH Maven.
    - 76.1-20241206.234144-4 = https://github.com/unicode-org/icu/packages/1954682
  - Then Elango and I realized that now that the utilities-for-cldr artifact is gone, we can just updated to the vanilla 76.1 release which comes from Maven Central.

This also removes references to the utilities-for-cldr artifact which @mihnita removed in ICU-21757, and changes the package of UnicodeMap from com.ibm.icu.dev.util to com.ibm.icu.impl which happened as part of that work.

Probably best to review one commit at a time, but I will probably squash the commits.


echeran (approved):
LGTM. Thanks @mihnita

PR #985, opened 2024-12-09
Fix placeholder name for addition comparison test

eggrobin:
See https://github.com/unicode-org/utc-release-management/issues/154.

Also get rid of some static variables whose initialization was unreliable.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/text/UCA/MakeNamesChart.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCA/MakeNamesChart.java
@@ -69,17 +69,19 @@ public class MakeNamesChart {
         isWhiteSpace = new UnicodeSet(up.getSet("whitespace=Yes"));
 
         Utility.copyTextFile(
-                Settings.SRC_UCA_DIR + "nameslist_index.html",
+                MakeNamesChart.class.getResource("nameslist_index.html").getPath(),
                 Utility.UTF8,
                 NAMESLIST_DIR + "index.html");
         Utility.copyTextFile(
-                Settings.SRC_UCA_DIR + "charts.css", Utility.LATIN1, NAMESLIST_DIR + "charts.css");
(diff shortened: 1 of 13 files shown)


markusicu:
https://github.com/unicode-org/unicodetools/actions/runs/12241473884/job/34146639916?pr=985
```
Error:  Tests run: 4, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 44.12 s <<< FAILURE! - in org.unicode.text.UCD.TestTestUnicodeInvariants
Error:  org.unicode.text.UCD.TestTestUnicodeInvariants.testAdditionComparisons  Time elapsed: 0.001 s  <<< ERROR!
java.lang.NullPointerException
	at org.unicode.text.UCD.TestTestUnicodeInvariants.testAdditionComparisons(TestTestUnicodeInvariants.java:47)
```

eggrobin:
`/home/runner/work/unicodetools/unicodetools/unicodetools/mine/src/unicodetools/nullorg/unicode/text/UCD/AdditionComparisons does not exist`

Well, I can believe that. But why did renaming this file cause `src/main/resources/` to turn into `null`??

eggrobin:
@markusicu if this were C++ I would say « static initialization order fiasco », but I don’t know how this works in Java. Are you aware of similar pitfall there? It sure looks like UnicodeTools.UNICODETOOLS_RSRC_DIR is not what it should be when it is used to initialize Settings.SRC_DIR, even though it eventually seems to hold the right value.

markusicu:
Java basically allows circular dependencies among classes, which can cause class initialization trouble.
AFAIK: As soon as a class is first used, it is loaded, running its class initialization code, and initializing fields top to bottom (I believe literally). But that can use other classes, and then those get initialized. If they in turn use the first class, then its initialization is incomplete, and some fields are still undefined/null.

@echeran @macchiati @josh-hadley for more ideas?

PR #988, opened 2024-12-13
Greedy context before in LB20a

eggrobin:
Alternatives are left-to-right, not greedy: In the string
\<U+2E09 LEFT TRANSPOSITION BRACKET, U+2E0D RIGHT RAISED OMISSION BRACKET, U+1F3FC EMOJI MODIFIER FITZPATRICK TYPE-3>,
Searching for `( ^ | [^[\p{ea=F}\p{ea=W}\p{ea=H}]] ) \p{Line_Break=Quotation}` will find a match consisting of the first character, whereas searching for `( [^[\p{ea=F}\p{ea=W}\p{ea=H}]] | ^ ) \p{Line_Break=Quotation}` will find a match consisting of the first two characters.

ICU depends on a maximal match in its generated old monkeys: https://github.com/unicode-org/icu/blob/02951053b45f270c04ffeac6e0787dd2313bbd23/icu4c/source/test/intltest/rbbitst.cpp#L1779. This is messy to solve in general, but our alternatives here are simple enough that we can make them greedy by moving sot at the end.
Specifically, this works because except for sot, these alternatives are really unions of character classes. We could prohibit alternatives altogether and use unions of character classes, splitting the sot case into its own sub-rule, but that would be a divergence from the UAX14 formulation, which seems worse.

### DIFF
--- a/unicodetools/src/main/resources/org/unicode/tools/SegmenterDefault.txt
+++ b/unicodetools/src/main/resources/org/unicode/tools/SegmenterDefault.txt
@@ -183,7 +183,7 @@ $NS=[$NSorig $CJ]
 # LB 15a Do not break after an unresolved initial punctuation that lies at the start of the line,
 # after a space, after opening punctuation, or after an unresolved quotation mark, even after
 # spaces.
-15.11) ( $sot | $BK | $CR | $LF | $NL | $OP | $QU | $GL | $SP | $ZW ) $QU_Pi $SP* ×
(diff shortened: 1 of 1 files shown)


eggrobin:
I think I accidentally did this as part of https://github.com/unicode-org/unicodetools/pull/1006 as well?

But it tells me it doesn’t conflict, so let’s merge it I guess, that way we’ll have the explanation somewhere in the Merkle tree…

markusicu (approved):

PR #990, opened 2024-12-16
remove CollationTest.html from data folder

markusicu:
- [[181-A148](https://www.unicode.org/cgi-bin/GetL2Ref.pl?181-A148)] Action Item for Markus Scherer, PAG: Merge the contents of CollationTest.html into UTS10 and omit CollationTest.html from /Public/UCA/. For Unicode Version 17.0. See [L2/24-224](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/24-224) item 7.1.
- which is from PAG issue https://github.com/unicode-org/properties/issues/324

Moving into the spec in:
- https://github.com/unicode-org/unicode-reports/pull/184


eggrobin (approved):

srl295:
Late to the party here, but CLDR points to this .html file.
Should <https://www.unicode.org/Public/UCA/latest/CollationTest.html> be a redirect somewhere? Maybe to <https://www.unicode.org/reports/tr10/#Conformance_Tests>

PR #993, opened 2024-12-20
Test that emoji are emoji

eggrobin:
Now passes thanks to https://github.com/unicode-org/unicodetools/pull/996.


nedley:
It feels like the first emoji build happens sooner every year… GenerateEmoji.java bootstraps itself from candidateData.txt but I haven’t done that yet. Partly I was waiting for Chris Chapman to watch over my shoulder but I will try to get on this sooner rather than later. Sorry for the inconvenience!

nedley (approved):

PR #996, opened 2025-01-08
Emoji 17.0 alpha data

nedley:
Add new emoji as draft candidates and rebuild data files. Known issues:
1. ~~proposalData.txt has not been updated yet, there is an issue with the naming of the new group sequences that I haven’t tracked down yet.~~
2. ~~internal/emoji-proposals.txt thinks new group sequences are missing proposals, likely another manifestation of the previous issue.~~
3. ~~`MAP_TO_COUPLES` is using MAN rather than ADULT, maybe this is the same issue yet again?~~
4. emoji-released.html (not part of this commit) has two Sample columns, probably due to how I handled the Sark flag for 16.0.
5. New wrestling sequences have a claimed age of E3.0, which is interesting because the bunny ears sequences are fine.
6. ~~Ordering is not final but should be nailed down by the upcoming UTC.~~


unicodetools/src/main/java/org/unicode/tools/emoji/CountEmoji.java
```diff
@@ -382,17 +382,23 @@ enum Attribute {
         zwj("Ⓩ"),
         gender(Emoji.FEMALE),
         role(Emoji.WOMAN_STR),
-        family(Emoji.NEUTRAL_FAMILY),
+        family(Emoji.NEUTRAL_FAMILY, "group"),
```
nedley, line 385: This is a minor fix to chart tooltips, which currently call any multi-person group a family.

markusicu (approved):
rslgtm

nedley:
Issue 1 is because a generated sequence looks like "1F468 1F3FF 200D 1F430 200D 1F468 1F3FF" but that should have been normalized to "1F46F 1F3FF".

nedley:
Force-pushed a fix for the first three issues I noted.

unicodetools/src/main/java/org/unicode/tools/emoji/ProposalData.java
```diff
@@ -317,6 +317,12 @@ public static Set<String> cleanProposalString(String proposalString) {
             ImmutableMap.<String, String>builder()
                     .put("🧑🏿‍❤️‍💋‍🧑🏿", "💏🏿")
                     .put("🧑🏿‍❤️‍🧑🏿", "💑🏿")
+                    .put("👨🏿‍🐰‍👨🏿", "👯🏿")
```
nedley, line 320: I’m still not entirely sure what distinguishes these from handshakes, but it works for now.

nedley:
One more force-push to fix ordering. Any additional changes for the remaining issues will be handled in a new PR.

eggrobin (approved):

macchiati (approved):

PR #998, opened 2025-01-16
Reorder a few emoji based on ESR discussion

nedley:


cjchapman:
@nedley Looks like you got some security data invariant test failures.

eggrobin:
Security data invariant test failures are expected until the start of β review (note that this check fails at main), as we prepare that data only for β.

eggrobin (approved):

Issue #1007, opened 2025-01-28
feat: publish the JSP docker image via github


srl295:
https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry

we already build the docker image here

https://github.com/unicode-org/unicodetools/blob/main/.github/workflows/build-jsp.yml#L76-L77

so just need to push it to the container registry

srl295:
this could in the meantime give people another way to easily run the latest locally per #1004

PR #1008, opened 2025-01-28
jsp: publish Docker Image to Github Packages on tag

srl295:
Fixes: #1007 

- automatically, on tag (or release), push a Docker image to Github packages (see: https://github.com/srl295/unicodetools/pkgs/container/unicodetools/345094535?tag=5.4.3.2.1-srltest )
- change the JSP build to be on-PR instead of on-push
- also updates some action versions

### DIFF
--- a/.github/workflows/build-jsp.yml
+++ b/.github/workflows/build-jsp.yml
@@ -22,7 +22,7 @@ jobs:
       attestations: write
       id-token: write
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
     - name: Get the CLDR_REF from pom.xml
       id: cldr_ref
       run: echo "CLDR_REF="$(mvn help:evaluate -Dexpression=cldr.version -q -DforceStdout | cut -d- -f3) >> $GITHUB_OUTPUT && cat ${GITHUB_OUTPUT}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-jsp.yml
```diff
@@ -73,6 +80,32 @@ jobs:
       with:
         name: UnicodeJsps
         path: UnicodeJsps/target/UnicodeJsps.war
-    - name: build docker image
-      run: cd UnicodeJsps && bash update-bidic-ucd.sh && docker build .
-
+    - name: Build C Bidi
```
eggrobin, line 83: I think one step is missing, namely to get the Last Resort font:
https://github.com/unicode-org/unicodetools/blob/27875cb669668c0e42ea6866080f78c1566fe6e5/docs/unicodejsps/index.md#running-a-docker-based-build

eggrobin:
How do you then get the image from GitHub packages to gcp?

srl295:
@eggrobin you could pull them down and then push them. But the point here is that someone could run the packages locally.

.github/workflows/build-jsp.yml
eggrobin: https://github.com/unicode-org/unicodetools/issues/1010.

eggrobin (approved):

eggrobin (approved):

PR #1018, opened 2025-01-29
Bring back the JSPs

eggrobin:
Four commits, deployed in three builds of the JSPs:
1. Drop properties.jsp (we should bring something like it back eventually, but besides its abysmal performance it had not been seriously maintained and was becoming borderline useless);
2. Remove links to UnicodeSet queries for multivalued properties from character.jsp (these queries are very slow as currently implemented—minutes for some of the worst offenders—, and the link was to an invalid query when the character being inspected actually had multiple values anyway);
3.
   * Remove the ICU-based toLowercase etc. pseudoproperties (these are from before the JSPs showed the real Lowercase_Mapping etc. properties, and UnicodeSet queries to compute the preimage of a function are intrinsically very slow);
   * Fail multivalued Unihan property queries if they are in the exact format we used to link to, to deal with residual traffic from links we served before deployment 2., with a suggestion to use a different format in case a hapless human comes across this (this one can hopefully be reverted eventually).

image
image

### DIFF
--- a/UnicodeJsps/src/main/java/org/unicode/jsp/XPropertyFactory.java
+++ b/UnicodeJsps/src/main/java/org/unicode/jsp/XPropertyFactory.java
@@ -145,48 +145,6 @@ public class XPropertyFactory extends UnicodeProperty.Factory {
                                 },
                                 false)
                         .setMain("toNFKD", "toNFKD", UnicodeProperty.STRING, "1.1"));
-
-        add(
-                new StringTransformProperty(
-                                new StringTransform() {
-                                    @Override
-                                    public String transform(String source) {
-                                        return UCharacter.foldCase(source, true);
-                                    }
-                                },
-                                false)
-                        .setMain("toCasefold", "toCF", UnicodeProperty.STRING, "1.1"));
-        add(
-                new StringTransformProperty(
-                                new StringTransform() {
-                                    @Override
-                                    public String transform(String source) {
-                                        return UCharacter.toLowerCase(ULocale.ROOT, source);
-                                    }
-                                },
-                                false)
-                        .setMain("toLowercase", "toLC", UnicodeProperty.STRING, "1.1"));
-        add(
-                new StringTransformProperty(
-                                new StringTransform() {
-                                    @Override
-                                    public String transform(String source) {
-                                        return UCharacter.toUpperCase(ULocale.ROOT, source);
-                                    }
-                                },
-                                false)
-                        .setMain("toUppercase", "toUC", UnicodeProperty.STRING, "1.1"));
-        add(
-                new StringTransformProperty(
-                                new StringTransform() {
-                                    @Override
-                                    public String transform(String source) {
-                                        return UCharacter.toTitleCase(ULocale.ROOT, source, null);
-                                    }
-                                },
-                                false)
-                        .setMain("toTitlecase", "toTC", UnicodeProperty.STRING, "1.1"));
-
         add(
                 new StringTransformProperty(
                                 new StringTransform() {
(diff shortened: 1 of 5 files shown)


eggrobin:
> residual traffic from links we served before deployment 2.

This is strangely still a thing, see, e.g., for kJapaneseKun over the past 24 hours (in 15 minute buckets):
image

I guess some crawlers saved those links somewhere…

UnicodeJsps/src/main/java/org/unicode/jsp/XPropertyFactory.java
```diff
@@ -145,48 +145,6 @@ public String transform(Integer source) {
                                 },
                                 false)
                         .setMain("toNFKD", "toNFKD", UnicodeProperty.STRING, "1.1"));
-
```
macchiati, line 148: I find these very useful in practice. Is there a compelling reason to remove?

UnicodeJsps/src/main/webapp/properties.jsp
```diff
@@ -20,7 +20,8 @@ th           { text-align: left }
         UtfParameters utfParameters = new UtfParameters(queryString);
 
         String propForValues = utfParameters.getParameter("a");
-		UnicodeJsp.showPropsTable(out, propForValues, "properties.jsp");
+        // TODO(egg): Cache this page.
```
macchiati, line 23: Ok for now. I think just a static page with an explicit list of the properties (and if enumerated, the values) without links would be very helpful for the users, and not hard to generate at build time.

macchiati (changes requested):

UnicodeJsps/src/main/java/org/unicode/jsp/XPropertyFactory.java
eggrobin: Well, they are very slow (getSet involves computing the preimage of a function).

But what do you use these for, as opposed to Lowercase_Mapping, Uppercase_Mapping, and Titlecase_Mapping?

eggrobin: I just checked (on the staging JSPs which do not yet have these changes), and both of the following contain only characters whose mappings changed in 16.0 (because we are lagging on the ICU side used in the right-hand side):
* [:^Lowercase_Mapping=@toLowercase@:]
* [:^Uppercase_Mapping=@toUppercase@:]

[:^Titlecase_Mapping=@toTitlecase@:] contains one interesting difference, U+0345, because ICU skips past a leading U+0345 in titlecasing. But I think for all practical purposes, you can just use the mappings.

UnicodeJsps/src/main/webapp/properties.jsp
eggrobin: Yes. And we should probably grab some information from IndexUnicodeProperties as well, e.g., which files they come from.

But not today.

macchiati: Agreed

UnicodeJsps/src/main/java/org/unicode/jsp/XPropertyFactory.java
macchiati: Ok, makes sense.

macchiati (approved):

macchiati:
BTW, I tried converting 

- a UnicodeMap<X> to a construct that used TreeMap<Integer, X> + internally generated a Map<X,UnicodeSet> once built.
- a UnicodeRelation<X> to a construct that used TreeMultimap<Integer, X> in building + internally generated a Map<X,UnicodeSet> once built.

At the expense of additional memory, it sped up the queries significantly.

eggrobin:
> Fail multivalued Unihan property queries if they are in the exact format we used to link to, to deal with residual traffic from links we served before deployment 2., with a suggestion to use a different format in case a hapless human comes across this (this one can hopefully be reverted eventually).

Wow, we are still getting about a hundred queries per day for those, someone must have saved all the links we used to serve and keeps crawling them…

PR #1026, opened 2025-01-30
Parse pre-5.2 Unihan

eggrobin:
And actually carry 3.2 Unihan rather than 3.1.1 in the 3.2 and 4.0 directories (there was no 4.0 Unihan).

With this, the reason for the code in https://github.com/unicode-org/unicodetools/pull/1022/files#diff-68c25d5c3f8106eeba887e1dbb453ed8275f660468e68f8d061e447a345b19d2 becomes clear on the U+4EAC and U+5793 pages:

image
image

See also **[[94-C4](http://www.unicode.org/cgi-bin/GetL2Ref.pl?94-C4)] Consensus:** 1) Incorporate the numeric values from the Unihan file without adding a new property (using numeric_type), 2)  remove the han numeric property from U+5793 and U+4EAC, and 3) drop source standards as a property. [[L2/03-094](http://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/03-094)]

CC @Ken-Whistler who counted the zeros on the modern kPrimaryNumeric value of U+4EAC at UTC-182.


eggrobin:
> [Check UCD consistency, invariants, smoke-test generators](https://github.com/unicode-org/unicodetools/actions/runs/13059235609/job/36437842467?pr=1026#logs)
> failed 1 hour ago in 0s
> This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v3`

Time to mindlessly bump some version numbers I guess…

unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
```diff
@@ -839,10 +842,20 @@ private static void parsePropertyValueFile(
                                     + " in "
                                     + indexUnicodeProperties.ucdVersion);
                 }
+                // Unihan 4.0 and earlier implemented multivalued properties by repeating the
+                // property value record instead of using a delimiter.
+                var merger =
+                        propInfo.property.getShortName().startsWith("cjk")
+                                        && indexUnicodeProperties.ucdVersion.compareTo(
+                                                        VersionInfo.UNICODE_4_0)
```
markusicu, line 850: nit: testing `... .ucdVersion.getMajor() <= 4` seems more readable

unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
```diff
@@ -851,6 +851,13 @@ private static void parsePropertyValueFile(
                                                 <= 0
                                 ? new PropertyUtilities.Joiner("|")
                                 : null;
+                final var originalMultivaluedSplit = propInfo.multivaluedSplit;
+                // The first version of kPrimaryNumeric had spaces in values.
+                if (propName.equals("kPrimaryNumeric")
+                        && indexUnicodeProperties.ucdVersion.compareTo(VersionInfo.UNICODE_4_0)
```
markusicu, line 857: nit: second check for ucdVersion <= 4.0 in the same block...

markusicu (approved):

unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
eggrobin: More readable, less correct, since it would include Unicode 4.0.1 and 4.1.0.

unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
eggrobin: Yes, but it doesn’t factor nicely since I am declaring things.

(Also, morally that second one is == 3.2, since Unicode 4.0 did not update Unihan, whereas the other one really covers multiple versions.)

PR #1027, opened 2025-01-30
ヨリ

eggrobin:
From CJK.
* **[[182-C31](https://www.unicode.org/cgi-bin/GetL2Ref.pl?182-C31)] Consensus:** Provisionally assign U+1B126 KATAKANA DIGRAPH YORI in the Kana Extended-A block, based on document [L2/24-279](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/24-279) (Sim) and Section 14 of document [L2/25-009](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/25-009).
* https://github.com/unicode-org/utc-release-management/issues/190


markusicu:
CLA check is happy now

PR #1030, opened 2025-02-07
UCDXML and TR42 v2

jowilco:


unicodetools/src/main/java/org/unicode/xml/GeneratePropertyValues.java
```diff
@@ -123,7 +124,7 @@ public static void main(String[] args) throws Exception {
 
         if (options[HELP].doesOccur) {
             System.out.println(
-                    "GeneratePropertyValuesList --ucdversion {version number} --outputfolder {destination}");
+                    "GeneratePropertyValuesList [--ucdversion {version number}] [--outputfolder {destination}]");
```
markusicu, line 127: you added brackets to the outputfolder, and you no longer throw an exception if it's absent, but it still REQUIRES_ARG

unicodetools/src/main/java/org/unicode/xml/GeneratePropertyValues.java
```diff
@@ -154,8 +154,19 @@ public static void main(String[] args) throws Exception {
                             "Could not find or create " + options[OUTPUTFOLDER].value);
                 }
             } else {
-                throw new IllegalArgumentException(
-                        "Missing command line option: --outputfolder (or -f)");
+                try {
+                    destinationFolder = new File(Settings.Output.GEN_DIR + "uax42\\fragments\\");
```
markusicu, line 158: let's please use forward slashes, so that this works on non-Windows machines, or use Java APIs to build the path from pieces; Settings.java should have some useful code to look at

unicodetools/src/main/java/org/unicode/xml/GeneratePropertyValues.java
```diff
@@ -154,8 +154,19 @@ public static void main(String[] args) throws Exception {
...
+                    if (!destinationFolder.exists()) {
```
markusicu, line 159: duplicate code -- I would try to get the destinationFolder into a variable whether it's from the command line or the default, and then use the same code to check exists(), call mkdirs(), etc.

unicodetools/src/main/java/org/unicode/xml/UCDXML.java
```diff
@@ -152,8 +152,26 @@ public static void main(String[] args) throws Exception {
...
-                throw new IllegalArgumentException(
-                        "Missing command line option: --outputfolder (or -f)");
+                try {
+                    destinationFolder =
+                            new File(
+                                    Settings.Output.GEN_DIR
+                                            + "ucdxml\\"
+                                            + ucdVersion.getVersionString(3, 3)
+                                            + "\\");
+                    if (!destinationFolder.exists()) {
```
markusicu, line 162: ditto duplication

markusicu (commented):
I have not looked at the whole thing again now, and I think I won't.
- I have looked at your replies that I hadn't seen before
- I have looked at your changes other than rebuilding the branch and satisfying CI
- I caught up on emails that I hadn't read before
- I have read your updated readme

Here I have a couple of small comments on recent changes. After this I am inclined to unpack a big rubber stamp :-)

unicodetools/src/main/java/org/unicode/xml/GeneratePropertyValues.java
jowilco: Fixed. (changing to forward slashes was sufficient)

unicodetools/src/main/java/org/unicode/xml/GeneratePropertyValues.java
jowilco: Fixed. And the original was pretty lazy on my part...

markusicu (approved):
> After this I am inclined to unpack a big rubber stamp :-)

... and here is... **the big rubber stamp**.

Thank you!!

eggrobin (approved):
# \*STAMP\*

PR #1033, opened 2025-02-10
Initial teaching alphabet

eggrobin:
**[[182-C6](https://www.unicode.org/cgi-bin/GetL2Ref.pl?182-C6)] Consensus:** Provisionally assign 13 code points U+1DF5A..U+1DF66 to Initial Teaching Alphabet characters with names and code points as described in the “Properties” section of [L2/24-273](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/24-273). [Ref. 1.5 in [L2/25-010](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/25-010)]

* unicode-org/sah#224
* https://github.com/unicode-org/utc-release-management/issues/177


kirkrmiller:
Consideration might be given as to whether U+1DF5D (the i-e ligature) and 1DF5E (the o-i ligature) should be made soft-dotted. This was not requested in the proposal because the ITA does not use diacritics. However, if these characters are ever used for other things, combining diacritics might be relevant.

roozbehp:
> Consideration might be given as to whether [U+1DF5D](https://util.unicode.org/UnicodeJsps/character.jsp?a=1DF5D) (the i-e ligature) and 1DF5E (the o-i ligature) should be made soft-dotted. This was not requested in the proposal because the ITA does not use diacritics. However, if these characters are ever used for other things, combining diacritics might be relevant.

I believe such a decision would definitely need explicit UTC approval. We don't have any soft-dotted digraphs/ligatures in the standard as of now (so we can't assign them Soft_Dotted based on precedent), and there is no evidence that these would lose their dots with combining marks.

PR #1042, opened 2025-02-13
Update the pipeline documentation to cover changes in process

eggrobin:
We now have comparison tests, summaries in the reports, and some dashboard status wrangling as part of this exercise.


markusicu (approved):
lgtm pse squash

PR #1043, opened 2025-02-13
Centralize and fix age comparison

### LINKED ISSUE #54: "age" filter for "list-unicodeset" utility does not work correctly for values >= 10
@srutzky wrote:

<sup>(I originally posted this in JIRA as [CLDR-14700](https://unicode-org.atlassian.net/browse/CLDR-14700) but was told that this was a more appropriate forum for reporting this type of problem)</sup>

This concerns the *Unicode Utilities: UnicodeSet* utility, located at:
https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp

Filtering on the "age" property does not work correctly. The issue started a few years ago, after version 10.0 was released (and this utility was updated for that version). Versions starting at 10.0 do not behave correctly, and are somehow "linked" (for lack of a better term) to version 1.1. The examples below should illustrate the issue.

&nbsp;
#### Correct, but only when "=1.1"
https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AAge%3D1.1%3A%5D&g=age&i=
```
[:Age=1.1:]
33,977 Code Points
```

&nbsp;
#### A value of "=10.0" or higher returns 1.1 _and_ version 10.0 through the specified version
But, _not_ versions 2.0 through 9.0 (which should be returned). For example, specifying version 12.0 will return versions:
* 1.1
* 10.0
* 11.0
* 12.0

https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AAge%3D12.0%3A%5D&g=age&i=
```
[:Age=12.0:]
43,733 Code Points
```

&nbsp;
#### A value of "=2.0" through "=9.0" returns 1.1 through the specified version _and_ versions 10.0 and higher
(versions 10.0 and higher should not be returned). For example, specifying version 3.1 will return versions:
* 1.1
* 2.0
* 2.1
* 3.0
* 3.1
* 10.0
* 11.0
* 12.0

https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AAge%3D3.1%3A%5D&g=age&i=
```
[:Age=3.1:]
243,477 Code Points
```

&nbsp;
#### A value of "=10.0" or higher _and_ "\!=1.1" returns version 10.0 through the specified version
But, _not_ versions 2.0 through 9.0 (which should be returned). For example, specifying version 11.0 _and_ \!1.1 will return versions:
* 10.0
* 11.0

https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AAge%3D11.0%3A%5D+%26+%5B%3AAge%21%3D1.1%3A%5D&g=age&i=
```
[:Age=11.0:] & [:Age!=1.1:]
9,202 Code Points
```

&nbsp;
#### A value of "=10.0" or higher _and_ "\!=2.0" through "\!=9.0" returns nothing / empty-set
(one version higher than the specified "\!=" version through specified "=" version should be returned). For example, specifying version 11.0 _and_ \!2.0 will return versions:
* (none / nada / zip / zilch )
* (however, versions 2.1 through 11.0 _should_ be returned)
* (specifying version 9.0 and !2.0 does return the correct results)

https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AAge%3D11.0%3A%5D+%26+%5B%3AAge%21%3D2.0%3A%5D&g=age&i=
```
[:Age=11.0:] & [:Age!=2.0:]
0 Code Points
```

https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AAge%3D9.0%3A%5D+%26+%5B%3AAge%21%3D2.0%3A%5D&g=age&i= 
```
[:Age=9.0:] & [:Age!=2.0:]
89,287 Code Points
```

eggrobin:
Fix #54.

### DIFF
--- a/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
+++ b/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
@@ -12,9 +12,9 @@ import java.util.Comparator;
 import java.util.List;
 import java.util.regex.Pattern;
 import org.unicode.cldr.util.MultiComparator;
-import org.unicode.jsp.UnicodeSetUtilities.ComparisonMatcher.Relation;
 import org.unicode.props.UnicodeProperty;
 import org.unicode.props.UnicodeProperty.PatternMatcher;
+import org.unicode.props.UnicodePropertySymbolTable;
 
 public class UnicodeSetUtilities {
 
@@ -292,7 +292,13 @@ public class UnicodeSetUtilities {
                                         + prop.getValueAliases());
                     }
                     if (isAge) {
-                        set = prop.getSet(new ComparisonMatcher(propertyValue, Relation.geq));
+                        set =
+                                prop.getSet(
+                                        new UnicodePropertySymbolTable.ComparisonMatcher(
+                                                propertyValue,
+                                                UnicodePropertySymbolTable.Relation.geq,
+                                                UnicodePropertySymbolTable
+                                                        .VERSION_STRING_COMPARATOR));
                     } else {
                         if (prop.getName().equals("General_Category")) {
                             for (String[] coarseValue : COARSE_GENERAL_CATEGORIES) {
@@ -342,51 +348,6 @@ public class UnicodeSetUtilities {
             //        if (propertyValue)
             //      }
             return prop.isValidValue(propertyValue);
-        }
-    }
-    ;
-
-    public static class ComparisonMatcher implements PatternMatcher {
-        Relation relation;
-
-        enum Relation {
-            less,
-            leq,
-            equal,
-            geq,
-            greater
-        }
-
-        static Comparator comparator = new UTF16.StringComparator(true, false, 0);
-
-        String pattern;
-
-        public ComparisonMatcher(String pattern, Relation comparator) {
-            this.relation = comparator;
-            this.pattern = pattern;
-        }
-
-        @Override
-        public boolean test(String value) {
-            int comp = comparator.compare(pattern, value);
-            switch (relation) {
-                case less:
-                    return comp < 0;
-                case leq:
-                    return comp <= 0;
-                default:
-                    return comp == 0;
-                case geq:
-                    return comp >= 0;
-                case greater:
-                    return comp > 0;
-            }
-        }
-
-        @Override
-        public PatternMatcher set(String pattern) {
-            this.pattern = pattern;
-            return this;
         }
     }
 }
--- a/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
+++ b/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
@@ -301,9 +301,7 @@ public class UnicodeSetUtilities {
                                                         propertyValue),
                                                 UnicodePropertySymbolTable.Relation.geq,
                                                 Comparator.nullsFirst(Comparator.naturalOrder()),
-                                                (s) ->
-                                                        UnicodePropertySymbolTable
-                                                                .parseVersionInfoOrMax(s)));
+                                                UnicodePropertySymbolTable::parseVersionInfoOrMax));
                     } else {
                         if (prop.getName().equals("General_Category")) {
                             for (String[] coarseValue : COARSE_GENERAL_CATEGORIES) {
--- a/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
+++ b/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
@@ -7,6 +7,7 @@ import com.ibm.icu.text.UTF16;
 import com.ibm.icu.text.UTF16.StringComparator;
 import com.ibm.icu.text.UnicodeSet;
 import com.ibm.icu.util.ULocale;
+import com.ibm.icu.util.VersionInfo;
 import java.text.ParsePosition;
 import java.util.Comparator;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


unicodetools/src/main/java/org/unicode/props/UnicodePropertySymbolTable.java
```diff
@@ -281,8 +279,11 @@ public PatternMatcher set(String pattern) {
...
     }
 
-    /** Special parser for doubles. Anything not parsable is higher than everything else. */
-    public static final Comparator<String> DOUBLE_STRING_COMPARATOR =
+    /**
+     * Special parser for version strings. Anything not parsable is higher than everything
+     * parseable.
+     */
+    public static final Comparator<String> VERSION_STRING_COMPARATOR =
             new Comparator<String>() {
```
macchiati, line 287: Much cleaner, thanks.

BTW, when writing new code (or rewriting old code), I like using https://guava.dev/releases/21.0/api/docs/com/google/common/collect/ComparisonChain.html

unicodetools/src/test/java/org/unicode/unittest/TestUnicodeSet.java
```diff
@@ -33,11 +33,11 @@ private void checkOrder(String d1, String d2, int expected) {
         assertEquals(
                 d1 + " ?< " + d2,
                 expected,
-                UnicodePropertySymbolTable.DOUBLE_STRING_COMPARATOR.compare(d1, d2));
+                UnicodePropertySymbolTable.VERSION_STRING_COMPARATOR.compare(d1, d2));
         assertEquals(
                 d2 + " ?< " + d1,
                 -expected,
-                UnicodePropertySymbolTable.DOUBLE_STRING_COMPARATOR.compare(d2, d1));
+                UnicodePropertySymbolTable.VERSION_STRING_COMPARATOR.compare(d2, d1));
```
markusicu, line 40: assertEquals() is now stricter than the actual return value from the comparator. It expects the return values to be pinned to -1/0/+1 but compareTo() can return larger negative/positive values.

```
TestUnicodeSet.TestAge:16->checkOrder:33->TestFmwkMinusMinus.assertEquals:50 3.1 ?< 3.2 ==> expected: <-1> but was: <-32768>
```

I suggest changing the test code to extract the "sign" of the return value.
Alternatively, you could change the comparator to return only the sign value.

unicodetools/src/main/java/org/unicode/props/UnicodePropertySymbolTable.java
```diff
@@ -202,7 +202,9 @@ public boolean applyPropertyAlias0(
                     set =
                             prop.getSet(
                                     new ComparisonMatcher(
-                                            propertyValue, Relation.geq, DOUBLE_STRING_COMPARATOR));
+                                            propertyValue,
+                                            Relation.geq,
+                                            VERSION_STRING_COMPARATOR));
```
markusicu, line 207: performance issue: the propertyValue String (too bad we still always store strings...) is parsed in each comparison. would it be reasonably easy to make a variant of a ComparisonMatcher -- a VersionMatcher -- which parses the input propertyValue only once on construction, and parses each other value before comparing the VersionInfo's?

we could use a special, static final VersionInfo with, say, 255.255.255.255 for when a string is not parsable.
and 0.0.0.0 for null strings.

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
```diff
@@ -294,11 +295,15 @@ private boolean applyPropertyAlias0(
...
+                                                VersionInfo>(
+                                                UnicodePropertySymbolTable.parseVersionInfoOrMax(
+                                                        propertyValue),
                                                 UnicodePropertySymbolTable.Relation.geq,
-                                                UnicodePropertySymbolTable
-                                                        .VERSION_STRING_COMPARATOR));
+                                                Comparator.nullsFirst(Comparator.naturalOrder()),
+                                                (s) ->
+                                                        UnicodePropertySymbolTable
+                                                                .parseVersionInfoOrMax(s)));
```
markusicu, line 306: Can we write this as just `UnicodePropertySymbolTable::parseVersionInfoOrMax` ?

unicodetools/src/main/java/org/unicode/props/UnicodePropertySymbolTable.java
```diff
@@ -201,10 +202,14 @@ public boolean applyPropertyAlias0(
...
-                                            propertyValue,
+                                    new ComparisonMatcher<VersionInfo>(
+                                            UnicodePropertySymbolTable.parseVersionInfoOrMax(
+                                                    propertyValue),
                                             Relation.geq,
-                                            VERSION_STRING_COMPARATOR));
+                                            Comparator.nullsFirst(Comparator.naturalOrder()),
+                                            (s) ->
+                                                    UnicodePropertySymbolTable
+                                                            .parseVersionInfoOrMax(s)));
```
markusicu, line 212: ditto?

unicodetools/src/main/java/org/unicode/props/UnicodePropertySymbolTable.java
```diff
@@ -244,20 +249,26 @@ public enum Relation {
...
     }
 
-    public static class ComparisonMatcher implements PatternMatcher {
+    public static class ComparisonMatcher<T> implements PatternMatcher {
         final Relation relation;
-        final Comparator<String> comparator;
-        String pattern;
+        final Comparator<T> comparator;
+        final Function<String, T> parser;
+        T expected;
```
markusicu, line 256: FYI: nice!

markusicu (dismissed):

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
eggrobin: I tried something like that, but I had forgotten the syntax…

Someday perhaps I should learn Java.

markusicu (approved):

PR #1057, opened 2025-03-11
Don’t invert twice on comparison queries

eggrobin:
\p{Uppercase≠@Changes_When_Lowercased@} is currently equal to \p{Uppercase=@Changes_When_Lowercased@} rather than  \P{Uppercase=@Changes_When_Lowercased@}, because it is re-inverted before it is returned.

Also added support for null queries and align identity queries with the working draft (`code point` like in `@missing` lines rather than `cp`, and LM3 matching), and brought LM3 matching up to date with Unicode 6.0, stripping leading `is` (see also https://www.unicode.org/reports/tr44/tr44-6.html#UAX44-LM3, the proposed update https://www.unicode.org/reports/tr44/tr44-5.html#UAX44-LM3, UTC-120-A106, and L2/09-248).

### DIFF
--- a/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
+++ b/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
@@ -264,6 +264,8 @@ public class UnicodeSetUtilities {
                     otherProperty = factory.getProperty(otherPropName);
                 }
             }
+            // TODO(egg): Name and Name_Alias require special handling (UAX44-LM2), and
+            // treating Name_Alias as aliases for Name.
             boolean isAge = UnicodeProperty.equalNames("age", propertyName);
             if (prop != null) {
                 UnicodeSet set;
--- a/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
+++ b/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
@@ -250,13 +250,16 @@ public class UnicodeSetUtilities {
             }
             UnicodeProperty otherProperty = null;
             boolean testCp = false;
+            boolean testNone = false;
             if (trimmedPropertyValue.length() > 1
                     && trimmedPropertyValue.startsWith("@")
                     && trimmedPropertyValue.endsWith("@")) {
                 String otherPropName =
                         trimmedPropertyValue.substring(1, trimmedPropertyValue.length() - 1).trim();
-                if ("cp".equalsIgnoreCase(otherPropName)) {
+                if (UnicodeProperty.equalNames("code point", otherPropName)) {
                     testCp = true;
+                } else if (UnicodeProperty.equalNames("none", otherPropName)) {
+                    testNone = true;
                 } else {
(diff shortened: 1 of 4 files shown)


UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
```diff
@@ -270,15 +273,20 @@ private boolean applyPropertyAlias0(
                         if (invert != UnicodeProperty.equals(i, prop.getValue(i))) {
                             set.add(i);
                         }
+                        invert = false;
```
markusicu, line 278: please move this to after the loop. surely we need not do this 1.1M times...

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
```diff
@@ -270,15 +273,20 @@ private boolean applyPropertyAlias0(
...
                 } else if (otherProperty != null) {
+                    System.err.println(otherProperty + ", " + invert);
                     set = new UnicodeSet();
                     for (int i = 0; i <= 0x10FFFF; ++i) {
                         String v1 = prop.getValue(i);
                         String v2 = otherProperty.getValue(i);
                         if (invert != UnicodeProperty.equals(v1, v2)) {
                             set.add(i);
                         }
+                        invert = false;
```
markusicu, line 291: ditto

UnicodeJsps/src/test/java/org/unicode/jsptest/TestUnicodeSet.java
```diff
@@ -161,7 +161,7 @@ public void TestIdentityQuery() {
 
     @Test
     public void TestNullQuery() {
-        checkSetsEqual("\\p{Bidi_Paired_Bracket=@none@}", "\\p{Bidi_Paired_Bracket_Type=None}");
+        checkSetsEqual("\\p{Bidi_Paired_Bracket=@none@}", "\\p{Bidi_Paired_Bracket_Type=Is_None}");
```
markusicu, line 164: It makes sense to me that a *binary property* would be prefixed with "is", but I wouldn't expect anything else -- non-binary property, or some value -- to work like that. Seems like gratuitous, confusing flexibility.

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
eggrobin: 😂 At least we are sure that it will be _very_ false !

UnicodeJsps/src/test/java/org/unicode/jsptest/TestUnicodeSet.java
eggrobin: > Seems like gratuitous, confusing flexibility.

I agree. But I think you need to travel to Redmond and to the year 2009 to make that case…

I suppose for the implicit properties script and gc, \p{isGreek} or \p{isLetter} still looks vaguely nice…

macchiati:
Looking forward to having this in the JSPs too!

PR #1059, opened 2025-03-12
LM2 Name and Name_Alias matching

eggrobin:
And treat Name_Alias as a Name alias in the JSPs.

### DIFF
--- a/UnicodeJsps/src/test/java/org/unicode/jsptest/TestUnicodeSet.java
+++ b/UnicodeJsps/src/test/java/org/unicode/jsptest/TestUnicodeSet.java
@@ -184,7 +184,14 @@ public class TestUnicodeSet extends TestFmwk2 {
         checkSetsEqual("\\p{Name=HANGUL JUNGSEONG OE}", "[\\u116C]");
         checkSetsEqual("\\p{Name=Hangul jungseong o-e}", "[\\u1180]");
         checkSetsEqual("\\p{Name=Hangul jungseong oe}", "[\\u116C]");
+        checkSetsEqual("\\p{Name=HANGUL JUNGSEONG O -E}", "[\\u1180]");
+        checkSetsEqual("\\p{Name= HANGUL JUNGSEONG O-E }", "[\\u1180]");
+        checkSetsEqual("\\p{Name=_HANGUL_JUNGSEONG_O-E_}", "[\\u1180]");
+        checkSetsEqual("\\p{Name=HANGUL JUNGSEONG O-EO}", "[\\u117F]");
+        checkSetsEqual("\\p{Name=HANGUL JUNGSEONG OE O}", "[\\u117F]");
+        checkSetsEqual("\\p{Name=HANGUL JUNGSEONG O -EO}", "[]");
         checkSetsEqual("\\p{Name=MARCHEN LETTER -A}", "[\\x{11C88}]");
+        checkSetsEqual("\\p{Name=MARCHEN_LETTER_-A}", "[\\x{11C88}]");
         checkSetsEqual("\\p{Name=MARCHEN LETTER A}", "[\\x{11C8F}]");
         checkSetsEqual("\\p{Name=TIBETAN MARK TSA -PHRU}", "[\\u0F39]");
         checkSetsEqual("\\p{Name=TIBETAN MARK TSA PHRU}", "[]");
--- a/UnicodeJsps/src/test/java/org/unicode/jsptest/TestUnicodeSet.java
+++ b/UnicodeJsps/src/test/java/org/unicode/jsptest/TestUnicodeSet.java
@@ -37,6 +37,8 @@ import org.unicode.jsp.UnicodeJsp;
 import org.unicode.jsp.UnicodeSetUtilities;
 import org.unicode.jsp.UnicodeUtilities;
 import org.unicode.jsp.XPropertyFactory;
+import org.unicode.props.IndexUnicodeProperties;
+import org.unicode.props.UcdProperty;
 import org.unicode.props.UnicodeProperty;
 
 public class TestUnicodeSet extends TestFmwk2 {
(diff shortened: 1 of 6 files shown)


markusicu:
It looks like some of your commits are from the previous PR. Do you need to rebase?

eggrobin:
> It looks like some of your commits are from the previous PR.

Yes, I worked on the two in parallel and merged.

> Do you need to rebase?

I’d rather not. I don’t think the individual commits are useful to the reviewer here anyway, probably best to just look at https://github.com/unicode-org/unicodetools/pull/1059/files.

unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
```diff
@@ -720,39 +724,90 @@ public static String toSkeleton(String source) {
...
+                    is1180 = true;
+                    for (int j = i + 2; j < source.length(); ++j) {
+                        if (source.charAt(j) != ' ' && source.charAt(j) != '_') {
+                            is1180 = false;
+                        }
+                    }
+                }
+                if (!medial || is1180) {
+                    if (validate) {
+                        System.out.println("****** EXCEPTION " + source);
```
markusicu, line 792: what is the point of printing this?

unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
```diff
@@ -720,39 +724,90 @@ public static String toSkeleton(String source) {
...
             char ch = source.charAt(i);
+            final char uppercase = Character.toUpperCase(ch);
+            if (validate && uppercase != ch) {
+                throw new IllegalArgumentException(
+                        "Illegal Name Char: U+" + Utility.hex(ch) + ", " + ch);
+            }
+            ch = uppercase;
             if (('0' <= ch && ch <= '9') || ('A' <= ch && ch <= 'Z') || ch == '<' || ch == '>') {
                 result.append(ch);
             } else if (ch == ' ') {
```
markusicu, line 761: optional: handle underscore here, too?

unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
eggrobin: I am not sure.
This function was only used in TestNameUniqueness, and that is the only place that now calls it with validate=true.

It looks like that test is not actually a test in the sense of the test framework, and the only thing it does is print things out.

… maybe we should have a non-manual test for https://www.unicode.org/policies/stability_policy.html#Name_Uniqueness? 😬

unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
eggrobin: Note the different behaviour with validate=true.

unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
markusicu: > I am not sure. This function was only used in TestNameUniqueness, and that is the only place that now calls it with validate=true.

Let's drop this.

> … maybe we should have a non-manual test for Name_Uniqueness? 😬

We might already have an issue for that.
I don't think it's a high priority because there is a name uniqueness check built into the chart production tooling, so another WG would catch problems.

unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
```diff
@@ -720,39 +724,90 @@ public static String toSkeleton(String source) {
...
+                if (0 == i || i == source.length() - 1) {
+                    medial = false; // Name-initial or name-final.
+                } else {
+                    final char preceding = Character.toUpperCase(source.charAt(i - 1));
+                    final char following = Character.toUpperCase(source.charAt(i + 1));
+                    medial =
+                            (('0' <= preceding && preceding <= '9')
+                                            || ('A' <= preceding && preceding <= 'Z'))
+                                    && (('0' <= following && following <= '9')
+                                            || ('A' <= following && following <= 'Z'));
```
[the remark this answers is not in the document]
eggrobin: It gets even less readable, because then spotless wraps the second line but not the first:
```
                            (('0' <= prev && prev <= '9') || ('A' <= prev && prev <= 'Z'))
                                    && (('0' <= next && next <= '9')
                                            || ('A' <= next && next <= 'Z'));
```

But maybe I could use Character.isLetterOrDigit instead of hand-rolling it.

unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java
```diff
@@ -724,6 +725,10 @@ class IndexUnicodeProperty extends UnicodeProperty.BaseProperty {
             }
         }
 
+        public IndexUnicodeProperties getFactory() {
+            return IndexUnicodeProperties.this;
```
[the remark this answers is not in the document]
eggrobin: Type mismatch: cannot convert from IndexUnicodeProperties.IndexUnicodeProperty to IndexUnicodeProperties

UnicodeJsps/src/test/java/org/unicode/jsptest/TestUnicodeSet.java
```diff
@@ -175,6 +175,53 @@ public void TestInteriorlyNegatedComparison() {
...
+        // The medial hyphen is only significant in HANGUL JUNGSEONG O-E, not in arbitrary O-E/OE.
+        checkSetsEqual("\\p{Name=twoemdash}", "⸺");
+        checkSetsEqual("\\p{Name=SeeNoEvil_Monkey}", "🙈");
+        checkSetsEqual("\\p{Name=BALLET S-H-O-E-S}", "🩰");
+        checkSetsEqual("[\\p{Name=LATIN SMALL LIGATURE O-E}uf]", "[œuf]");
+    }
+
+    @Test
+    public void TestNameAliases() {
+        // Name_Alias values behave as aliases for Name, but not vice-versa.
```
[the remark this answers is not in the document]
eggrobin: > Does it make sense to support Name_Alias as a "public property"?

It is in https://www.unicode.org/Public/UCD/latest/ucd/PropertyAliases.txt…

More pragmatically, I think it is useful to the Unicode maintainer to have it accessible in UnicodeSet-the-query-language. (I do not think it is useful to the ICU user, and I do not think ICU should add support for it in its UnicodeSet implementation).

If you are trying to write the text below https://www.unicode.org/reports/tr44/#UAX44-LM2, you might want to be able to query explicitly the aliases in order to that
> Examples of medial hyphens in character name aliases include:
> * U+008E SINGLE-SHIFT-2
> * U+11EC HANGUL JONGSEONG YESIEUNG-KIYEOK
>
> Examples of non-medial hyphens in character name aliases include:
> * U+0FD0 TIBETAN MARK BKA- SHOG GI MGO RGYAN


> We don't document what its shape is.

Yeah, this is a real problem. The way I am treating it in the draft UnicodeSet spec is as a multivalued property whose values are all the aliases. But that means that we are dropping the type information on the floor.

Which is what the tools do. But this is a problem, because people use the type information. For instance C++ discriminates on the type to determine whether an alias is valid in a `\N` escape: https://eel.is/c++draft/lex.universal.char#3. They do not allow figments nor abbreviations.

I think this ties into the broader questions of multivalued properties that are not just set-valued: we should have a way to talk about the first kPrimaryNumeric value, or about the aliases of type correction.

But having the default be to treat it as the set of values seems OK; we could then extend with things like `\p{kPrimaryNumeric[0]=1000000000000}` or `\p{Name_Alias[correction]=/-/}`.

unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java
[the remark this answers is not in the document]
markusicu: oh, we are in a nested class here... nvm

markusicu (approved):

macchiati:
nice!

PR #1063, opened 2025-03-14
Miscellaneous character.jsp cleanups

eggrobin:
* Put a `<wbr>` between values of multivalued properties so exemplar doesn’t make https://util.unicode.org/UnicodeJsps/character.jsp?a=a stupidly wide.
* Add back yellowing on multivalued properties (which was accidentally removed by #1018, see, _e.g._,  kPrimaryNumeric in https://util.unicode.org/UnicodeJsps/character.jsp?a=5146&showDevProperties=1)

### DIFF
--- a/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeUtilities.java
+++ b/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeUtilities.java
@@ -1654,7 +1654,7 @@ public class UnicodeUtilities {
         class PropertyAssignment {
             VersionInfo first;
             VersionInfo last;
-            String value;
+            ArrayList<String> values;
         }
         final boolean isMultivalued = getFactory().getProperty(propName).isMultivalued();
         List<PropertyAssignment> history = new ArrayList<>();
@@ -1684,16 +1684,17 @@ public class UnicodeUtilities {
                 if (property == null) {
                     continue;
                 }
-                String value = property.getValue(codePoint);
+                ArrayList<String> values = new ArrayList<>();
+                for (String value : property.getValues(codePoint)) {
+                    values.add(value);
+                }
                 PropertyAssignment lastAssignment =
                         history.isEmpty() ? null : history.get(history.size() - 1);
-                if (lastAssignment == null
-                        || (value != null && !value.equals(lastAssignment.value))
-                        || (value == null && lastAssignment.value != null)) {
+                if (lastAssignment == null || (!values.equals(lastAssignment.values))) {
                     PropertyAssignment assignment = new PropertyAssignment();
                     assignment.first = version;
                     assignment.last = version;
-                    assignment.value = value;
+                    assignment.values = values;
                     history.add(assignment);
                 } else {
                     lastAssignment.last = version;
@@ -1704,7 +1705,10 @@ public class UnicodeUtilities {
             var current = new PropertyAssignment();
             current.first = Settings.LAST_VERSION_INFO;
             current.last = Settings.LAST_VERSION_INFO;
-            current.value = getFactory().getProperty(propName).getValue(codePoint);
+            current.values = new ArrayList<>();
+            for (String value : getFactory().getProperty(propName).getValues(codePoint)) {
+                current.values.add(value);
+            }
             history.add(current);
         }
         out.append(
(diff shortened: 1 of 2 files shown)


macchiati (dismissed):
looks good, thanks!

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeUtilities.java
```diff
@@ -1684,16 +1684,17 @@ class PropertyAssignment {
                 if (property == null) {
                     continue;
                 }
-                String value = property.getValue(codePoint);
+                ArrayList<String> values = new ArrayList<>();
+                for (String value : property.getValues(codePoint)) {
+                    values.add(value);
+                }
```
markusicu, line 1690: optional: looks like it could use a Stream
- if getValues() is only Iterable: `property.getValues(codePoint).forEach(values::add);`
- if getValues() is a Collection: `List<String> values = property.getValues(codePoint).stream().collect(Collectors.toList());`

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeUtilities.java
```diff
@@ -1704,7 +1705,10 @@ class PropertyAssignment {
             var current = new PropertyAssignment();
             current.first = Settings.LAST_VERSION_INFO;
             current.last = Settings.LAST_VERSION_INFO;
-            current.value = getFactory().getProperty(propName).getValue(codePoint);
+            current.values = new ArrayList<>();
+            for (String value : getFactory().getProperty(propName).getValues(codePoint)) {
```
markusicu, line 1709: optional / ditto

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeUtilities.java
```diff
@@ -1734,29 +1738,36 @@ class PropertyAssignment {
...
-                                                + propName
-                                                + "="
-                                                + hValue
-                                                + ":]'>"))
-                                + versionRange
-                                + hValue
-                                + (isMultivalued ? "" : "</a>")
-                                + "</td>");
+            var htmlValue = new StringBuilder();
+            for (int i = 0; i < assignment.values.size(); ++i) {
```
markusicu, line 1742: very optional: with some fiddling, this should be able to become a stream where we map null vs. other, then use a string joiner...

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeUtilities.java
```diff
@@ -1734,29 +1738,36 @@ class PropertyAssignment {
...
+                            + ">"
+                            + (isMultivalued
+                                    ? "<span" + (isNew ? " class='changed'" : "") + ">"
+                                    : ("<a target='u' "
+                                            + (isNew ? "class='changed' " : "")
+                                            + "href='list-unicodeset.jsp?a=[:"
+                                            + (isCurrent ? "" : "U" + last + ":")
+                                            + propName
+                                            + "="
+                                            + htmlValue.toString()
```
markusicu, line 1765: optional: can you drop this .toString()?

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeUtilities.java
```diff
@@ -1734,29 +1738,36 @@ class PropertyAssignment {
...
+                                            + ":]'>"))
+                            + versionRange
+                            + htmlValue.toString()
```
markusicu, line 1768: ditto

unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
```diff
@@ -265,9 +265,10 @@ public String getVersion() {
     }
 
     public Iterable<String> getValues(int codepoint) {
-        return isMultivalued
-                ? delimiterSplitter.split(getValue(codepoint))
-                : Collections.singleton(getValue(codepoint));
+        String value = getValue(codepoint);
+        return isMultivalued && value != null
+                ? delimiterSplitter.split(value)
```
markusicu, line 270: optional: does this return a Collection? if so, upgrade the return type?

markusicu (approved):

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeUtilities.java
eggrobin: I briefly thought of that, but I am not sure it would be more readable. I have a tendency to write Rust in that kind of extremely functional style, maybe I should try it in Java too…

Will toy with it and report back.

eggrobin: The result seems surprisingly palatable.

unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
eggrobin: It returns an Iterable. This is a Google thing. Its documentation says `Java 8 users may prefer {@link #splitToStream} instead.`. Do we prefer that?

eggrobin: I think I am at least as confused by the Iterable/Collection/Stream situation as you are by C++ iterators…

markusicu: > I think I am at least as confused by the Iterable/Collection/Stream situation as you are by C++ iterators…

I bet I am still more confused ;-)

markusicu: > It returns an Iterable. This is a Google thing. Its documentation says `Java 8 users may prefer {@link #splitToStream} instead.`. Do we prefer that?

Maybe. For the singleton case, we could then presumably use `Stream.of(value)`

markusicu (approved):
teaching each other programming ;-)

unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
markusicu: we could then call this valueStream()

macchiati: Streams are the preferred form where you normally just want to deal with the contents one at a time. You can always create a iterator from a stream.

eggrobin: I’ll do that in a later PR, since it is going to affect other callers.

Issue #1049, opened 2025-02-19
Update UAX42 to document which UCDXML fields correspond to UCD properties (UAX44) vs. which are “just data” corresponding to various UCD files


jowilco:
The 16.0 implementation and previous doesn't differentiate between UCD properties and additional data. One example of this is the <normalization-corrections> section. Should we differentiate between properties and data, and does it make sense to maintain sections like <normalization-corrections>?

markusicu:
FYI @eggrobin re types of Unicode data

eggrobin:
> Should we differentiate between properties and data

Extremely yes.

I am trying to sort out this mess in https://github.com/unicode-org/unicodetools/pull/1064, and in the process I found https://github.com/unicode-org/properties/issues/390.


> does it make sense to maintain sections like `<normalization-corrections>`?

No opinion, you are the UCDXML user here; do what you think is useful?

jowilco:
I just wanted to write down some thoughts about UCDXML and the property types.

[TR44](https://www.unicode.org/reports/tr44/#About_Property_Table) talks about 4 values for status for properties: Normative, Informative, Contributory, and Provisional. However, I believe that there are technically 10 PropertyStatus values as defined in PropertyStatus.java. PropertyStatus.java does align UcdProperties to sets, but see comment on [line 134](https://github.com/unicode-org/unicodetools/blob/c7f0fdb9265c33390ed203a246b9f625cae5d3cf/unicodetools/src/main/java/org/unicode/props/PropertyStatus.java#L134). UcdProperties doesn't track the PropertyStatus value, and I don't believe that data is in the UCD files either. Therefore, I don't believe that there is a single source for determining the PropertyStatus values for all UcdProperties.

In theory, I could work around this because I do have a class that stores additional metadata that I need for generating UCDXML: java/org/unicode/xml/UCDPropertyDetail.java. I could easily add PropertyStatus values, with the obvious caveat that I would then have to maintain these values. #1064 might be the solution.

The next question is how we would deliver this as part of UCDXML. I would assume that the goal is to reduce the file size for the minimum set of attributes that define a code point. There are essentially two sets of deliverables for UCDXML:

1. The UCDXML files themselves. There are 6 files (GROUPED|FLAT * ALL|NOUNIHAN|UNIHAN).
2. UAX42 and the corresponding RNC (RelaxNG) schema.

I don't think that the schema would need to change - it should validate all possible properties, and therefore all possible properties according to PropertyStatus values.
[...]

PR #1064, opened 2025-03-18
Propertiness

eggrobin:
A classification of properties derived from presence in PropertyAliases, or derived from a field that we are forced to fill in in ExtraPopertyAliases (contrast PropertyStatus.java which is out of date).

In character.jsp, split the information into (UCD properties, non-UCD properties, UCD non-properties, non-UCD non-properties), with a further split for Unihan (out of UCD properties and after UCD non-properties). See it in staging:
* https://unicode-jsps-staging-o2ookmn2oq-uc.a.run.app/UnicodeJsps/character.jsp?a=A7FE
* https://unicode-jsps-staging-o2ookmn2oq-uc.a.run.app/UnicodeJsps/character.jsp?a=3400


macchiati:
I'd suggest that the top be the properties on https://www.unicode.org/reports/tr18/#RL2.7, perhaps with those groupings.

Put all Contributory and Provisional into a separate bucket.

Not sure what the parens are for, as in [(kEH_Core)](https://unicode-jsps-staging-o2ookmn2oq-uc.a.run.app/UnicodeJsps/properties.jsp?a=kEH_Core#kEH_Core)

Some values don't have links, eg "Obsolete"

[Identifier_Status](https://unicode-jsps-staging-o2ookmn2oq-uc.a.run.app/UnicodeJsps/properties.jsp?a=Identifier_Status#Identifier_Status)	[Restricted](https://unicode-jsps-staging-o2ookmn2oq-uc.a.run.app/UnicodeJsps/list-unicodeset.jsp?a=[:Identifier_Status=Restricted:])
[Identifier_Type](https://unicode-jsps-staging-o2ookmn2oq-uc.a.run.app/UnicodeJsps/properties.jsp?a=Identifier_Type#Identifier_Type)	Obsolete

The 
If you are going to have a bucket _Non-UCD properties for U+A7FE_, then add confusable, emoji, ...

Will look it over more tomorrow.

eggrobin:
> Here I am instead doing what I can based on what we are forced to maintain, namely *PropertyAliases.txt.

Note that beyond the cosmetics of grouping character.jsp, we actually want to keep track of the « is this a UCD property » information, see #1049.

eggrobin:
Note: I tried splitting out Provisional from Normative+Informative, and that seemed counterproductive for Unihan and Unikemet (which are the only places where we have Informative properties) to have them in two blocks; hence the parentheses approach.

eggrobin:
> RGI_Emoji (but not RGI_Emoji_*_Sequence) should be there because it is described as a property in UTS51

Ah nevermind, I see UTS51 also describes the RGI_Emoji_*_Sequence zoo as properties. I’ll fix that.

eggrobin:
As noted in the TODOs, I’d like to move RGI_Emoji and IDNA2008_Category into IndexUnicodeProperties (rather than being patched into the JSPs), and to add RGI_Emoji_Qualification, all of these being NonUcdProperty.

But I will do that in a subsequent PR.

eggrobin:
@markusicu Friendly ping, since I think some of @jowilco’s work is blocked on this.

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeUtilities.java
```diff
@@ -1440,6 +1442,42 @@ public static void showProperties(
 
         String kRSUnicode = getFactory().getProperty("kRSUnicode").getValue(cp);
         boolean isUnihan = kRSUnicode != null;
+        List<UcdProperty> indexedProperties =
```
markusicu, line 1445: optional -- might simplify something:

How about, rather than just building separate lists of properties, you add an `enum PropCategory { UCD, NON_UCD, ... CJK, ...}`, and create a `Map<PropCategory, List<UcdProperty>>`?

You could then also use maps from PropCategory to table headings and such.

----
Does it matter if these lists are List's? Or do you just need Collection's?

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeUtilities.java
```diff
@@ -1709,12 +1766,14 @@ class PropertyAssignment {
...
         out.append(
-                "<tr><th><a target='c' href='properties.jsp?a="
+                "<tr><th width='50%'><a target='c' href='properties.jsp?a="
                         + propName
                         + "#"
                         + propName
                         + "'>"
+                        + (provisional ? "(" : "")
                         + propName
+                        + (provisional ? ")" : "")
```
markusicu, line 1776: optional:
```suggestion
                        + (provisional ? "(" + propName + ")" : propName)
```

markusicu (dismissed):

markusicu (approved):

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeUtilities.java
eggrobin: > Does it matter if these lists are List's?

Not really, I convert them to lists of String below anyway (because one of them is not a list of UcdProperty, namely the list of stuff that gets added in the tools).

eggrobin: I will merge this now to unblock John and see if I can come up with something cleaner in a subsequent PR.

PR #1069, opened 2025-03-20
Saudi ريال

eggrobin:
[[183-C4](https://www.unicode.org/cgi-bin/GetL2Ref.pl?183-C4)] Consensus: UTC accepts U+20C1 SAUDI RIYAL SIGN for encoding in the Currency Symbols block in Unicode Version 17.0 based on [L2/25-075](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/25-075). [Ref. 1.1 in [L2/25-091R](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/25-091R)]

* L2/25-075
* https://github.com/unicode-org/sah/issues/619
* https://github.com/unicode-org/utc-release-management/issues/198


unicodetools/src/main/resources/org/unicode/text/UCD/AdditionComparisons/198.txt
```diff
@@ -0,0 +1,18 @@
+# Symbol: Saudi riyal sign
+# https://github.com/unicode-org/utc-release-management/issues/198
+
+# Names always differ.
+# Age always differs since these tests are comparing additions to pre-existing characters.
+Ignoring Name Age:
+
+# Ignore the security and IDNA properties, as these are not yet included for provisionally assigned characters.
+Ignoring Confusable_MA Identifier_Status Identifier_Type Idn_Status Idn_Mapping Idn_2008:
```
markusicu, line 9: It's accepted, not provisionally assigned.
Should we just set the ID_Type etc. now?

markusicu (approved):

unicodetools/src/main/resources/org/unicode/text/UCD/AdditionComparisons/198.txt
eggrobin: I don’t want to go there for the addition comparison tests for now.

Also, I do not know how to turn the security crank.

Issue #1074, opened 2025-03-24
Never use ICU properties


eggrobin:
See #1072.

CollatorEquivalencesNew and GenerateNormalizeForMatch both have this:
```java
    static UnicodeSet.XSymbolTable NO_PROPS =
            new UnicodeSet.XSymbolTable() {
                @Override
                public boolean applyPropertyAlias(
                        String propertyName, String propertyValue, UnicodeSet result) {
                    throw new IllegalArgumentException(
                            "Don't use any ICU Unicode Properties! "
                                    + propertyName
                                    + "="
                                    + propertyValue);
                }
                ;
            };

    static {
        UnicodeSet.setDefaultXSymbolTable(NO_PROPS);
    }
```

We should have something equivalent to that everywhere.

macchiati:
+1

Issue #1073, opened 2025-03-24
There can only be one XSymbolTable


eggrobin:
See https://github.com/unicode-org/unicodetools/issues/1072.

In particular, UnicodePropertySymbolTable is eerily similar to UnicodeSetUtilities.MySymbolTable (used in the JSPs). TestUnicodeInvariants.ChainedSymbolTable (used in the invariants) has a lot of the same features (`=@none@`, unary-query for binary properties, etc.), but these are implemented in VersionedProperty rather than directly in `applyPropertyAlias`.

Previously:
* https://github.com/unicode-org/unicodetools/issues/66
* https://github.com/unicode-org/unicodetools/pull/667

markusicu:
Please remember to put issues onto the dashboard. At some point we should look at that again...

PR #1075, opened 2025-03-24
Fix segmentation tests

### LINKED ISSUE #1072: Segmentation test cases are being generated based on data for the previous version of Unicode
See the Unicode mailing list email titled _Question Regarding UCD Draft Files and GraphemeBreakTest Discrepancy_.

eggrobin:
* Fix #1072.
* Never return `false` from an override of applyPropertyAlias; this is a band-aid for #1074 (and caused segmentation test generation to fail with empty sets), but does not fix the issue systematically.
* No effort made to address #1073 in this PR.

### DIFF
--- a/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
+++ b/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java
@@ -195,34 +195,41 @@ public class UnicodeSetUtilities {
                 } else {
                     int debug = 0;
                 }
-                status = applyPropertyAlias0(prop, propertyValue, result, invert);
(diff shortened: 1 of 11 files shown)


markusicu (dismissed):

markusicu (approved):

eggrobin:
> * Never return false from an override of applyPropertyAlias; this is a band-aid for https://github.com/unicode-org/unicodetools/issues/1074 (and caused segmentation test generation to fail with empty sets), but does not fix the issue systematically.

Well, this band aid is too strong already, I am getting failures left and right from things that use unary-query-expression (which is not supported by some of the symbol tables).

So I will revert that safety, keep the redundant =true where we know its absence causes a bug in test data, and deal with #1073 first on my way to #1074.

markusicu (approved):

PR #1077, opened 2025-03-27
Unicode 17.0 Confusables

josh-hadley:


roozbehp (approved):

roozbehp (approved):

unicodetools/data/security/dev/data/confusablesSummaryIdentifier.txt
```diff
@@ -324,7 +324,7 @@
 
 #	bl	Ьl	Ь1	ЬІ	Ы
 	(‎ bl ‎)	0062 006C	 LATIN SMALL LETTER B, LATIN SMALL LETTER L
-←	(‎ Ьl ‎)	042C 006C	 CYRILLIC CAPITAL LETTER SOFT SIGN, LATIN SMALL LETTER L	# →ЬІ→→Ь1→
+←	(‎ Ьl ‎)	042C 006C	 CYRILLIC CAPITAL LETTER SOFT SIGN, LATIN SMALL LETTER L	# →Ь1→
```
josh-hadley, line 327: I was puzzled about this change in the comment. Neither of these characters were involved in the current set of changes. Possibly another upstream code change? I can investigate if it seems important to identify the cause.
cc: @markusicu @eggrobin

eggrobin: I can’t say I understand these comments. You are adding things into the equivalence classes of both `b` and `l`, so maybe this has some effect on how this gets turned into `bl`?

I hope it is not the result of a code change, I haven’t done anything recently that should have an effect.

markusicu:
I glanced at it but didn't look thoroughly. I am happy if y'all are happy. Thanks for getting this done!

josh-hadley:
> I glanced at it but didn't look thoroughly. I am happy if y'all are happy. Thanks for getting this done!

I'm happy with it, and @roozbehp has already given it a 👍 so I'm content to wrap it up

markusicu:
yes, go for it

macchiati:
There is some indeterminacy in the comments on the path between characters.
I've noticed it before but fixing it was always swamped by other things. It
might be due to using hash table internally. (I think internally the hash
tables in Java now change the iteration order, for security.)

roozbehp:
BTW, it looks like there are more Unicode 17.0 characters coming (at least one currency sign, and it looks like some disunified CJK characters). Just a reminder to look at confusables again when UTC approves them.

Issue #639, opened 2024-01-08
Retire ToolUnicodeProperties & UCD.java


macchiati:
There are currently two ways of parsing the property files, ToolUnicodeProperties.java/UCD.java (very old), and the more modern IndexUnicodeProperties.java. I've long had on my task list replacing the old with new; the new is more extensible, and much easier to update for new properties and property values.

Here is a draft of the steps to do that:

1. Analyze the differences between ToolUnicodeProperties.java and IndexUnicodeProperties.java to see if the latter can simply replace the former.
2. If the differences would affect the results (tests, file generation), then write a shim for IndexUnicodeProperties.java where necessary.
3. Find all the cases where internal code is calling UCD.java directly instead of using ToolUnicodeProperties.java.
4. Replace those by calls to ShimIndexUnicodeProperties.java, and make sure everything still works.
5. Replace calls to ToolUnicodeProperties by ShimIndexUnicodeProperties.java
6. Remove UCD.java, UTC_Types.java, UCD_Names.java, ToolUnicodeProperties.java, and code only used by them.

eggrobin:
See also https://github.com/unicode-org/unicodetools/issues/484.

macchiati:
thanks for the cross link

PR #1080, opened 2025-03-31
One symbol table to rule them all

eggrobin:
Clearly our problem of proliferation of symbol tables will be solved by adding yet another one; so far not used, and tested against examples from the working draft UnicodeSet spec.

This should be able to replace most of the existing symbol tables as-is. It needs some tweaking before I can use it in the invariant tests, so that it can use the versioned ToolUnicodeProperties[^retire] in some circumstances.

Towards #1073 and #1074.

[^retire]: Yes, those ones: #639. We’re not retiring them just yet…

### DIFF
--- a/unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
+++ b/unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
@@ -198,6 +198,7 @@ public abstract class UnicodeProperty extends UnicodeLabel {
             EXTENDED_MASK = 1,
             CORE_MASK = ~EXTENDED_MASK,
             BINARY_MASK = (1 << BINARY) | (1 << EXTENDED_BINARY),
+            NUMERIC_MASK = (1 << NUMERIC) | (1 << EXTENDED_NUMERIC),
             STRING_MASK = (1 << STRING) | (1 << EXTENDED_STRING),
             STRING_OR_MISC_MASK =
                     (1 << STRING) | (1 << EXTENDED_STRING) | (1 << MISC) | (1 << EXTENDED_MISC),
--- a/unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
+++ b/unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
@@ -445,27 +445,24 @@ public abstract class UnicodeProperty extends UnicodeLabel {
         if (isMultivalued && propertyValue != null && propertyValue.contains(delimiter)) {
             throw new IllegalArgumentException(
                     "Multivalued property values can't contain the delimiter.");
-        } else {
-            Comparator<String> comparator;
-            if (isType(NUMERIC_MASK)) {
-                // UAX44-LM1.
-                comparator = RATIONAL_COMPARATOR;
-            } else if (getName().equals("Name") || getName().equals("Name_Alias")) {
-                // UAX44-LM2.
-                comparator = CHARACTER_NAME_COMPARATOR;
-            } else if (isType(BINARY_OR_ENUMERATED_OR_CATALOG_MASK)) {
-                // UAX44-LM3
-                comparator = PROPERTY_COMPARATOR;
-            } else {
-                // String-valued or Miscellaneous property.
-                comparator = null;
-            }
-            return getSet(
-                    propertyValue == null
-                            ? NULL_MATCHER
-                            : new SimpleMatcher(propertyValue, comparator),
-                    result);
         }
+        Comparator<String> comparator;
+        if (isType(NUMERIC_MASK)) {
+            // UAX44-LM1.
+            comparator = RATIONAL_COMPARATOR;
+        } else if (getName().equals("Name") || getName().equals("Name_Alias")) {
+            // UAX44-LM2.
+            comparator = CHARACTER_NAME_COMPARATOR;
+        } else if (isType(BINARY_OR_ENUMERATED_OR_CATALOG_MASK)) {
+            // UAX44-LM3
+            comparator = PROPERTY_COMPARATOR;
+        } else {
+            // String-valued or Miscellaneous property.
+            comparator = null;
+        }
+        return getSet(
+                propertyValue == null ? NULL_MATCHER : new SimpleMatcher(propertyValue, comparator),
+                result);
     }
 
     private UnicodeMap<String> unicodeMap = null;
--- a/unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
+++ b/unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
@@ -446,19 +446,24 @@ public abstract class UnicodeProperty extends UnicodeLabel {
             throw new IllegalArgumentException(
                     "Multivalued property values can't contain the delimiter.");
         } else {
+            Comparator<String> comparator;
+            if (isType(NUMERIC_MASK)) {
+                // UAX44-LM1.
+                comparator = RATIONAL_COMPARATOR;
+            } else if (getName().equals("Name") || getName().equals("Name_Alias")) {
+                // UAX44-LM2.
+                comparator = CHARACTER_NAME_COMPARATOR;
+            } else if (isType(BINARY_OR_ENUMERATED_OR_CATALOG_MASK)) {
+                // UAX44-LM3
+                comparator = PROPERTY_COMPARATOR;
+            } else {
+                // String-valued or Miscellaneous property.
+                comparator = null;
+            }
             return getSet(
                     propertyValue == null
                             ? NULL_MATCHER
-                            : new SimpleMatcher(
-                                    propertyValue,
-                                    isType(NUMERIC_MASK)
-                                            ? RATIONAL_COMPARATOR
-                                            : getName().equals("Name")
-                                                            || getName().equals("Name_Alias")
-                                                    ? CHARACTER_NAME_COMPARATOR
-                                                    : isType(STRING_OR_MISC_MASK)
-                                                            ? null
-                                                            : PROPERTY_COMPARATOR),
+                            : new SimpleMatcher(propertyValue, comparator),
                     result);
         }
     }
--- a/unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
+++ b/unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
@@ -740,8 +740,8 @@ public abstract class UnicodeProperty extends UnicodeLabel {
 
     public static int compareRationals(String a, String b) {
         if (a == b) return 0;
-        if (a == null) return -1;
-        if (b == null) return 1;
+        if (a == null || a.equalsIgnoreCase("NaN")) return -1;
+        if (b == null || b.equalsIgnoreCase("NaN")) return 1;
         return RationalParser.BASIC.parse(a).compareTo(RationalParser.BASIC.parse(b));
     }
 
--- a/unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
+++ b/unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
@@ -740,8 +740,13 @@ public abstract class UnicodeProperty extends UnicodeLabel {
 
     public static int compareRationals(String a, String b) {
         if (a == b) return 0;
-        if (a == null || a.equalsIgnoreCase("NaN")) return -1;
-        if (b == null || b.equalsIgnoreCase("NaN")) return 1;
+        if (a == null) return -1;
+        if (b == null) return 1;
+        final boolean aIsNaN = equalNames(a, "NaN");
+        final boolean bIsNaN = equalNames(b, "NaN");
+        if (aIsNaN && bIsNaN) return 0;
+        if (aIsNaN) return -1;
+        if (bIsNaN) return 1;
         return RationalParser.BASIC.parse(a).compareTo(RationalParser.BASIC.parse(b));
     }
 
--- a/unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
+++ b/unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
@@ -446,6 +446,9 @@ public abstract class UnicodeProperty extends UnicodeLabel {
             throw new IllegalArgumentException(
                     "Multivalued property values can't contain the delimiter.");
         }
+        if (propertyValue == null) {
+            return getSet(NULL_MATCHER, result);
+        }
         Comparator<String> comparator;
         if (isType(NUMERIC_MASK)) {
             // UAX44-LM1.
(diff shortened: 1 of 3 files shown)


markusicu:
4 failing CI checks

unicodetools/src/main/java/org/unicode/text/UCD/VersionedSymbolTable.java
```diff
@@ -0,0 +1,426 @@
...
+
+    public VersionedSymbolTable setUnversionedExtensions(UnicodeProperty.Factory factory) {
+        unversionedExtensions = factory;
+        return this;
+    }
+
+    /**
+     * Parses a string prefixed with an optional-version-qualifier. If there is a version-qualifier,
+     * returns the corresponding VersionInfo and removes the prefix from the given StringBuilder.
+     */
```
markusicu, line 45:
```suggestion
    /**
     * Parses a string prefixed with an optional-version-qualifier. If there is a version-qualifier,
     * returns the corresponding VersionInfo and removes the prefix from the given StringBuilder.
     * Otherwise returns null.
     */
```

unicodetools/src/main/java/org/unicode/text/UCD/VersionedSymbolTable.java
```diff
@@ -0,0 +1,426 @@
...
+    private VersionInfo parseVersionQualifier(StringBuilder qualified) {
+        int posColon = qualified.indexOf(":", 0);
+        if (posColon < 0) {
+            return null;
+        } else {
```
markusicu, line 50: optional: remove the "else". just continue without indentation after the return-null block.

unicodetools/src/main/java/org/unicode/text/UCD/VersionedSymbolTable.java
```diff
@@ -0,0 +1,426 @@
...
+                }
+            }
+        }
+    }
+
+    private static Map<UcdPropertyValues.General_Category_Values, String[]>
+            COARSE_GENERAL_CATEGORIES =
+                    Map.of(
+                            UcdPropertyValues.General_Category_Values.Other,
+                            new String[] {"Cc", "Cf", "Cn", "Co", "Cs"},
```
markusicu, line 128: optional / cleaner: how about changing String[] to a List or a Set, and using `List.of("Cc", "Cf", "Cn", "Co", "Cs")` or Set.of()?

unicodetools/src/main/java/org/unicode/text/UCD/VersionedSymbolTable.java
```diff
@@ -0,0 +1,426 @@
...
+        String leftHandSide = beforeEquals;
+        String propertyPredicate = afterEquals;
+        boolean interiorlyNegated = false;
+        int posNotEqual = beforeEquals.indexOf('≠');
+        // TODO(egg): We cannot distinguish \p{X=} from \p{X} in this API, both give us an empty
+        // string as afterEquals.  This is an @internal API, so we could change it to pass null in
+        // the unary case.
+        if (posNotEqual >= 0) {
+            propertyPredicate =
+                    afterEquals.length() == 0
```
markusicu, line 176:
```suggestion
                    afterEquals.isEmpty()
```

unicodetools/src/main/java/org/unicode/text/UCD/VersionedSymbolTable.java
```diff
@@ -0,0 +1,426 @@
...
+                            ? beforeEquals.substring(posNotEqual + 1)
+                            : beforeEquals.substring(posNotEqual + 1) + "=" + afterEquals;
+            leftHandSide = beforeEquals.substring(0, posNotEqual);
+            interiorlyNegated = true;
+        }
+        if (interiorlyNegated) {
```
markusicu, line 182: nit: you don't need this variable if you do this processing inside the `if (posNotEqual >= 0)` block

unicodetools/src/main/java/org/unicode/text/UCD/VersionedSymbolTable.java
```diff
@@ -0,0 +1,426 @@
...
+            result.addAll(getNonNegatedPropertyQuerySet(leftHandSide, propertyPredicate));
+        }
+        return true;
+    }
+
+    private UnicodeSet getNonNegatedPropertyQuerySet(
+            String leftHandSide, String propertyPredicate) {
+        final var unqualifiedLeftHandSide = new StringBuilder(leftHandSide);
+        final var queriedVersion = parseVersionQualifier(unqualifiedLeftHandSide);
+        final var deducedQueriedVersion = queriedVersion == null ? implicitVersion : queriedVersion;
```
markusicu, line 195: totally optional: some of the variable names seem longer than necessary. for example, this one i would just call "version". for "queriedProperties", maybe just "properties". etc.

shorter names make for easier reading, shorter lines, and less wrapping. just not so short that they get confusing :-)

unicodetools/src/main/java/org/unicode/text/UCD/VersionedSymbolTable.java
```diff
@@ -0,0 +1,426 @@
...
+
+        final var queriedProperties = IndexUnicodeProperties.make(deducedQueriedVersion);
+
+        if (propertyPredicate.length() == 0) {
```
markusicu, line 199: .isEmpty()

unicodetools/src/main/java/org/unicode/text/UCD/VersionedSymbolTable.java
```diff
@@ -0,0 +1,426 @@
...
+            // Either unary-property-query, or binary-property-query with an empty property-value.
+            final var script = queriedProperties.getProperty(UcdProperty.Script);
+            final var generalCategory = queriedProperties.getProperty(UcdProperty.General_Category);
+            if (script.isValidValue(unqualifiedLeftHandSide.toString())) {
```
markusicu, line 203: lots of .toString() calls here. please make a String after parsing the version out of leftHandSide. if there was no version prefix, then the leftHandSide itself is already the String you want.

probably easiest to do something like
```
if (there was a version prefix) {
    leftHandSide = unqualifiedLeftHandSide.toString();
}
... use leftHandSide ...
```

unicodetools/src/main/java/org/unicode/text/UCD/VersionedSymbolTable.java
```diff
@@ -0,0 +1,426 @@
...
+    private UnicodeSet getNonNegatedPropertyQuerySet(
+            String leftHandSide, String propertyPredicate) {
+        final var unqualifiedLeftHandSide = new StringBuilder(leftHandSide);
+        final var queriedVersion = parseVersionQualifier(unqualifiedLeftHandSide);
+        final var deducedQueriedVersion = queriedVersion == null ? implicitVersion : queriedVersion;
+
+        final var queriedProperties = IndexUnicodeProperties.make(deducedQueriedVersion);
+
+        if (propertyPredicate.length() == 0) {
+            // Either unary-property-query, or binary-property-query with an empty property-value.
```
markusicu, line 200: It seems like we should first check if leftHandSide is a valid property name, and do a normal prop=value lookup, before trying the script and gc shortcuts. Except for the isc=C issue.

unicodetools/src/main/java/org/unicode/text/UCD/VersionedSymbolTable.java
```diff
@@ -0,0 +1,426 @@
...
+                // TODO(egg): Remove when we can tell this is a unary query.
+                if (queriedProperty.isType(UnicodeProperty.STRING_OR_MISC_MASK)) {
+                    return queriedProperty.getSet("");
+                }
+                throw new IllegalArgumentException(
+                        "Invalid unary-query-expression for non-binary property "
+                                + queriedProperty.getName());
+            }
+            return queriedProperty.getSet(UcdPropertyValues.Binary.Yes);
+        } else {
```
markusicu, line 230: optional: long function; maybe break it into multiple smaller ones? e.g., for most of the above call a getSomethingSetForEmptyValue(queriedProperties, leftHandSide)?

unicodetools/src/main/java/org/unicode/text/UCD/VersionedSymbolTable.java
```diff
@@ -0,0 +1,426 @@
...
+                            + queriedProperty.getTypeName()
+                            + " property "
+                            + queriedProperty.getName()
+                            + " and "
+                            + comparisonProperty.getTypeName()
+                            + " property "
+                            + comparisonProperty.getName());
+        }
+        final var result = new UnicodeSet();
+        for (int cp = 0; cp <= 0x10FFFF; ++cp) {
```
markusicu, line 412: Should we note that this excludes empty/multi-character strings of properties-of-strings?

unicodetools/src/test/java/org/unicode/text/UCD/TestVersionedSymbolTable.java
```diff
@@ -0,0 +1,337 @@
...
+
+        public <T extends CharSequence> UnicodeSetTestFluent isEqualToUnicodeSet(
+                String expectedExpression) {
+            final var expected = new UnicodeSet(expectedExpression);
+            assertTrue(
+                    set.containsAll(expected),
+                    expected + " ⊆ " + expression + " = " + set.toPattern(true));
+            assertTrue(
+                    expected.containsAll(set),
+                    expected + " ⊇ " + expression + " = " + set.toPattern(true));
```
markusicu, line 281: optional: when these fail, it can help to show the differences, for example the non-empty set.removeAll(expected) for the first assertion

unicodetools/src/main/java/org/unicode/text/UCD/VersionedSymbolTable.java
eggrobin: Good point.

In my draft spec I express things solely in terms of character properties, and leave properties of strings to extensions.
Including strings seems potentially useful at first glance, but then it rapidly gets messy:
* What is the value of a character property on a string of length other than 1? (null?)
* The set of strings where one property equals another need not be finite (in particular, if two properties are null for all but 1-character strings, that set is the set of all strings of length other than 1).

I’ll add a comment here for now, and think about a note in the draft spec.

unicodetools/src/main/java/org/unicode/text/UCD/VersionedSymbolTable.java
eggrobin: I get your point, and I think I would normally have called this `version` rather than `queriedVersion`, but since this is written against the UnicodeSet spec, I am trying to match the terminology there. And since there I need to talk about two versions, they both have qualified names.

unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
```diff
@@ -449,11 +451,14 @@ public final UnicodeSet getSet(String propertyValue, UnicodeSet result) {
...
-                                                    ? null
-                                                    : PROPERTY_COMPARATOR),
+                                    isType(NUMERIC_MASK)
+                                            ? RATIONAL_COMPARATOR
+                                            : getName().equals("Name")
+                                                            || getName().equals("Name_Alias")
+                                                    ? CHARACTER_NAME_COMPARATOR
+                                                    : isType(STRING_OR_MISC_MASK)
+                                                            ? null
+                                                            : PROPERTY_COMPARATOR),
```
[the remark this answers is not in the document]
eggrobin: Yes, this had gotten out of hand…

Also now I can add comments linking that to UAX44-LM rules.

unicodetools/src/main/java/org/unicode/text/UCD/VersionedSymbolTable.java
eggrobin: Yes. I probably want to do that once I can deal with the isc=C issue by looking at whether there is an =.

This should be possible by bumping the ICU dependency to a snapshot past https://github.com/unicode-org/icu/pull/3456, but I will look at that in a subsequent PR.

unicodetools/src/main/java/org/unicode/text/UCD/VersionedSymbolTable.java
eggrobin: Done, but kept the long name because I think I will need to look at the original leftHandSide in a future revision.

unicodetools/src/main/java/org/unicode/text/UCD/VersionedSymbolTable.java
eggrobin: Yes, but the name is also acting as documentation (it matches a defined term in the spec).

unicodetools/src/main/java/org/unicode/text/UCD/VersionedSymbolTable.java
markusicu: > * What is the value of a character property on a string of length other than 1? (null?)

We mostly have binary properties of strings, so values are true/yes / false/no.
The value for a string that's not listed is false.

> * The set of strings where one property equals another need not be finite

We discussed this at length. Unicode properties of strings are only defined for finite sets of strings. We define them by writing all strings into a data file. We definitely do not define them via a matching algorithm.

eggrobin: > We definitely do not define [properties of strings] via a matching algorithm.

Yes. But in contrast, the value of a property-comparison _is_ defined by matching values.

eggrobin:
Note: The failing Python thing is #1081.

unicodetools/src/test/java/org/unicode/text/UCD/TestVersionedSymbolTable.java
```diff
@@ -17,14 +19,17 @@
  * reasonable. If they are broken by changes to property assignments, feel free to update them.
  */
 public class TestVersionedSymbolTable {
+   XSymbolTable old_default;
```
markusicu: optional but cleaner: I would prefer writing UnicodeSet.XSymbolTable here and not importing this nested type. Just like I prefer using `Map.Entry` rather than the mysterious `Entry`.

unicodetools/src/main/java/org/unicode/text/UCD/VersionedSymbolTable.java
markusicu: > > We definitely do not define [properties of strings] via a matching algorithm.
> 
> Yes. But in contrast, the value of a property-comparison _is_ defined by matching values.

I think we can compare values on the union of the keySet()'s, plus all code points. That would be well-defined and finite. It's fine if that's a TODO comment or a repo issue for now.

unicodetools/src/main/java/org/unicode/props/UnicodeProperty.java
```diff
@@ -446,19 +446,24 @@ public final UnicodeSet getSet(String propertyValue, UnicodeSet result) {
...
+            } else if (isType(BINARY_OR_ENUMERATED_OR_CATALOG_MASK)) {
+                // UAX44-LM3
+                comparator = PROPERTY_COMPARATOR;
+            } else {
+                // String-valued or Miscellaneous property.
+                comparator = null;
+            }
             return getSet(
                     propertyValue == null
                             ? NULL_MATCHER
```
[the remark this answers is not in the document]
eggrobin: Done. This made me think that this does a tiny bit of value validation (checking that the delimiter is not in the query), and made me wonder whether it should do more (in particular, null is valid only for some property types).

Probably a question for another day and another PR.

markusicu (approved):

PR #1081, opened 2025-04-01
Python lint

eggrobin:
It looks like the linter got an update, and it is now annoyed at an unnecessary `global` (`_prop_data` is used but not assigned in `main()`).


josh-hadley (approved):
Looks good and appears to yield the same results when running on unzipped https://www.unicode.org/Public/UNIDATA/Unihan.zip

rodneymcgee:
thanks have a good day

PR #1082, opened 2025-04-01
Test property name uniqueness

### LINKED ISSUE #1079: Add tests for Property Alias Uniqueness and related invariants
See https://www.unicode.org/policies/stability_policy.html#Property_Alias_Uniqueness.

Also check that `code point` and `none` are unique in the property namespace, that the union of the binary property names, GC values, and script values also doesn’t have LM3 overlaps.

Flag new things like Other=ISO_Comment. They are OK, but they are very surprising.

eggrobin:
Fix #1079.

Also note redundant aliases (those may sometimes be needed because of the precise spelling stability or other consistency, but they are unusual).

### DIFF
--- a/unicodetools/src/test/java/org/unicode/text/UCD/TestCodeInvariants.java
+++ b/unicodetools/src/test/java/org/unicode/text/UCD/TestCodeInvariants.java
@@ -250,7 +250,6 @@ public class TestCodeInvariants {
             if (IndexUnicodeProperties.make()
                     .getProperty(property)
                     .isType(UnicodeProperty.BINARY_OR_ENUMERATED_OR_CATALOG_MASK)) {
-                noisy = property == UcdProperty.Decomposition_Type;
                 Set<String> expectedRedundant;
                 switch (property) {
                     case Block:
@@ -263,7 +262,9 @@ public class TestCodeInvariants {
                                         "med", "nar", "nb", "none", "sml", "sqr", "sub", "sup",
                                         "vert", "wide");
                         break;
-
+                    case Sentence_Break:
+                        expectedRedundant = Set.of("Sp");
+                        break;
                     default:
                         expectedRedundant = Set.of();
                         break;
@@ -284,12 +285,9 @@ public class TestCodeInvariants {
         final Map<String, String> aliasesByLM3Skeleton = new HashMap<>();
         for (T entity : namespace) {
             for (String alias : getNames.apply(entity)) {
-                if (noisy) System.err.println(alias);
                 final var matchingEntity = entitiesByAlias.get(alias);
                 final var lm3Skeleton = UnicodeProperty.toSkeleton(alias);
                 final var matchingAlias = aliasesByLM3Skeleton.get(lm3Skeleton);
-                if (noisy) System.err.println("->" + matchingAlias);
-                if (noisy) System.err.println("=>" + matchingEntity);
                 assertTrue(
                         matchingEntity == null || entity.equals(matchingEntity),
                         "!!Stability policy violation!! (Property Alias Uniqueness): alias "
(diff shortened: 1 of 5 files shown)


unicodetools/src/test/java/org/unicode/text/UCD/TestCodeInvariants.java
```diff
@@ -230,4 +242,132 @@ private static String showInfo(
...
+                                + entitiesByAlias.get(alias));
+                if (matchingEntity != null && !expectedRedundant.contains(alias)) {
+                    assertEquals(
+                            matchingAlias,
+                            alias,
+                            "Unusual (not a stability policy violation): distinct aliases for "
+                                    + entity
+                                    + " match each other");
+                }
+                entitiesByAlias.put(alias, entity);
```
markusicu, line 368: optional:

putIfAbsent()?
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Map.html#putIfAbsent(K,V)

or don't put() if  matchingEntity != null  ?

unicodetools/src/test/java/org/unicode/text/UCD/TestCodeInvariants.java
```diff
@@ -230,4 +242,132 @@ private static String showInfo(
...
+                        matchingEntity == null || entity.equals(matchingEntity),
+                        message
+                                + ": alias "
+                                + alias
+                                + " for "
+                                + entity
+                                + " matches alias "
+                                + matchingAlias
+                                + " for "
+                                + entitiesByAlias.get(alias));
```
markusicu, line 359:
```suggestion
                                + matchingEntity);
```

unicodetools/src/test/java/org/unicode/text/UCD/TestCodeInvariants.java
```diff
@@ -230,4 +242,132 @@ private static String showInfo(
...
+            Function<T, List<String>> getNames,
+            Set<String> expectedRedundant,
+            String message) {
+        final Map<String, T> entitiesByAlias = new TreeMap<>(UnicodeProperty.PROPERTY_COMPARATOR);
+        final Map<String, String> aliasesByLM3Skeleton = new HashMap<>();
+        for (T entity : namespace) {
+            for (String alias : getNames.apply(entity)) {
+                final var matchingEntity = entitiesByAlias.get(alias);
+                final var lm3Skeleton = UnicodeProperty.toSkeleton(alias);
+                final var matchingAlias = aliasesByLM3Skeleton.get(lm3Skeleton);
```
markusicu, line 348: random thought:
if the comparator was mutable and could remember the values when they compared equal, then you wouldn't need this extra Map because you could fetch the earlier alias from the comparator state...

markusicu (commented):
very nice!

unicodetools/src/test/java/org/unicode/text/UCD/TestCodeInvariants.java
eggrobin: Yes, but squirreling things away in mutable state feels weird.
If getEntry on the TreeMap had been public rather than an implementation detail of get, I could have use that, but alas…

unicodetools/src/test/java/org/unicode/text/UCD/TestCodeInvariants.java
eggrobin: For this one, if it is present, we have just checked that it is consistent with what is already there, so it doesn’t matter. So this comes down to expressing intent/readability.
On the one hand `put` feels like the default operation, so might as well use that if it does not matter.
But maybe `putIfAbsent` makes it clearer that we never mutate the map?

For the other one, `put` could be mutating (if we have an expectedRedundant alias), and we probably should not mutate (the earlier alias should be the canonical one).

So let’s say `putIfAbsent` for both.

markusicu (approved):

PR #1083, opened 2025-04-03
Actually check values of enumerated properties

eggrobin:
And fix some older silenced exceptions in historical UCD loading.


unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
```diff
@@ -1172,16 +1202,21 @@ private static void parseSimpleFieldFile(
                     IntRange range = new IntRange();
                     range.start = Utility.codePointFromHex(line.getParts()[0]);
                     range.end = Utility.codePointFromHex(line.getParts()[1]);
+                    // Unicode 2 puts FEFF both in Arabic Presentation Forms-B and in Specials.
+                    // We are not going to make Block multivalued for that, so we let the second
+                    // assignment win.
```
markusicu, line 1207: I was going to protest here... what a mess...

We pretty much have an invariant that block boundaries are xxx0..yyyF.
That makes a single-code point block just for FEFF basically ill-formed.

Plus, we can't have a discontiguous Block. This thing claims to make a Specials Block of FEFF + separate FFF0..FFFF.

And, it's probably the only thing listed out of code point order.

Actually, it gets worse: Unicode 2.1.4 changed Arabic Presentation Forms-B to end at FEFE (not some ...F) and moved FEFF to be listed in code point order. ... and removed the noncharacters FFFE/FFFF from the regular Specials, so that that block ended with FFFD, also not on a ...F!

Unicode 3/3.1 kept these. :-(

Only Unicode 3.2 regularized them.

I was going to claim that this was a mistake, but it looks like Unicode 2.1.4..3.1.1 confirm how you are resolving the collision for 2.0.

https://www.unicode.org/Public/2.0-Update/Blocks-1.txt
```
FE70; FEFF; Arabic Presentation Forms-B
FF00; FFEF; Halfwidth and Fullwidth Forms
FEFF; FEFF; Specials
FFF0; FFFF; Specials
```

https://www.unicode.org/Public/2.1-Update4/Blocks-2.txt
```
FE70; FEFE; Arabic Presentation Forms-B
FEFF; FEFF; Specials
FF00; FFEF; Halfwidth and Fullwidth Forms
FFF0; FFFD; Specials
```

https://www.unicode.org/Public/3.2-Update/Blocks-3.2.0.txt
```
FE70..FEFF; Arabic Presentation Forms-B
FF00..FFEF; Halfwidth and Fullwidth Forms
FFF0..FFFF; Specials
```

```suggestion
                    // Unicode 2 puts FEFF both in Arabic Presentation Forms-B and in Specials.
                    // We are not going to make Block multivalued for that, so we let the second
                    // assignment win.
                    // This fits with assignments in Unicode 2.1.4..3.1.1 where
                    // Arabic Presentation Forms-B ended on FEFE and Specials was a
                    // split Block of FEFF & FFF0..FFFD.
                    // Since Unicode 3.2, blocks were contiguous xxx0..yyyF:
                    // https://www.unicode.org/reports/tr28/tr28-3.html#database
                    // The normative blocks defined in Blocks.txt have been adjusted slightly,
                    // in accordance with Unicode Technical Committee decisions.
                    // - Every block starts and ends on a column boundary.
                    //   That is, the last digit of the first code point in the block is always 0,
                    //   and the last digit of the final code point in the block is always F.
                    // - Every block is contiguous. [...]
```

unicodetools/src/main/java/org/unicode/text/utility/Utility.java
```diff
@@ -1468,7 +1468,10 @@ public static String getMostRecentUnicodeDataFile(
                 }
                 Path path = Settings.UnicodeTools.getDataPath(base, element);
                 if (path != null) {
-                    result = path.resolve(parts[2] + fileType).toString();
+                    var filePath = path.resolve(parts[2] + fileType);
+                    if (filePath.toFile().exists()) {
+                        result = path.resolve(parts[2] + fileType).toString();
+                    }
```
markusicu, line 1474: why not
```suggestion
                    var filePath = path.resolve(parts[2] + fileType);
                    if (filePath.toFile().exists()) {
                        result = filePath.toString();
                    }
```
?

unicodetools/src/main/java/org/unicode/text/utility/Utility.java
```diff
@@ -1500,6 +1500,11 @@ public static String getMostRecentUnicodeDataFile(
                             + fileType
                             + "'");
         }
+        if ((versionString.startsWith("2.") || versionString.startsWith("3.0"))
+                && filename.startsWith("Prop")) {
+            return null; // Ignore the property dumps for now, as we do not have parsing logic for
+            // them.
```
markusicu, line 1506:
```suggestion
            // Ignore the property dumps for now, as we do not have parsing logic for them.
            return null;
```

unicodetools/src/main/java/org/unicode/text/utility/Utility.java
eggrobin: Because I was tired when I got to that point, I think.

PR #1087, opened 2025-04-09
Supply the field indices and correct the types of NormalizationCorrections data

eggrobin:
Before this change, all three normalization_correction_* pseudoproperties have the value `96FB` (that is, a four-character string) for U+F951.

This was not a problem for UCDXML as that one parses the file independently, https://github.com/unicode-org/unicodetools/blob/762843832ca6d9e05bbc0fa09f7af9afb575e642/unicodetools/src/main/java/org/unicode/xml/UCDDataResolver.java#L189-L197.

Also fix UnicodeProperty getSet on string-valued or miscellaneous properties of strings (although UnicodeProperty still leaves a lot to be desired for properties of strings; in particular, it has no way to get the value for a string!).

### DIFF
--- a/unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java
+++ b/unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java
@@ -799,7 +799,8 @@ public class IndexUnicodeProperties extends UnicodeProperty.Factory {
 
         @Override
         protected boolean hasStrings() {
-            return _getRawUnicodeMap().stringKeys() != null && !_getRawUnicodeMap().stringKeys().isEmpty();
+            return _getRawUnicodeMap().stringKeys() != null
+                    && !_getRawUnicodeMap().stringKeys().isEmpty();
         }
 
         private UnicodeSet getDiffSet() {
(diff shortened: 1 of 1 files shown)


markusicu (approved):

eggrobin:
> Will UCDXML be able to call the regular Unicode Tools functions after this, rather than parsing it independently?

That is a question for @jowilco, but of course I would like to minimize the number of UCD parsers we spawn here…

jowilco:
> > Will UCDXML be able to call the regular Unicode Tools functions after this, rather than parsing it independently?
> 
> That is a question for @jowilco, but of course I would like to minimize the number of UCD parsers we spawn here…

@markusicu , @eggrobin 
In theory, yes I could read the values using IndexUnicodeProperties.getResolvedValue(UcdProperty.normalization_correction_*, codepoint), however;
1. There are only 6 code points, and I don't "know" what these are without a source of truth (like parsing NormalizationCorrections.txt). Parsing NormalizationCorrections.txt would put me back where I am now. Obviously, I could run the full 0x0-0x10FFFF range, but...
2. The <normalization-corrections> section of UCDXML is separated from the main <repertoire> section. There are comparable sections for <blocks>, <named-sequences>, etc. All of these sections are implemented by parsing the corresponding source file. Most of them are not single code point -> value(s). I'm therefore not sure it makes too much sense in implementing a different path for <normalization-corrections> just so I can use IndexUnicodeProperties.

eggrobin:
> I'm therefore not sure it makes too much sense in implementing a different path for just so I can use IndexUnicodeProperties.

Fair point. Effectively, it would mean removing one parser, but adding something equivalent to MakeUnicodeFiles (the UCD file generator), and that way madness also lies.

PR #1089, opened 2025-04-09
Do not fall back when looking for 4.1 data or later

### LINKED ISSUE #1088: We shouldn’t fall back on older files for modern versions of Unicode
The tools think that kGB7 still exists, because while the kGB7.txt file produced by splitunihan was removed, we fall back to older files.

This dates back to releases as deltas that did not republish unmodified files; but we should not do that for modern files.

eggrobin:
Fix #1088.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/text/utility/Utility.java
+++ b/unicodetools/src/main/java/org/unicode/text/utility/Utility.java
@@ -1451,6 +1451,11 @@ public final class Utility implements UCD_Types { // COMMON UTILITIES
             if (version != null && version.compareTo(currentVersion) < compValue) {
                 continue;
             }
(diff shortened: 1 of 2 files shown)


unicodetools/src/main/java/org/unicode/text/utility/Utility.java
```diff
@@ -1451,6 +1451,14 @@ public static String getMostRecentUnicodeDataFile(
             if (version != null && version.compareTo(currentVersion) < compValue) {
```
markusicu, line 1451: FYI: Hopefully in the future we can completely stop looping over the versions in the searchPath...

markusicu (approved):

PR #1090, opened 2025-04-09
Version is not age

eggrobin:
[DRAFT to run CI]

### DIFF
--- a/unicodetools/src/main/java/org/unicode/text/utility/Utility.java
+++ b/unicodetools/src/main/java/org/unicode/text/utility/Utility.java
@@ -37,6 +37,7 @@ import java.util.Collection;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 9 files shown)


unicodetools/src/main/java/org/unicode/text/utility/Utility.java
```diff
@@ -928,6 +930,34 @@ public static String join(long[] array, String divider) {
...
+            }
+        }
+        return null;
+    }
+
+    public static VersionInfo getVersionFollowing(VersionInfo version) {
+        VersionInfo result = null;
+        for (VersionInfo candidate : UNICODE_VERSIONS) {
+            if (candidate.compareTo(version) > 0) {
+                return result = candidate;
```
markusicu, line 951: This looks wrong: It returns the first thing in the list if it's > candidate.
I think you want to remove the "return" here and just remember the last candidate > version going back in time.

eggrobin: Ah yes, it looks like I forgot to remove the `return` (returning an assignment to a local variable definitely doesn’t look intentional!).

markusicu (approved):

PR #1095, opened 2025-04-15
Tutg 𑀈

eggrobin:
* L2/22-260 and L2/25-113
* https://github.com/unicode-org/sah/issues/641
* https://github.com/unicode-org/utc-release-management/issues/203


roozbehp:
I would really appreciate more readable names for pull requests. This one is really cryptic.

PR #1096, opened 2025-04-15
One more Tangut ideograph

eggrobin:
* WG2-5303
* https://github.com/unicode-org/sah/issues/645
* https://github.com/unicode-org/utc-release-management/issues/205


unicodetools/data/ucd/dev/JurchenSources.txt
markusicu: Why are we adding JurchenSources.txt when adding one Tangut ideograph?

PR #1097, opened 2025-04-15
Better handling of history

eggrobin:
Do not show properties that are gone unless we are showing a slice of history where they exist.

When we show nontrivial history, always show version spans, since it is useful to know when the property was introduced, even if it never changed for a character.


markusicu (approved):
squinted at it...

PR #1098, opened 2025-04-15
The real 1.1.5

eggrobin:
[on top of #1097]

* Use the real 1.1.5, not one from which pre-move Hangul has been expurged.
* Add the reconstructed 1.0.0 and 1.0.1, correcting errors in the reconstruction (these have been communicated to the reconstructor).
* Correctly parse Unicode 2 CJK compatibility ranges.
* Show Name among the properties, since it may now have interesting history.
* Drop the id thing at the top, which is based on outdated data and showing that data inverted anyway.


eggrobin:
(Needs some workarounds in PropertyParsingInfo to apply `4E00;<CJK IDEOGRAPH REPRESENTATIVE>;Lo;0;L;;;;;N;;;;;` to the CJK Unified and Compatibility ideographs ranges, and to inject the PUA. While I am there, I should fix PropertyParsingInfo so that it does not think the compatibility ideographs were nameless in Unicode 2.)

markusicu (approved):

PR #1105, opened 2025-04-25
Improvements to property history

eggrobin:
## Old dumps.

PropList-2.1.8.txt includes a diff between Unilib properties and what it calls « UnicodeData ». The « UnicodeData » values appear to prefigure the 3.1 derivation for, e.g., Alphabetic, but without the Other_Alphabetic exceptions. Ignoring the Discrepancy heading means we currently take the union of the Unilib and « UnicodeData » sets, which is nonsense. The Unilib set is the more useful one for historical investigation, as it is consistent with the files included in 2.1.5 and 2.1.9 (whereas taking the « UnicodeData » set introduces massive discrepancies that are immediately reverted in 2.1.9).
image

Currently, `\P{U2.1.8:Alphabetic=@U2.1.9:Alphabetic@}` has 21226 code points, and `\P{U2.1.8:Alphabetic=@U2.1.5:Alphabetic@}` has 21236 code points, whereas `\P{U2.1.9:Alphabetic=@U2.1.5:Alphabetic@}` has 10 code points. With this change, U2.1.9:Alphabetic and U2.1.8:Alphabetic are identical.

## History tables.

Currently, the Unihan history for U+3400 looks like (A) below.
Because of multiple changes to a few properties, many other properties are squished into a column so narrow that the null does not fit on one line. In addition, there is no relation between the columns: synchronous changes are not aligned, and aligned changes may be decades apart.

With this change, the history of the same properties is displayed as in (B), both more compact and more informative as to the synchronicity of changes and the lifetime of properties.

| (A) | (B) |
|---|---|
| image image | image image |

### DIFF
--- a/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeUtilities.java
+++ b/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeUtilities.java
@@ -1702,9 +1702,12 @@ public class UnicodeUtilities {
             VersionInfo first;
             VersionInfo last;
             ArrayList<String> values;
+            int span;
         }
         final boolean isMultivalued = getFactory().getProperty(propName).isMultivalued();
         List<PropertyAssignment> history = new ArrayList<>();
+        int prehistoricSpan = 0;
+        int posthistoricSpan = 0;
         if (getFactory().getProperty(propName)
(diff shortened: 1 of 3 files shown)


markusicu (approved):
I didn't try to really follow the logic... amazing effects...

PR #1112, opened 2025-04-30
UCA 17.0 d1..d4

markusicu:
UCA 17 updates from Ken

The UCA check fails too now. I will generate the test files after Unihan is updated (PR #1111).


Ken-Whistler (approved):
Changes look correct. No opinion about the invariant test failures, which mostly seem unrelated.

PR #1113, opened 2025-04-30
[WIP] Identifier_Type updates for Unicode 17.0 from UTC183

josh-hadley:
- UTC-183-C33: Change Identifier_Type for 40 characters listed in L2/25-032R to Recommended.
- UTC-183-C34: Change Identifier_Type of 392 characters marked Proposed:Uncommon_Use, Proposed:Technical or Proposed:Obsolete in document L2/25-034R4 to Uncommon_Use, Technical, or Obsolete, respectively
- UTC-183-C35: Change Identifier_Type of 753 characters from document L2/25-033R4 to their proposed new values.

PR #1116, opened 2025-05-01
Tangut Nushu rename fields & update data for 17 beta

markusicu:
Also takes care of
- [[183-A63](https://www.unicode.org/cgi-bin/GetL2Ref.pl?183-A63)] Action Item for Michel Suignard, Ken Whistler, PAG: In TangutSources.txt, change the tag kRSTUnicode to kTGT_RSUnicode. For Unicode Version 17.0. See [L2/25-087](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/25-087) item 1.9.
- [[183-A64](https://www.unicode.org/cgi-bin/GetL2Ref.pl?183-A64)] Action Item for Michel Suignard, Ken Whistler, PAG: In NushuSources.txt, change the tag kSrc_NushuDuben to kNSHU_DubenSrc, and change the tag kReading to kNSHU_Reading. For Unicode Version 17.0. See [L2/25-087](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/25-087) item 1.9.


unicodetools/src/main/java/org/unicode/xml/UCDPropertyDetail.java
```diff
@@ -1884,9 +1884,9 @@ public class UCDPropertyDetail {
                     true,
                     false,
                     true);
-    public static UCDPropertyDetail kRSTUnicode_Detail =
+    public static UCDPropertyDetail kTGT_RSUnicode_Detail =
```
jowilco, line 1887: @markusicu - I think that the best way of handling this is to add a line for maxVersion as VersionInfo.getInstance(16, 0, 0), then have a new UCDPropertyDetail for kTGT_RSUnicode_Detail (and others) with the minVersion as VersionInfo.getInstance(17, 0, 0). In that way, we'll still be able to build older versions of UCDXML.

eggrobin:
> The second commit adds a hack. Without it, UCD.Main crashed for me. Hopefully we can figure out soon what's going on there.

Something is seriously broken here, besides the fact that multivaluedness is not new, the values of the relevant properties did not change at UTC-183. And this code has been working recently, since I changed it recently.

BTW, I think you forgot to run GenerateEnums, though I do not understand how this could cause that issue.

unicodetools/src/main/java/org/unicode/xml/UCDPropertyDetail.java
jowilco: Having said that, if you want to complete this PR as is, I'll go back and implement this change after the fact.

jowilco (approved):

eggrobin:
> I think you forgot to run GenerateEnums

The CI check for that is after one that we intentionally broke.

I think we urgently need to get at least the first two commits of https://github.com/unicode-org/unicodetools/pull/1113 in so that we have some tests again.

eggrobin:
> I think you forgot to run GenerateEnums

Ah nevermind, I see UcdProperty and UcdPropertyValues are updated.

eggrobin:
> UCD.Main crashed for me

We know that UCD.Main works at head in the CI setup, because the the Make Unicode Files command passes: https://github.com/unicode-org/unicodetools/actions/runs/14784230778/job/41509553862.

So if it is broken, it is either broken on your machine, or broken by this change.

This change really does not look like it does anything here. Do you happen to have a stale Generated directory?

markusicu:
> So if it is broken, it is either broken on your machine, or broken by this change.

I don't know what to say. I am not happy about this hack.
I ran UCD.Main as always, with the existing configuration in my IntelliJ.

I just set breakpoints here and started the tool again. It does get here.

UCD.java
```
    public byte getNumericType(int codePoint) {
        if (hanExceptions == null) {
            populateHanExceptions();
        }
```
calls
```
    private void populateHanExceptions() {
        IndexUnicodeProperties iup = IndexUnicodeProperties.make(versionInfo);
        hanExceptions = new IntMap();
        try {
            populateHanExceptions(iup.getProperty(UcdProperty.kPrimaryNumeric));
```
where versionInfo is 16, not 17.

This then calls the hacked function where code=0x5146 and valueString="1000000 1000000000000".

I am not making this up...

Stack:
```
populateHanExceptions:550, UCD (org.unicode.text.UCD)
populateHanExceptions:567, UCD (org.unicode.text.UCD)
getNumericType:590, UCD (org.unicode.text.UCD)
_getValue:2181, ToolUnicodePropertySource$ToolUnicodeProperty (org.unicode.text.UCD)
getValue:287, UnicodeProperty (org.unicode.props)
_getUnicodeMap:646, UnicodeProperty (org.unicode.props)
getUnicodeMap_internal:632, UnicodeProperty (org.unicode.props)
getSet:493, UnicodeProperty (org.unicode.props)
getSet:466, UnicodeProperty (org.unicode.props)
getSet:429, UnicodeProperty (org.unicode.props)
generateValueAliasFile:1134, MakeUnicodeFiles (org.unicode.text.UCD)
generateFile:557, MakeUnicodeFiles (org.unicode.text.UCD)
generateFile:544, MakeUnicodeFiles (org.unicode.text.UCD)
main:84, MakeUnicodeFiles (org.unicode.text.UCD)
main:183, Main (org.unicode.text.UCD)
```

> Do you happen to have a stale Generated directory?

I don't know what that means. Generated is output. Why would it matter what's in there?
I did blow away the BIN files for 17 before running the tool.

eggrobin:
> I don't know what that means. Generated is output. Why would it matter what's in there?

Because it’s a cache.

> I did blow away the BIN files for 17 before running the tool.

Hmm, that _should_ be enough, I think…

What happens if you revert the hack commit and push? Does MakeUnicodeFiles work on the CI setup?

eggrobin:
Oh wait

> where versionInfo is 16, not 17.

So if you only blew the 17 BIN files, you may have old 16 BIN files.

unicodetools/src/main/java/org/unicode/xml/UCDPropertyDetail.java
markusicu: Not sure what you mean. This code uses named objects from UcdProperty which are changing names. It cannot compile with `UcdProperty.kRSTUnicode` because that object name no longer exists.

eggrobin:
(My general recipe is to _always_ clear the whole Generated directory when changing things in the UCD, it is too easy to end up in an inconsistent state otherwise; see https://github.com/unicode-org/unicodetools/blob/main/docs/pipeline.md#regenerate-ucd.)

markusicu:
> I did blow away the BIN files for 17 before running the tool.

I might have it. I just blew away the entire BIN folder, and now the debugger didn't reach the breakpoints.
I guess I had cached data in an old format.
After dinner, I will try to remove the second commit.

markusicu:
It all works fine without the hacky second commit, so I backed that out. PTAL.

I hate the BIN folder. I have been tripped up by it many times.
Does it save so much time that it's worth the problems?
Can we add a flag to ignore it? And make that the default?

eggrobin (approved):

markusicu:
FYI

> > I think you forgot to run GenerateEnums

I did run that. More than once.
I ran it, it produced its output, but then crashed because it had modified its own code.
Fixed compiler errors, ran GenerateEnums again, generated output and also broke again.
After one or two more rounds, it was finally stable.

> The CI check for that is after one that we intentionally broke.
> 
> I think we urgently need to get at least the first two commits of #1113 in so that we have some tests again.

I advised Josh today to first adjust the security data for the repertoire changes, hoping that that would fix the CI failures, and then make the changes for the security-specific action items.

unicodetools/src/main/java/org/unicode/xml/UCDPropertyDetail.java
jowilco: Ugh. That will mean that we won't be able to build 16.0.0 or previous UCDXML files without some hackery...

jowilco (approved):

eggrobin:
> I hate the BIN folder. I have been tripped up by it many times.
Does it save so much time that it's worth the problems?

Let’s try various workloads.

#### Regenerating the UCD:
But of course, if you are changing the UCD, you are often invalidating the cache! Running this command without clearing the BIN files is only useful when the only thing you are changing is the derivations of derived properties.
```powershell
Measure-Command { rm .\Generated\* -recurse -force; mvn compile exec:java '-Dexec.mainClass="org.unicode.text.UCD.Main"'  '-Dexec.args="build MakeUnicodeFiles"' -am -pl unicodetools  "-DCLDR_DIR=..\cldr\"  "-DUNICODETOOLS_GEN_DIR=Generated"  "-DUNICODETOOLS_REPO_DIR=." }
```
```
TotalSeconds      : 100.6704397
```

```powershell
Measure-Command { mvn compile exec:java '-Dexec.mainClass="org.unicode.text.UCD.Main"'  '-Dexec.args="build MakeUnicodeFiles"' -am -pl unicodetools  "-DCLDR_DIR=..\cldr\"  "-DUNICODETOOLS_GEN_DIR=Generated"  "-DUNICODETOOLS_REPO_DIR=." }
```
```
TotalSeconds      : 55.1698726
```

#### Running the invariant tests:
```powershell
Measure-Command { rm .\Generated\* -recurse -force; mvn test -am -pl unicodetools "-DCLDR_DIR=$(gl|split-path -parent)\cldr\"  "-DUNICODETOOLS_GEN_DIR=$(gl|split-path -parent)\unicodetools\Generated\"  "-DUNICODETOOLS_REPO_DIR=$(gl|split-path -parent)\unicodetools\" "-DUVERSION=17.0.0" "-Dtest=TestTestUnicodeInvariants" -DfailIfNoTests=false -DtrimStackTrace=false }
```
```
TotalSeconds      : 99.0030359
```
```powershell
Measure-Command { mvn test -am -pl unicodetools "-DCLDR_DIR=$(gl|split-path -parent)\cldr\"  "-DUNICODETOOLS_GEN_DIR=$(gl|split-path -parent)\unicodetools\Generated\"  "-DUNICODETOOLS_REPO_DIR=$(gl|split-path -parent)\unicodetools\" "-DUVERSION=17.0.0" "-Dtest=TestTestUnicodeInvariants" -DfailIfNoTests=false -DtrimStackTrace=false }
```
```
TotalSeconds      : 89.8550277
```
#### Getting ready to serve the JSPs with full history
But note that the JSPs use a different format for the BIN files, so they _must_ be regenerated between building the JSPs and going back to normal UCD work.

After adding, at the end of [`loadUcdHistory`](https://github.com/unicode-org/unicodetools/blob/ce39bf465f690c47cd156a533698e1ac7f47bd7e/unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java#L920),
```java
        if (earliest == null) {
            System.exit(0);
        }
```

```powershell
Measure-Command { rm .\Generated\* -recurse -force; mvn "-DCLDR_DIR=..\cldr\"  "-DUNICODETOOLS_GEN_DIR=Generated"  "-DUNICODETOOLS_REPO_DIR=." org.eclipse.jetty:jetty-maven-plugin:run "-Djetty.port=8081"; }
```
```
TotalSeconds      : 880.8308483
```

```powershell
Measure-Command { mvn "-DCLDR_DIR=..\cldr\"  "-DUNICODETOOLS_GEN_DIR=Generated"  "-DUNICODETOOLS_REPO_DIR=." org.eclipse.jetty:jetty-maven-plugin:run "-Djetty.port=8081"; }
```
```
TotalSeconds      : 143.1834969
```

markusicu:
> > I hate the BIN folder. I have been tripped up by it many times.
> > Does it save so much time that it's worth the problems?
> 
> Let’s try various workloads.

Thanks. So it does save time, but I am not sure it's worth the hassle.

> But of course, if you are changing the UCD, you are often invalidating the cache! Running this command without clearing the BIN files is only useful when the only thing you are changing is the derivations of derived properties.

Normally it suffices to blow away the BIN files for just the latest Unicode version, because we are not changing old versions.

Sometimes however, like happened to me here, other work has changed the data structures, and there is no logic to detect that, so the tools happily load incompatible data.

Do you know why running UCD.Main ends up populating the BIN files for every version? Why does it not just build the latest, and maybe the last, files, when generating the latest .txt files? If there is something unnecessary being computed, then we might be able to speed up the process a lot by no longer doing that, even without BIN files.

eggrobin:
> Thanks. So it does save time, but I am not sure it's worth the hassle.

I agree that it is not worth the time it loses, except as used in the JSPs.

One piece of unicodetools lore you might not have heard me rant about yet is that when following the process in https://github.com/unicode-org/unicodetools/blob/main/docs/pipeline.md, regenerating the UCD produces correct files for Word_Break, and Sentence_Break, _but_ incorrect BIN files for them (presumably because the BIN files, which are cleared at the beginning of the procedure, are computed from the derived files before the tools regenerate them). So systematically the addition-comparison tests fail for these two properties on the machine where you perform the procedure, unless and until you reclear the cache. I have since learned to ignore local comparison failures on these two properties when they involve the default values, but this has deeply confused me a few times.

> Do you know why running UCD.Main ends up populating the BIN files for every version?

Populating _all_ the BIN files for _every_ version should be quite rare unless you are building the JSPs (and it takes about a quarter of an hour on a substantial machine, see above).
Looking at _some_ old versions of _some_ properties happens for various reasons (tests, etc.). But that does not take very long, even if it creates a few directories.

markusicu:
> ... regenerating the UCD produces correct files for Word_Break, and Sentence_Break, _but_ incorrect BIN files for them (presumably because the BIN files, which are cleared at the beginning of the procedure, are computed from the derived files before the tools regenerate them).

ugh! :-(

> > Do you know why running UCD.Main ends up populating the BIN files for every version?
> 
> Populating _all_ the BIN files for _every_ version should be quite rare unless you are building the JSPs (and it takes about a quarter of an hour on a substantial machine, see above). Looking at _some_ old versions of _some_ properties happens for various reasons (tests, etc.). But that does not take very long, even if it creates a few directories.

I just tried this:
- clear out all of BIN
- run ... UCD.Main "-Dexec.args=version 17.0.0 build MakeUnicodeFiles" ...
- now BIN contains
  - a folder for 16 with 61 files
  - a folder for 17 with 99 files
  - a .bin file like "UCD_Data17.0.0.bin" for each version from 1.1 to 17

markusicu:
> > now BIN contains
> > a folder for 16 with 61 files
> > a folder for 17 with 99 files
> 
> That means we looked at 99 properties for the dev version, 61 for the last released version. That does not seem altogether unreasonable.

Do we _need_ to read dozens of last-version properties in order to build dev-version properties?

> > a .bin file like "UCD_Data17.0.0.bin" for each version from 1.1 to 17
> 
> I think that one is a completely different format, produced by the 1996 UnicodeData parser. Probably needed for, e.g., deriving the Age property, but also probably cheap.

Maybe.
We should not need to read previous versions at all for building DerivedAge.txt, given that we already have a file from at least the last version. We should be able to:
- parse the dev=latest file
- recent additions: take the set of `[[:age=NA:]&[:^gc=Cn:]]` and map them to the latest version
- recent removals: take the set of `[[:^age=NA:]&[:gc=Cn:]]` and map them to "NA"

eggrobin:
> We should not need to read previous versions at all for building DerivedAge.txt, given that we already have a file from at least the last version. We should be able to:

I agree. But the way that derivation is done at the moment (and the way it has been [for more than twenty years](https://github.com/unicode-org/unicodetools/blame/5c930e8c1b8ed10f91529c82fa5dcf74d19f51b1/org/unicode/text/UCD/ToolUnicodePropertySource.java#L462-L476)) is by looking back at every major or minor version since Unicode 1.1 using a thirtyish-year-old parser. (Of course, back then that meant looking at 7 versions, not 27).

In fact, I just tweaked some of that code last month, when I switched us to the Real UnicodeData 1.1.5:
https://github.com/unicode-org/unicodetools/blob/c2c5ba99b9d390cc4804ed21ccbcfba0c55be328/unicodetools/src/main/java/org/unicode/text/UCD/ToolUnicodePropertySource.java#L2246-L2257

But note that if we constructed Age incrementally like that, I would probably like to have a test that checks that it matches the actual historical assignments, so _something_ would still have to look it.

eggrobin:
I think the discussion of performance improvements is a bit of a distraction though. Performance in these tools is disastrous in various places, and I would like to do something about it (a striking example is that queries on multivalued Unihan properties are a good way to take the servers of the JSPs for a ride), but this is immaterial to the question of using the cache, since the playbook literally says « purge the cache _every time_ you run the command that benefits significantly from the cache ». So people actually following the process (well, that’s mostly me) do not benefit from the cache (and in fact, I get bitten by the Word_Break stupidity above).

eggrobin:
We could keep the cache in CI via a flag, since CI is not supposed to change the UCD and checks that before it does anything else; that could avoid losing a few tens of seconds on the workflows that run several commands in sequence. But I agree that when invoking the commands by hand, reading the cache is just a nuisance.

markusicu:
> We could keep the cache in CI via a flag, since CI is not supposed to change the UCD and checks that before it does anything else

+1

> I agree that when invoking the commands by hand, reading the cache is just a nuisance.

+1

markusicu:
> > We should not need to read previous versions at all for building DerivedAge.txt, given that we already have a file from at least the last version. We should be able to:
> 
> I agree.

- https://github.com/unicode-org/unicodetools/issues/1124

markusicu:
> We could keep the cache in CI via a flag, since CI is not supposed to change the UCD and checks that before it does anything else; that could avoid losing a few tens of seconds on the workflows that run several commands in sequence. But I agree that when invoking the commands by hand, reading the cache is just a nuisance.

- https://github.com/unicode-org/unicodetools/issues/1125

PR #1123, opened 2025-05-02
Correctly parse Numeric_Value since 1.0

### LINKED ISSUE #1118: IndexUnicodeProperties fails to load Numeric_Value prior to Unicode 5.1
Because it uses DerivedNumericValues.txt rather than UnicodeData.txt, and before 5.1 DerivedNumericValues.txt had a different format.

See, in the tools, the properties of U+0031 1 DIGIT ONE:
Image

Note that this hits a bug in the display code whereby we assume that properties do not pop in and out of existence, so that it looks like Numeric_Value is removed after 17.0. After fixing that bug, we can see that NV is all-NaN in 5.0 and then somehow all null prior to that.

Image

eggrobin:
Fix #1118.

When it exists, use DerivedNumericValues to tell us:
1. that there is a numeric value,
2. what the numeric value is, if it does not come from UnicodeData.

But use UnicodeData as the source of truth if available. This is preferable because of the lack of star in https://www.unicode.org/reports/tr44/#Derived_Extracted, but also because it is the only thing that works in the past.

Before DerivedNumericValues, use UnicodeData alone (at that time, Unihan did not contribute to Numeric_Value, so this is correct).

In addition, this allows us to read the data we would want to write back (such as 2/12 for U+109F7), which seems like the right way to go in the future. Right now we are reverse-engineering that from the name when generating UnicodeData, which is very silly.
https://github.com/unicode-org/unicodetools/blob/c2c5ba99b9d390cc4804ed21ccbcfba0c55be328/unicodetools/src/main/java/org/unicode/text/UCD/MakeUnicodeFiles.java#L2220-L2257


unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
```diff
@@ -227,6 +238,17 @@ public String getFileName(VersionInfo ucdVersionRequested) {
         }
     }
 
+    public int getFieldNumber(VersionInfo ucdVersionRequested) {
```
markusicu, line 241: Could we optimize this a little?

For example, if fieldNumbers.size() == 1 then just return the one and only value.

Maybe even better, once the parsing info is all built, populate an `int singleFieldNumber` with the one and only value if there is exactly one. If there are two or more, then set that to -1.

eggrobin: > For example, if fieldNumbers.size() == 1 then just return the one and only value.

Sure. Done.

> Maybe even better, once the parsing info is all built, populate an int singleFieldNumber with the one and only value if there is exactly one. If there are two or more, then set that to -1.

That is probably not the direction this wants to go in. This is clearly in dire need of refactoring, because this logic of version-dependent chunk of parsing info is now triplicated (we do this for field number, file, default value). This should just get transposed into a type that has (field number, file, default value, whatever else we want to make version-dependent), and then a map keyed on version. I don’t think we have much to gain from making the map sparse (the maps are sparse here because they match contents of the source files, which is rarely a good guide for data structures).

But that refactoring isn’t going to happen today.

markusicu (dismissed):

markusicu (approved):

PR #1126, opened 2025-05-02
change CaseFolding example from MASSE to FUSS

markusicu:
[[183-A59](https://www.unicode.org/cgi-bin/GetL2Ref.pl?183-A59)] Action Item for Markus Scherer, PAG: In CaseFolding.txt, change the full case folding example from "MASSE"/"Maße" to "FUSS"/"Fuß". For Unicode Version 17.0. See [L2/25-087](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/25-087) item 1.6.


echeran (approved):
RSLGTM

PR #1128, opened 2025-05-05
NamesList in IUP, drop the outdated JSP copy of NamesList.txt

eggrobin:
Besides subheader which we already had (but only at 15.1), this gives us versioned informal aliases, cross references, comments (effectively the successor to ISO_Comment), and any notice under the subheader.

It also means we can show the β subheads with highlighting when we show the β characters with highlighting in UnicodeSet lists.


unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
```diff
@@ -324,8 +324,33 @@ public void put(
                     } else {
                         insertedValue = value;
                     }
-                    if (nextVersion != null && Objects.equals(insertedValue, nextValue)) {
-                        insertedValue = IndexUnicodeProperties.UNCHANGED_IN_BASE_VERSION;
+                    if (nextVersion != null) {
+                        if (merger != null) {
+                            var oldValue = data.get(codepoint);
+                            if (oldValue != null) {
+                                // If there is a merger and it is about to do something something,
```
markusicu, line 331: I think you want to remove one "something".

markusicu (commented):
Looks like good work, but I didn't review the parser closely...

markusicu (approved):
rubber stamp...

PR #1129, opened 2025-05-06
Bopomofo -> Limited_Use

josh-hadley:
[UTC-183-C37] Consensus: Change Bopomofo to "Limited Use" in UAX #31. For Unicode Version 17.0. See L2/25-087 item 5.5.
[UTC-183-A79] Action Item for Josh Hadley, PAG: Update Identifier_Status and Identifier_Type properties according to the change of status of Bopomofo to “Limited Use”. For Unicode Version 17.0. See L2/25-087 item 5.5.


josh-hadley (commented):
@asmusf First pass at this just assigning Bopomofo as Limited_Use resulted in a couple of double assignments from previous updates. Are these OK per the consensus? If not I think I can use `override` to force them to Limited_Use only.

asmusf:
Hmmm. Not maintaining info on "obsolete" for "limited_use" scripts is acceptable, though discarding any vetted information is always regrettable, but I get that we may not want the burden of maintaining a consistent analysis. (While this means we inadvertently force people who want to support a limited_use script to start from scratch, we can survive this one for now). 

However, the Not_XID seems something that is "stronger" than "limited_use". It seems quite wrong to have that information "hidden", because we do want people to be able to add Limited_Use scripts to their own identifiers and that should not force them to recreated everything.

However, again, this data is discoverable, and we can sit down and have a discussion on whether Not_XID should be handled differently. For now, let's just match all other limited_use scripts.

josh-hadley:
> For now, let's just match all other limited_use scripts.

OK I think this means "no double assignments"? Please confirm.

If so, that will translate into:
- remove the prior explicit `Obsolete` assignment for U+312E
- override the `Not_XID` assignments of U+02EA..02EB

asmusf:
@josh-hadley , looking into the Identifier_Type.txt file for 16.0.0 there are many, many instances of multiple assignments combining Limited_Use with Obsolete and Not_XID. That means you should not remove the Obsolete or Not_XID.

josh-hadley:
> you should not remove the Obsolete or Not_XID.

OK, great, then this is ready to go 😄. Thanks!

markusicu:
@josh-hadley ok if I merge on your behalf?

PR #1153, opened 2025-05-16
Address feedback on character.jsp from the public mailing list

eggrobin:
See https://corp.unicode.org/pipermail/unicode/2025-May/011446.html from @neatnit and my reply https://corp.unicode.org/pipermail/unicode/2025-May/011447.html.

* Drop the links to the blank properties.jsp (blanked in #1018).
* Show links to the individual character property pages for string-valued properties, e.g., for U+FDFA,  image or for U+00DF, image
* Link to the null query from _null_ values.

### DIFF
--- a/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeUtilities.java
+++ b/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeUtilities.java
@@ -40,6 +40,8 @@ import java.util.TreeSet;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
 import org.unicode.cldr.tool.TablePrinter;
 import org.unicode.cldr.util.Predicate;
 import org.unicode.cldr.util.UnicodeSetPrettyPrinter;
@@ -90,7 +92,7 @@ public class UnicodeUtilities {
 
         String CONTENT_RULES = "'>' > '&gt;' ;";
 
-        String HTML_RULES = BASE_RULES + CONTENT_RULES + "'\"' > '&quot;' ; '' > '&apos;'";
+        String HTML_RULES = BASE_RULES + CONTENT_RULES + "'\"' > '&quot;' ; '' > '&apos;' ;";
 
         toHTMLInput = Transliterator.createFromRules("any-xml", HTML_RULES, Transliterator.FORWARD);
 
@@ -514,7 +516,8 @@ public class UnicodeUtilities {
         }
     }
 
-    public static String getIdentifier(String script, boolean showDevProperties) {
+    public static String getIdentifier(
+            String script, boolean showDevProperties, List<String> originalParameters) {
         StringBuilder result = new StringBuilder();
         UnicodeProperty scriptProp = getFactory().getProperty("sc");
         UnicodeSet scriptSet;
@@ -560,7 +563,8 @@ public class UnicodeUtilities {
                 showSet(
                         allowed,
                         showDevProperties,
-                        new CodePointShower("", "", showDevProperties, true, false, false),
+                        new CodePointShower(
+                                "", "", showDevProperties, true, false, false, originalParameters),
                         result);
             }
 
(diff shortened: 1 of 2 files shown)


markusicu:
CI check failures...

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeUtilities.java
```diff
@@ -560,7 +563,8 @@ public static String getIdentifier(String script, boolean showDevProperties) {
                 showSet(
                         allowed,
                         showDevProperties,
-                        new CodePointShower("", "", showDevProperties, true, false, false),
+                        new CodePointShower(
```
markusicu, line 566: 🚿

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeUtilities.java
```diff
@@ -874,25 +895,32 @@ private static String getName(
...
                         + "</a></code>";
         return string;
     }
 
-    private static String getHex(String string, String separator, boolean ucdFormat) {
+    private static String getHex(
+            String string, String separator, boolean ucdFormat, List<String> additionalParameters) {
         StringBuilder result = new StringBuilder();
         int cp;
         for (int i = 0; i < string.length(); i += UTF16.getCharCount(cp)) {
```
markusicu, line 919: FYI: At some point, we could change this kind of code to use [CharSequence.codePoints()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/CharSequence.html#codePoints()).

markusicu (commented):
as usual, i am not trying to understand every aspect...

eggrobin:
> CI check failures...

Ah, we are using an older JDK than the one I have on my machine…

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeUtilities.java
eggrobin: Yes.

UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeUtilities.java
eggrobin: For the longest time I thought that it actually meant 🚿 (the list-unicodeset tool has a tendency of showering the user with code points!).

markusicu (approved):

NeatNit:
Thank you so much for addressing my feedback.

Will you add mirroring to bidi-c in this PR as well, or in a separate one?

My guess is it'd be something like this:

```java
                    int resIdx = Integer.parseInt(resOrdArray[i]);
                    String cp = arrInputCpSeq.get(resIdx);
                    int lev = Integer.parseInt(resLevArray[resIdx]);
                    String dir = (lev % 2) ? "rtl" : "ltr";
                    valDisplayChars += "<td dir=\"" + dir + "\">&#x" + cp + "</td>";
                    valDisplayCodes += "<td><pre>" + cp + "</pre></td>";
```

In place of:

https://github.com/unicode-org/unicodetools/blob/852f2859c3b07acf66ba44c3ab22de850409104c/UnicodeJsps/src/main/webapp/bidic.jsp#L210-L213

eggrobin:
> Will you add mirroring to bidi-c in this PR as well

I will deal with that separately, when I can find time.

Please don’t post code in issue comments, we can’t look at it anyway because you haven’t signed the CLA.

And the point of the bidi-c demo is not to be a layout engine (this is most obvious in that it shows unshaped Arabic and does nothing about rule [L3](https://www.unicode.org/reports/tr9/#L3)), nor to illustrate which characters the font happens to support mirroring for, but to be a versioned demo of the algorithm as specified, so of course it needs to look up the Bidi_Mirrored for the correct version of Unicode and flag those characters to which rule [L4](https://www.unicode.org/reports/tr9/#L4) applies.

It is not clear that it would be a good idea to actually cause the characters to mirror (or rather to maybe mirror if the font supports it) by setting the dir attribute; the display really shows the _characters_ in isolation, not the glyphs they would have (this is striking in Arabic where it is unshaped), so changing the glyphs can make it trickier to figure out who is who; in addition, for the characters that do not have a Bidi_Mirroring_Glyph, actual support for the RTL glyph is usually limited to specialized fonts, so the display would be unreliable.

PR #1155, opened 2025-05-20
write FractionalUCA_blanked.txt

markusicu:
... like running blankweights.sed over FractionalUCA.txt.

The generated FractionalUCA.txt contains all of the data for building the CLDR root collation data. Diffing it across Unicode versions is nearly unusable, because inserting any character into the sort order changes the primary weights of every following character. And the fractional byte sequence weights are subject to manual and automatic tweaks in their computation.

Therefore, for many years I have used the [blankweights.sed script](https://github.com/unicode-org/cldr/blob/main/tools/scripts/uca/blankweights.sed) to generate a modified file with “blanked weights” that preserves the sort order and the number of non-zero weights. I find diffing this file across versions valuable. See the [UCA tools docs](https://github.com/unicode-org/unicodetools/blob/main/docs/uca/index.md) for some more details.

For example:
```
FractionalUCA.txt
04C1; [62 2C, 05, 9B][, 8C, 05]	# Cyrl Lu	[2837.0020.0008][0000.0026.0002]	* CYRILLIC CAPITAL LETTER ZHE WITH BREVE

FractionalUCA_blanked.txt
04C1; [pp pp, ss, tt][, ss, tt]	# Cyrl Lu	[pppp.ssss.0008][0000.ssss.0002]	* CYRILLIC CAPITAL LETTER ZHE WITH BREVE
```

With these code changes, the Java code directly writes FractionalUCA_blanked.txt, just like it has also already written a FractionalUCA_SHORT.txt file (which omits the inline comments with allkeys.txt weights and character types & names). It should make the UCA-CLDR update workflow easier.

I am adding this file to the [CLDR root collation data files](https://github.com/unicode-org/cldr/tree/main/common/uca).
- https://github.com/unicode-org/cldr/pull/4720

I suggest reviewing one commit at a time. I intend to rebase & merge the three commits.


echeran:
If you could add a one line description for the purpose of this `_blanked.txt` file somewhere, whether in a Readme or a spec doc, that would be useful.

unicodetools/src/main/java/org/unicode/text/UCA/FractionalUCA.java
```diff
@@ -916,7 +936,7 @@ static void writeFractionalUCA(String filename) throws IOException {
             final String gcInfo = getStringTransform(chr, "/", ScriptTransform);
             final String scriptInfo = getStringTransform(chr, "/", GeneralCategoryTransform);
 
-            longLog.print(
+            output.frac.print(
```
echeran: just checking: Is this intended to be `output.frac.print` and not `output.fracPrint`?

unicodetools/src/main/java/org/unicode/text/UCA/FractionalUCA.java
```diff
@@ -975,12 +995,12 @@ static void writeFractionalUCA(String filename) throws IOException {
             highByteToScripts.addScriptsIn(np, sample);
             fractionalStatistics.printAndRecord(true, fakeString.next(), np, ns, nt, null);
 
-            longLog.print(
+            output.frac.print(
```
echeran: ditto

unicodetools/src/main/java/org/unicode/text/UCA/CEList.java
```diff
@@ -302,18 +302,30 @@ public static byte remap(int ch, byte type, int t) {
         return (byte) t;
     }
 
+    enum ToString {
```
echeran, line 305: optional: name is vague and therefore confusing. Rename for clarity.

possible options:
* `CEFormatStyle`
* `CEStyle`
* `CEToString`
* `CEPrintStyle`

unicodetools/src/main/java/org/unicode/text/UCA/FractionalUCA.java
```diff
@@ -270,26 +268,33 @@ String getBuffer(long val, boolean haveHigher) {
...
+            return option == ToString.DEFAULT
+                    ? Fractional.hexBytes(val)
+                    : Fractional.blankedHexBytes(val, blanked, null).toString();
         }
 
         long getValue(int zeroBasedLevel) {
             return key[zeroBasedLevel];
         }
 
+        enum ToString {
```
echeran, line 280: optional: ditto

echeran (dismissed):
rslgtm. all comments are optional

unicodetools/src/main/java/org/unicode/text/UCA/FractionalUCA.java
markusicu: It is. Note that the old code wrote to longLog instead of the DualWriter fractionalLog.
This code writes the comment after each mapping, and the comment goes to the regular file but not to the short file.
In the third commit, I move this code into an Output method which writes to frac and fracBlanked but not to fracShort.

unicodetools/src/main/java/org/unicode/text/UCA/FractionalUCA.java
markusicu: same thing, writes a comment which goes to output.frac (former longLog) but not to output.fracShort

unicodetools/src/main/java/org/unicode/text/UCA/CEList.java
markusicu: changed to WeightStyle -- ok?

unicodetools/src/main/java/org/unicode/text/UCA/FractionalUCA.java
markusicu: changed to WeightStyle -- ok?

unicodetools/src/main/java/org/unicode/text/UCA/CEList.java
echeran: yep, sgtm

echeran (dismissed):
LGTM

echeran (approved):

PR #1158, opened 2025-05-30
Index-17.0.0d2.txt

markusicu:
From Ken:

I've done the obvious updates for Index.txt for 17.0, adding the new scripts and other blocks. I also added a mere handful of names of individual characters. I settled on SAUDI RIYAL SIGN, TROMBONE, and ORCA as the most lookup-worthy additions. And I caught up a bit on the indexing of the CJK extensions.


nedley (approved):

nedley:
Sorry, I noticed something after approving: there should also be `Miscellaneous Symbols	1CEC0` matching the one that was added with `Symbols` coming first.

PR #1160, opened 2025-06-14
Updated Unikemet regular expressions and regenerated UCDXML

jowilco:
Fix for https://github.com/unicode-org/properties/issues/435

Updated Unikemet regular expressions; however, RegEx for kEH_FVal doesn't validate against current data.

There are two issues:

1. The separator is ambiguous: "The delimiters '/' or '|' are used to separate alternative values..."
2. The syntax specifies letters and combining characters (e.g., h + COMBINING DOT BELOW [U+0323], rather than ḥ [U+1E25]). But Unikemet.txt in UCD uses the composed forms.

To avoid this for now, we'll treat kEH_FVal as SINGLE_VALUED.

I updated the rest of the Unikemet regular expressions from the latest version of UAX 57.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/props/UcdProperty.java
+++ b/unicodetools/src/main/java/org/unicode/props/UcdProperty.java
@@ -217,12 +217,7 @@ public enum UcdProperty {
     kEACC(PropertyType.Miscellaneous, DerivedPropertyStatus.Provisional, "cjkEACC"),
     kEH_Cat(PropertyType.Miscellaneous, DerivedPropertyStatus.Approved, "kEH_Cat"),
     kEH_Desc(PropertyType.Miscellaneous, DerivedPropertyStatus.Approved, "kEH_Desc"),
-    kEH_FVal(
-            PropertyType.Miscellaneous,
-            DerivedPropertyStatus.Provisional,
-            null,
-            ValueCardinality.Unordered,
-            "kEH_FVal"),
+    kEH_FVal(PropertyType.Miscellaneous, DerivedPropertyStatus.Provisional, "kEH_FVal"),
     kEH_Func(
             PropertyType.Miscellaneous,
             DerivedPropertyStatus.Provisional,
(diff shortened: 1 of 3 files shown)


markusicu (approved):
changes lgtm, assuming CI is happy now -- please **squash**-and-merge when ready

markusicu:
> The separator is ambiguous: "The delimiters '/' or '|' are used to separate alternative values..."

I haven't looked at it. If it's ambiguous or otherwise problematic, do you want to report it? Via a PAG issue?

jowilco:
> > The separator is ambiguous: "The delimiters '/' or '|' are used to separate alternative values..."
> 
> I haven't looked at it. If it's ambiguous or otherwise problematic, do you want to report it? Via a PAG issue?

Add for the next PAG discussion? I was more worried about the composed vs decomposed characters for the syntax.

markusicu:
> Add for the next PAG discussion?

yes, asap given the timing and holidays/vacations

> I was more worried about the composed vs decomposed characters for the syntax.

I didn't quite understand the problem. Usually a composite of letter+mark is also a letter.
Can you give an example from spec & data where you think there is a problem?

jowilco:
Hi Markus:
Compare the syntax for:

- https://www.unicode.org/reports/tr57/#kEH_FVal
- https://www.unicode.org/reports/tr57/proposed.html#kEH_FVal

Then compare those to `U+13002	kEH_FVal	ḥmsꞽ` from [Unikemet.txt]https://www.unicode.org/Public/UCD/latest/ucd/Unikemet.txt).

The ḥ (U+1E25) from the first char of kEH_FVal for cp 13002 is specified in https://www.unicode.org/reports/tr57/#kEH_FVal, and would be covered by h combined with \x{323} from https://www.unicode.org/reports/tr57/proposed.html#kEH_FVal
However, I've tried the syntax with two different regular expression tools and neither of them match ḥ on `[h\x{323}]+`.

So, the problem is that the syntax makes sense for humans, but not for (at least) some parsers.

jowilco:
Do we make that caveat in any other UAX?

jowilco:
Ok - I concur. So, we're back to just the original problem - the delimiter.
There's a secondary issue in that all other multi-valued properties appear to use space as the delimiter. In the original issue, IndexUnicodeProperties sees `? | mnḫ.t` as 5 separate groups: `?`, the space before the |, the |, the space after the |, and `mnḫ.t`.
I'll spend some more time on this to see if I can figure out a sustainable solution. But let's close this issue.

PR #1161, opened 2025-06-16
Handle the ambiguous delimiter in kEH_FVal

jowilco:


unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
```diff
@@ -538,6 +538,9 @@ public void setMultiValued(String multivalued2) {
         if (property == UcdProperty.Name_Alias || property == UcdProperty.Standardized_Variant) {
             multivaluedSplit = NO_SPLIT;
         }
+        if (property == UcdProperty.kEH_FVal) {
+            multivaluedSplit = Pattern.compile(" [\\|/] ");
```
jowilco, line 542: Note that I wasn't sure whether just to compile the pattern here as this is (hopefully) a 1-off, or whether to add it to the group of patterns (like COMMA) at the top of this file.

jowilco: 1. I don't - fixed.
2. Done.
3. kEH_Func, although marked as MULTI_VALUED, is picking up each entire string via `[^\t"]+`. I don't know if this is Michel's long-term plan for the syntax. So, no - I'm not doing anything for delimiters for kEH_Func.
4. I'm now handling both / and | with the optional leading and trailing spaces.

https://github.com/unicode-org/properties/issues/437

markusicu: > 3. kEH_Func, although marked as MULTI_VALUED, is picking up each entire string via `[^\t"]+`.

That's after splitting:
https://www.unicode.org/reports/tr57/proposed.html
> Validation is done as follows: The entry is split into subentries using the Delimiter (if defined), and each subentry converted to Normalization Form D (NFD). The value is valid if and only if each normalized subentry matches the property’s Syntax regular expression. Note that any given property’s Syntax is not guaranteed to be stable and may change in the future.

Follow-up PR?

jowilco: > Follow-up PR?

https://github.com/unicode-org/unicodetools/pull/1162

PR #1168, opened 2025-07-20
More Unicode 17.0 Confusables

josh-hadley:
This PR rolls up several individual confusables items:
 - Arabic confusables for Hindko (PAG ref https://github.com/unicode-org/properties/issues/412)
 - New confusables data based on L2/22-108 (PAG ref https://github.com/unicode-org/properties/issues/415)
 - An analysis of confusable sequences proposed in L2/22-107 (PAG ref https://github.com/unicode-org/properties/issues/416)
 - New confusable data based on L2/22-114 (PAG ref https://github.com/unicode-org/properties/issues/417)
 - An analysis of confusable sequences mentioned in L2/22-081 (PAG ref https://github.com/unicode-org/properties/issues/418)
 - Missing Coptic confusables (PAG ref https://github.com/unicode-org/properties/issues/419)
 - Confusable Katakana-Han pair (PAG ref https://github.com/unicode-org/properties/issues/442)

Each item has 2 commits: one for the `confusables-source.txt` and a second one containing the generated data from the updated source.


roozbehp:
Thanks so much! 

Something in this PR is causing U+004F O and U+006F o to become confusable with each other, which is not intended or desired. I will investigate, but just a heads-up that it is not OK to merge this as-is.

macchiati:
The way the code works, if you introduce an intermediary that matches both of those (eg half-way between), it will form a transitive closure.

roozbehp:
> Something in this PR is causing [U+004F](https://util.unicode.org/UnicodeJsps/character.jsp?a=004F) O and [U+006F](https://util.unicode.org/UnicodeJsps/character.jsp?a=006F) o to become confusable with each other, which is not intended or desired. I will investigate, but just a heads-up that it is not OK to merge this as-is.

I think the source of the problem is adding these three lines without removing their previous confusables: 
```
09E6 ; 006F
0B66 ; 006F
0CE6 ; 004F
```

@josh-hadley, please comment these three lines out for now and regenerate. We'll figure out how to add them properly later.

josh-hadley:
> comment these three lines out for now and regenerate

@roozbehp looks like that did the trick (for eliminating the test failure at least). Please continue reviewing with the updated data.

roozbehp (dismissed):
LGTM. Will try to figure out how to properly fix the commented-out lines in the next round.

roozbehp (approved):

josh-hadley:
@markusicu do you want to give this a quick 👀 before merge (squash-and-merge)?

PR #1183, opened 2025-08-01
Reflect use of Macron Below instead of Minus Below in Syriac in scx

roozbehp:
[UTC-184-A57] Action Item for Roozbeh Pournader, SAH: Update ScriptExtensions.txt to reflect the U+0331 COMBINING MACRON BELOW usage in Syriac, for Unicode Version 17.0. [Ref. 4.3 in L2/25-187]


markusicu:
Minor: Please either paste the action item text as rich text, with the explicit links, or else prefix the action item code with "UTC-" for auto-linking (which I just added here). I like to look at them in situ / in context.

PR #1190, opened 2025-08-05
UAE Dirham sign

jowilco:
See [unicode-org/sah#650](https://github.com/unicode-org/sah/issues/650) and [L2/25-159](https://www.unicode.org/cgi-bin/GetDocumentLink?L2/25-159),

[[184-C17](https://www.unicode.org/cgi-bin/GetL2Ref.pl?184-C17)] Consensus: UTC accepts U+20C3 UAE DIRHAM SIGN for encoding in the Currency Symbols block based on [L2/25-159](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/25-159), for Unicode Version 18.0. [Ref. 2.1 in [L2/25-187](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/25-187)]

[unicode-org/utc-release-management/issues/232](https://github.com/unicode-org/utc-release-management/issues/232)


unicodetools/data/ucd/dev/LineBreak.txt
```diff
@@ -971,7 +971,9 @@
 20BF           ; PR # Sc         BITCOIN SIGN
 20C0           ; PO # Sc         SOM SIGN
 20C1           ; PR # Sc         SAUDI RIYAL SIGN
-20C2..20CF     ; PR # Cn    [14] <reserved-20C2>..<reserved-20CF>
+20C2           ; PR # Cn         <reserved-20C2>
+20C3           ; AL # Sc         UAE DIRHAM SIGN
```
markusicu, line 975: Probably not lb=AL.
L2/25-159 proposes PO but the Saudi Riyal has PR.
@eggrobin @Ken-Whistler PO or PR?
John see
- https://www.unicode.org/reports/tr14/#PO
- https://www.unicode.org/reports/tr14/#PR
- [`[:Uβ:gc=Sc:]` grouped by lb](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AU%CE%B2%3Agc%3DSc%3A%5D&g=&i=lb)

unicodetools/data/ucd/dev/Index.txt
```diff
@@ -1603,6 +1603,7 @@ direct sum	2295
 Directional Format Characters	202A
 DIRECTIONAL FORMATTING, POP	202C
 DIRECTIONAL ISOLATE, POP	2069
+DIRHAM SIGN, UAE	20C3
```
markusicu, line 1606: @Ken-Whistler usually adds things into the Index late in a release. Ken, do you want this here in the pending-for-18 pull request?

unicodetools/data/ucd/dev/NamesList.txt
```diff
@@ -13711,6 +13711,8 @@
 	* Kyrgyzstan
 20C1	SAUDI RIYAL SIGN
 	* Saudi Arabia
+20C3	UAE DIRHAM SIGN
+	* United Arab Emirates
```
markusicu, line 13715: Ken _generates_ the NamesList from other sources, and likes to add change comments at the top. I assume that he would prefer not to add this in this kind of PR.

unicodetools/data/ucd/dev/DerivedNormalizationProps.txt
markusicu, line 1: please revert files that only have time stamp changes

markusicu (commented):
welcome to the property encoding club :-)

unicodetools/data/ucd/dev/LineBreak.txt
eggrobin: Hmm, I think I had written something about PO vs. PR in discussion of the Riyal, let me see if I can find that…

eggrobin: Ah yes, in https://github.com/unicode-org/sah/issues/619, published in L2/25-087:
> In particular, Line_Break=PR, not PO. Note that the difference between PR and PO is nowadays only relevant in an East Asian context; see Section 3 of document [L2/05-292](https://www.unicode.org/cgi-bin/GetDocumentLink?L2/05-292), adopted by decision [105-C37](http://www.unicode.org/cgi-bin/GetL2Ref.pl?105-C37), has useful background here. The point of picking lb=PR or lb=PO is to allow a break, as in お白湯は`÷`¥500頂戴しております。  or 上記税込価格にサービス料として10％`÷`を頂戴いたします. Since the currency sign is meant to be prefix in LTR text, it would likely be prefix in CJK text as currency signs usually are (with exceptions such as ¢), so that we would expect お白湯は`÷`<span></span>`SAR sign`13頂戴しております。 and the character should be lb=PR like ¥.

So the question is whether it would be prefix in a CJK context.

unicodetools/data/ucd/dev/NamesList.txt
eggrobin: Yes, NamesList should not be touched by these PRs.

Ken-Whistler: I agree with Robin. Don't touch NamesList.txt directly. (And don't deal with Index.txt, either -- updating that is outside the context of establishing the properties.)

unicodetools/data/ucd/dev/DerivedAge.txt
```diff
@@ -2116,4 +2116,14 @@ FDC8..FDCE    ; 17.0 #   [7] ARABIC LIGATURE RAHIMAHU ALLAAH TAAALAA..ARABIC LIG
 
 # Total code points: 4803
 
+# ================================================
+
+# Age=V18_0
+
+# Newly assigned in Unicode 18.0.0 (September, 2026)
+
+20C3          ; 18.0 #       UAE DIRHAM SIGN
```
eggrobin, line 2125: DerivedAge (like the Derived files generally) is never hand-edited, only regenerated as described in https://github.com/unicode-org/unicodetools/blob/main/docs/pipeline.md#regenerate-ucd.

This means that it will falsely say that this file is targeted for 17.0 (because we don’t have an 18.0 yet). That is OK, it will get regenerated to 18 once we have an 18 and the merge/regenerate commands get run on that PR.

The CI is supposed to put up a warning saying that this does not actually target 17 (otherwise some people in the public get false hopes looking at the diffs), but it does not know about 18, only about provisional assignment. I will fix the CI.

PR #1191, opened 2025-08-05
NamesList-17.0.0d2.html

markusicu:
from Ken, just updating version 16 to 17

### DIFF
--- a/unicodetools/data/ucd/dev/NamesList.html
+++ b/unicodetools/data/ucd/dev/NamesList.html
@@ -100,27 +100,27 @@ a.headernav:hover {
     <tbody>
       <tr>
         <td>Revision</td>
-        <td>16.0.0</td>
+        <td>17.0.0</td>
       </tr>
       <tr>
-        <td>Authors</td>
+        <td>Editors</td>
         <td>Asmus Freytag, Ken Whistler</td>
       </tr>
       <tr>
         <td>Date</td>
-        <td>2024-08-21</td>
+        <td>2025-08-05</td>
       </tr>
       <tr>
         <td>This Version</td>
         <td >
-		<a href="https://www.unicode.org/Public/16.0.0/ucd/NamesList.html">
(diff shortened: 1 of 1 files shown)


Ken-Whistler (approved):
As expected.

markusicu:
@josh-hadley @echeran @eggrobin is the [build failure](https://github.com/unicode-org/unicodetools/actions/runs/16763582440/job/47464868781?pr=1191) real? Is something grabbing the version number from NamesList.html and fails if something else is not yet on the same new version??

eggrobin:
I certainly hope nothing looks at NamesList.html, and it is far too late for me to try to understand the build failure. But not too late for me to rerun the build and go to bed…

josh-hadley:
> @josh-hadley @echeran @eggrobin is the [build failure](https://github.com/unicode-org/unicodetools/actions/runs/16763582440/job/47464868781?pr=1191) real? Is something grabbing the version number from NamesList.html and fails if something else is not yet on the same new version??

I'm not exactly sure what is reading what but the failure is in a command that comes after this: https://github.com/unicode-org/unicodetools/blob/560c2906d5e0577423cc0aba9c1bf322c0a18dfa/UnicodeJsps/Dockerfile#L7

Do we need to set that to 17.0.0? (probably)

eggrobin:
> Do we need to set that to 17.0.0? (probably)

Probably not actually, because it is used to get something from https://www.unicode.org/Public/PROGRAMS/BidiReferenceC/, which has no 17 (it is updated _after_ the release, BRS steps R80 sqq.).

eggrobin:
I reran the build, it worked.

I noticed last night that various bits of the unicode.org website were extremely flaky (while looking at reports I got them without the CSS, the index of https://www.unicode.org/Public/ would sometimes fail to load and sometimes fail to show the icons for the parent directory etc.).

Ken-Whistler:
Yeah, something was redlining the server, I suspect. I had a lot of trouble loading things, too.

PR #1194, opened 2025-08-07
Moved MathClass source and generated files from unicode-reports

jowilco:


unicodetools/data/math/MathClass.txt
```diff
@@ -0,0 +1,61 @@
+# File: MathClass.txt
+# Revision: 17.0
```
eggrobin, line 2: Move these files down into unicodetools/data/math/revision-17/*.txt.

At some point I can then add the older versions, parse them, and then we will be able to look at MathClass history using our usual tooling.

unicodetools/data/math/revision-16/MathClass.txt
```diff
@@ -0,0 +1,61 @@
...
1EEF1;L
1F784;N
1F78C;N
1F78D;N
1F797;N
1F798;N
1F799;N
1F79D;N
1F79E;N
1F79F;N
```
eggrobin, line 59: I think this file is missing line terminators.

jowilco:
>Windows git clients change the line endings
They do. The issue was it was CR, not LF. I could have just changed the endings to LF, but I went with CRLF (because, Windows...) and let git do the rest.

PR #1196, opened 2025-08-11
Confusables data for Devanagari UE and UUE

josh-hadley:
Devanagari UE and UUE (PAG ref https://github.com/unicode-org/properties/issues/449)


roozbehp:
The changes look good to me, but I'm wondering if it's too late for 17.0 data changes and we should keep these for 18.0, or it's fine landing it now. @markusicu? @PeterConstable?

josh-hadley:
Sorry I meant to have this be a draft PR, not for 17.0

PR #1205, opened 2025-09-12
Make GenerateConfusables generate NFKD instead of NFKC

roozbehp:
Also fix a typo in the name of the method

This fixes problems described in the following issue.
Rephrased from simple "This fixes (issue)" to prevent GitHub from auto-closing the issue.

- https://github.com/unicode-org/properties/issues/459

### DIFF
--- a/unicodetools/src/main/java/org/unicode/text/UCD/GenerateConfusables.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/GenerateConfusables.java
@@ -2639,6 +2639,7 @@ public class GenerateConfusables {
 
     private static final class ModifiedNFKD {
         private static Normalizer INSTANCE;
+
         static String normalize(String cf) {
             if (INSTANCE == null) {
                 INSTANCE = new Normalizer(UCD_Types.NFKD, Default.ucdVersion());
--- a/unicodetools/src/main/java/org/unicode/text/UCD/IdentifierInfo.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/IdentifierInfo.java
@@ -140,6 +140,7 @@ public class IdentifierInfo {
 
     private static final class ModifiedNFKC {
         private static Normalizer INSTANCE;
+
         static String normalize(String cf) {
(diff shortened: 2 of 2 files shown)


markusicu:
@roozbehp FYI -- I rephrased the PR description from simple "This fixes (issue)" to prevent GitHub from auto-closing the issue.

unicodetools/src/main/java/org/unicode/text/UCD/GenerateConfusables.java
```diff
@@ -2637,9 +2637,19 @@ private static String getReasonFromFilename(String type) {
...
 
+    private static Normalizer modNFKD;
+
+    static String getModifiedNFKD(String cf) {
+        if (modNFKD == null) {
+            modNFKD = new Normalizer(UCD_Types.NFKD, Default.ucdVersion());
+            modNFKD.setSpacingSubstitute();
+        }
+        return modNFKD.normalize(cf);
+    }
```
markusicu, line 2648: optional: A better way to do this in Java is to create a private inner class like
```
    private static final class ModifiedNFKD {
        private static Normalizer INSTANCE;
        private ModifiedNFKD() {
            // INSTANCE = new Normalizer(...) ...
        }

        static String normalize(String s) {
            return ModifiedNFKD.INSTANCE.normalize(s);
        }
    }
```
and then you can use ModifiedNFKD.normalize(s) elsewhere.

Java does thread-safe lazy initialization of an inner class.

(same for the existing code below)

markusicu (commented):
I can kind-of-rubber-stamp but have an optional code improvement.

unicodetools/src/main/java/org/unicode/text/UCD/GenerateConfusables.java
roozbehp: Thank you very much for the review. I couldn't use your snippet verbatim since I was getting some null references and couldn't figure out exactly why. I ended up checking for null in normalize().

roozbehp:
PTAL.

PR #1208, opened 2025-09-17
Update doc for setup of Github tokens to use Github Maven artifacts

echeran:
Updates to the CLDR docs on Maven will come in a subsequent PR in CLDR

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -10,20 +10,19 @@ can be used to:
 *   build 4 chart folders on the unicode site.
 *   build files for ICU (collation, NFSkippable)
 
-**WARNING!!**
-
-*   This is NOT production level code, and should never be used in programs.
-*   The API is subject to change without notice, and will not be maintained.
-*   The source is uncommented, and has many warts; since it is not production
-    code, it has not been worth the time to clean it up.
-*   It will probably need some adjustments on Unix or Windows, such as changing
-    the file separator.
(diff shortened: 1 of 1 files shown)


docs/build.md
```diff
@@ -86,6 +85,31 @@ Currently, some tests run on the generated output files of a tool (ex: in order
...
+```
+
+This happens because [Github's Maven artifact registry requires authentication](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-apache-maven-registry), 
+via a Github token, even when you are only reading artifacts.
+
+> [!IMPORTANT]
+> If you run into this issue, then:
+> 
+> 1. revisit the [CLDR Maven setup link](http://cldr.unicode.org/development/maven)
+to ensure that you have created the correct type of Github access token and that you have granted the correct permissions to it
```
eggrobin, line 109:
```suggestion
to ensure that you have created the correct type of Github access token: classic, _not_ fine-grained—with a fine-grained token, you will get a 403 rather than a 401—; and that you have granted the correct permissions to it.
```

echeran:
FYI, I created https://github.com/unicode-org/cldr/pull/5056 in conjunction with this PR since we all use the CLDR Maven setup page as the point of reference for these instructions, and it's good to improve & add on to that centralized place for the extra details on Github token types, Github's way of offering Maven artifacts, etc.

PR #1213, opened 2025-09-30
Parse MathClass and MathClassEx

eggrobin:


unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
```diff
@@ -674,6 +686,20 @@ static void parseSourceFile(
...
                                         : nextProperties.getProperty(propInfo.property));
                     } else {
+                        if (fileName.equals("math/*/MathClassEx")
+                                && indexUnicodeProperties.ucdVersion.compareTo(
+                                                VersionInfo.UNICODE_6_3)
+                                        <= 0) {
+                            parser =
+                                    parser.withLinePreprocessor(
+                                            s ->
+                                                    s.startsWith("FE61-FE68;")
```
jowilco, line 696: Same here. Ideally people shouldn't be using two different syntaxes for ranges, but is it worth matching a pattern rather than a string?

unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
```diff
@@ -1510,6 +1536,38 @@ private static void parseFields(
...
+                                < 0) {
+                    merger = new PropertyUtilities.NullIgnorer();
+                }
+                if (propInfo.property == UcdProperty.Math_Class_Ex
+                        && indexUnicodeProperties.ucdVersion.compareTo(Utility.UTR25_REVISION_16)
+                                < 0) {
+                    merger = new PropertyUtilities.RedundancyIgnorer();
+                }
+                if (propInfo.property == UcdProperty.Math_Class_Ex
+                        && indexUnicodeProperties.ucdVersion.compareTo(VersionInfo.UNICODE_6_0) < 0
```
jowilco, line 1556: Is it worth adding a comment explaining why you are doing this test?
Also, if you are proposing that we *should* support multiple values (e.g., exclamation mark), will we need a broader support for this type of construction?

unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
```diff
@@ -1646,6 +1704,27 @@ private static void parseSimpleFieldFile(
                         }
                     }
                     continue;
+                } else if (propInfo.property == UcdProperty.Math_Class
+                        && version.compareTo(VersionInfo.UNICODE_6_0) < 0) {
+                    // MathClass-11 had conflicting assignments for these two characters.  Instead
```
jowilco, line 1709: And here is the explanation!

unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
eggrobin: If we actually want multivalued MathClass (which I am starting to think we might), this will all get a lot simpler (because it will be handled like the other multivalued Math properties, with a MultivaluedJoiner).

I will copy the comment from Math_Class for now.

jowilco (dismissed):

eggrobin:
(This one needs a new rubberstamp…)

jowilco (approved):

PR #1227, opened 2025-10-25
Add some high-priority confusables

roozbehp:
See https://github.com/unicode-org/properties/issues/458

Note that U+031B vs U+02BC and U+0674 vs U+0654 are not added yet. They seem to need changes to the Java code.


roozbehp:
Please take a look

roozbehp:
> A few of them seem to have sufficient differences in size that the wouldn't be confusinle
> 
> @@ -70,6 +76,7 @@ A6F0 ; 0302 ; MA # ( ꛰ → ̂ ) BAMUM COMBINING MARK KOQNDON → COMBINING CIR
> 
> I would usually test with a few fonts... But that may be hard with Bamum

That's existing data. I haven't changed that here in this pull request.

PR #1233, opened 2025-11-11
UAX60 promotions

eggrobin:
From non-property UCD data to normative properties for the sources and provisional properties for the others.


markusicu (approved):
looks... plausible. the extraNames logic is for parsing old files?

I see that we have a CI failure from org.unicode.text.UCD.TestTestUnicodeInvariants.testSecurityInvariants but assume that we are ignoring those (and UCA) for a while.

eggrobin:
> the extraNames logic is for parsing old files?

Yes. Internally the tools need property alias stability to extend back to their provisional origins, but we don’t want to publish that history in PropertyAliases.txt, so these are « extra », tools-only.

PR #1234, opened 2025-11-11
Refine the types of some provisional properties

eggrobin:
Most obviously the numeric properties are numeric; some of the variants can be strings too.


jowilco (commented):
@eggrobin  Changes mostly make sense, but why the change for kSpoofingVariant from PropertyType.Miscellaneous to PropertyType.String?

eggrobin:
The practical effect is that, for example, on https://util.unicode.org/UnicodeJsps/character.jsp?a=4039, we will show the character 鿃 as well as `U+9FC3` for kSpoofingVariant, just like we already show `𥅴 <U+25174>` for kSimplifiedVariant. We still won’t show 睫 for `U+776B<kMatthews`, because we are not parsing these fancier properties (we do not have fancier types for those). But since we have a type for code point-valued properties, we might as well use it.

Issue #860, opened 2024-06-07
MakeUnicodeFiles should regenerate ArabicShaping.txt


eggrobin:
With our process that involves ~50 PRs in flight (and counting, this will only start going down after 16.0 is out), any impediment to automatic merge resolution is a significant pain. For most files, merges are resolved by ignoring the conflict markers and regenerating the files. ArabicShaping is manually maintained, and has bitten me recently while trying to synchronize all my PRs.

If we treat the schematic name as a pseudoproperty (we treat the joining type as a pseudoproperty Other_Joining_Type anyway), this should be tractable.

macchiati:
Makes sense

Issue #454, opened 2023-04-30
Add test for ArabicShaping schematic name


roozbehp:
The schematic name for Arabic characters in ArabicShaping.txt that have a joining group must match a specific pattern. In an early version of pull request https://github.com/unicode-org/unicodetools/pull/435, schematic names that didn't match that pattern were being added, which we caught on time.

An earlier version of the pattern was specified in L2/11-092, but it needs to be expanded slightly to cover some new additions.

Unicode tools should have a test to check the pattern.

Assigning to myself for now to provide the updated pattern.

CC @eggrobin @LornaSIL

PR #1237, opened 2025-11-12
Parse the ArabicShaping.txt schematic name

eggrobin:
A first step towards both #454 and #860.


unicodetools/src/main/java/org/unicode/props/UcdPropertyValues.java
```diff
@@ -95,6 +95,7 @@ public static Age_Values forName(String name) {
         }
     }
 
+    // Arabic_Shaping_Schematic_Name
```
roozbehp: What is this comment supposed to do?

eggrobin: This is a generated file; it has a comment for each property that is not enumerated.

roozbehp (approved):
LGTM, as far as I understand it.

PR #1238, opened 2025-11-13
Remove or correct non-NFD confusables

roozbehp:
See https://github.com/unicode-org/properties/issues/486


roozbehp:
Converting to draft to fix Java style issues.

PR #1247, opened 2025-11-26
Look at the reverse Do_Not_Emit mappings

eggrobin:
This would have caught the error fixed in #1246.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
+++ b/unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
@@ -708,11 +708,15 @@ public class PropertyParsingInfo implements Comparable<PropertyParsingInfo> {
                             propInfoSet);
                     break;
                 case Field:
-                    final var mapping = propInfo.getFieldMapping(indexUnicodeProperties.ucdVersion);
+                    FieldMapping mapping;
                     if (propInfoSet.size() == 1
                             && (propInfo = propInfoSet.iterator().next()).special
                                     == SpecialProperty.None
-                            && mapping.keyField == 0
+                            && (mapping =
+                                                    propInfo.getFieldMapping(
+                                                            indexUnicodeProperties.ucdVersion))
+                                            .keyField
+                                    == 0
                             && mapping.valueField == 1) {
                         if (fileName.equals("math/*/MathClass")
                                 && indexUnicodeProperties.ucdVersion.compareTo(


unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
```diff
@@ -48,12 +49,34 @@ enum SpecialProperty {
...
+        public int compareTo(FieldMapping other) {
+            return Comparator.<FieldMapping>comparingInt(m -> m.keyField)
+                    .thenComparing(m -> m.valueField)
+                    .compare(this, other);
+        }
+
+        final int keyField;
+        final int valueField;
+    }
+    ;
```
markusicu: ?

unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
```diff
@@ -48,12 +49,34 @@ enum SpecialProperty {
...
+
+        FieldMapping(int keyField, int valueField) {
+            this.keyField = keyField;
+            this.valueField = valueField;
+        }
+
+        @Override
+        public int compareTo(FieldMapping other) {
+            return Comparator.<FieldMapping>comparingInt(m -> m.keyField)
+                    .thenComparing(m -> m.valueField)
```
markusicu: I would turn the resulting Comparator into a static final field, so that it need not be built each time you call compareTo().

unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
```diff
@@ -48,12 +49,34 @@ enum SpecialProperty {
     public final UcdProperty property;
     public final SpecialProperty special;
 
+    public static class FieldMapping implements Comparable<FieldMapping> {
```
markusicu: please add a brief `/** class API doc */` describing the purpose/use of this new class

unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
```diff
@@ -185,7 +218,7 @@ public String toString() {
                 + " ;\t"
                 + property
                 + " ;\t"
-                + fieldNumbers
+                + fieldMappings
```
markusicu: don't you want a custom FieldMapping.toString() for this to be readable?

unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
```diff
@@ -665,7 +699,11 @@ static void parseSourceFile(
                     if (propInfoSet.size() == 1
                             && (propInfo = propInfoSet.iterator().next()).special
                                     == SpecialProperty.None
-                            && propInfo.getFieldNumber(indexUnicodeProperties.ucdVersion) == 1) {
+                            && propInfo.getFieldMapping(indexUnicodeProperties.ucdVersion).keyField
+                                    == 0
+                            && propInfo.getFieldMapping(indexUnicodeProperties.ucdVersion)
+                                            .valueField
```
markusicu: optional: looking up the same mapping twice... maybe add an .equals(k, v)?

unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
eggrobin: Spot the C++ programmer (and then spotless kicked the semicolon down).

markusicu (approved):

PR #1250, opened 2025-12-02
Drop ICUProperty

eggrobin:
Its _raison d’être_ was showing some properties in the JSPs. With the fixing of #502 and the subsequent #1018, the JSPs no longer use ICUProperty. The handful of usages in unicodetools (as opposed to UnicodeJsps) are a bad idea (see #1074).

Also get rid of UnicodePropertyX which was never implemented nor used, whatever it was meant to be.


markusicu:
@macchiati FYI

PR #1198, opened 2025-08-13
Emoji 17.0 charts

nedley:
Update pom.xml to point at the newest available ICU and CLDR snapshots and fix the one place where the emoji tools were using removed CLDRFile functionality. Also make a first pass at cleaning up proposal data to avoid associating emoji with the original skin tone or gender proposals just because they had skin tone or gender.

PR #1251, opened 2025-12-03
Make GenerateEmojiData work as intended on Windows, and run it in CI

eggrobin:
Some of the files it generates are part of the UCD, so we should treat them as such and ensure they are fresh.

It turns out the tool was never meant to spit files with random names all over the place, it’s just that
> [Many aspects of the behavior of [File.renameTo] are inherently platform-dependent: The rename operation might not be able to move a file from one filesystem to another, it might not be atomic, and it might not succeed if a file with the destination abstract pathname already exists.](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#renameTo-java.io.File-)

Update the documentation accordingly (it had been written by Windows users).

Run the tool, bumping version numbers and picking up the skin tone proposal mapping changes from #1198 in the internal file.

### DIFF
--- a/unicodetools/data/emoji/dev/emoji-sequences.txt
+++ b/unicodetools/data/emoji/dev/emoji-sequences.txt
@@ -1,11 +1,11 @@
 # emoji-sequences.txt
-# Date: 2025-07-25, 17:54:32 GMT
+# Date: 2025-12-03, 13:29:04 GMT
 # © 2025 Unicode®, Inc.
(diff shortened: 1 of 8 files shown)


jowilco (approved):

eggrobin:
Yes, that is the TempPrintWriter alluded to above, which GenerateEmojiData uses. It did not work right on Windows, but this fixes it.

It might make sense to make MakeUnicodeFiles use it.

PR #1326, opened 2026-04-08
Pseudoproperties to distinguish the types of name aliases

eggrobin:
This would be useful for the indexer (searching for 0000 should probably show NULL rather than NUL).

### DIFF
--- a/unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
+++ b/unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
@@ -1609,7 +1609,12 @@ public class PropertyParsingInfo implements Comparable<PropertyParsingInfo> {
                                 .name()
                                 .substring(11)
                                 .toLowerCase(Locale.ROOT)
-                                .equals(parts[2])) {
+                                .equals(
+                                        indexUnicodeProperties.ucdVersion.compareTo(
+                                                                VersionInfo.UNICODE_6_0)
+                                                        <= 0
+                                                ? "correction"
+                                                : parts[2])) {
                     continue;
                 }
                 String value =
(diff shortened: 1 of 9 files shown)


markusicu:
> (searching for 0000 should probably show NULL rather than NUL).

FYI: I generally write NULL for pointers and NUL for the control code... :-}

markusicu:
Strange error in “In-source Instructions (Examples)”

unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
```diff
@@ -1602,6 +1604,19 @@ private static void parseFields(
...
+                                .substring(11)
+                                .toLowerCase(Locale.ROOT)
+                                .equals(
+                                        indexUnicodeProperties.ucdVersion.compareTo(
+                                                                VersionInfo.UNICODE_6_0)
+                                                        <= 0
+                                                ? "correction"
+                                                : parts[2])) {
+                    continue;
+                }
```
markusicu: would you mind pulling the inner string into a variable for readability?
```suggestion
                if (propInfo.property.name().startsWith("Name_Alias_")) {
                    String type = indexUnicodeProperties.ucdVersion.compareTo(VersionInfo.UNICODE_6_0) <= 0 ? "correction" : parts[2]));
                        if (!propInfo.property.name().substring(11).toLowerCase(Locale.ROOT).equals(type) {
                        continue;
                    }
                }
```
modulo spotless

eggrobin:
> > (searching for 0000 should probably show NULL rather than NUL).
> 
> FYI: I generally write NULL for pointers and NUL for the control code... :-}

Fair. But right now I think it returns the first of the aliases alphabetically, which is definitely weird:
https://eggrobin.github.io/unicode-annotations/charindex.html?q=0001 `SOH`
https://eggrobin.github.io/unicode-annotations/charindex.html?q=0002 `START OF TEXT`
https://eggrobin.github.io/unicode-annotations/charindex.html?q=000A `END OF LINE`
https://eggrobin.github.io/unicode-annotations/charindex.html?q=000C `FF`

markusicu (approved):

PR #1329, opened 2026-04-13
A script to pick up updates to the NamesList

eggrobin:
It now lives in the charts repository, so we can pick it up from there.

Also move egg.md where it belongs, and run the script.


docs/update_nameslist.ps1
```diff
@@ -0,0 +1,33 @@
...
+# .\docs\update_nameslist.ps1.
+
+$ErrorActionPreference = "Stop"
+
+pushd ..\charts
+git fetch
+$versions = git log --pretty=format:"%H" main -- .\nomenclator\output\NamesList.txt
+for ($i = 0; $i -lt $versions.Length; ++$i) {
+    git checkout $versions[$i] .\nomenclator\output\NamesList.txt
+    if (-not (Compare-Object                                     `
```
markusicu, line 14: This deserves a comment.
It looks like the loop goes from the latest charts NamesList to the oldest, and stops when it finds a version that matches what we have in the unicodetools...?

What if the unicodetools version is not an exact match for any charts version? For example, what if I change the time stamp on a file just so that all UCD files have current-year time stamps?

docs/update_nameslist.ps1
```diff
@@ -0,0 +1,33 @@
...
+    $author = git show -s --format='%an <%ae>' $version
+    $date = git show -s --format='%aD' $version
+    $message = git show -s --format='%B' $version
+    git checkout $versions[$i] .\nomenclator\output\NamesList.txt
+    cp .\nomenclator\output\NamesList.txt ..\unicodetools\unicodetools\data\ucd\dev\NamesList.txt
+    popd
+    git add .\unicodetools\data\ucd\dev\NamesList.txt
+    git commit -m $message
+    git commit --amend --author=$author --no-edit
+    git commit --amend --date=$date --no-edit
```
markusicu: Why not one `git commit` command with the relevant options?

docs/update_nameslist.ps1
```diff
@@ -0,0 +1,33 @@
+# Picks up updates to NamesList.txt from a neighbouring clone of the charts repository.
+# This should be run from the root of the unicodetools repository, as
+# .\docs\update_nameslist.ps1.
```
markusicu: Please add that this needs to be run in (I assume) a Windows Powershell.

eggrobin: Powershell certainly. I think that can be installed on other OSes. Not sure what happens to the path separators then.

docs/update_nameslist.ps1
eggrobin: > What if the unicodetools version is not an exact match for any charts version? For example, what if I change the time stamp on a file just so that all UCD files have current-year time stamps?

Then this script will not work. I think in that case we should probably make sure that we are not stomping over a significant change. Checking out unicodetools before the unicodetools-side change would work (and then they can be merged).

I will add an error message in that case so that it does not fail in a stupid way (I think right now it would reënact the entire history).

docs/update_nameslist.ps1
```diff
@@ -0,0 +1,42 @@
...
+        break
+    }
+}
+if ($i -eq $versions.Length) {
+    popd
+    Write-Error ("Could not find charts revision matching unicodetools "     `
+    + "NamesList.txt.  If NamesList.txt has been modified in the tools, "    `
+    + "check out the tools before that revision and run this script again, " `
+    + "then merge.")
+}
```
markusicu, line 27: this should terminate, right? or does Write-Error do so?

eggrobin: Write-Error does that (at least with $ErrorActionPreference = "Stop" from above).

eggrobin: Editing NamesList.txt, this is the result of running the tool:
```
C:\Users\robin\Projects\Unicode\unicodetools [nameslist-update-script ≡ +1 ~1 -0 !]> .\docs\update_nameslist.ps1
Updated 0 paths from b3fe5a7
Updated 1 path from 24b5b35
C:\Users\robin\Projects\Unicode\unicodetools\docs\update_nameslist.ps1 : Could not find charts
revision matching unicodetools NamesList.txt.  If NamesList.txt has been modified in the tools,
check out the tools before that revision and run this script again, then merge.
At line:1 char:1
+ .\docs\update_nameslist.ps1
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException,update_nameslist.ps1
```

markusicu (approved):

PR #1331, opened 2026-04-14
A more compact charindex

eggrobin:
Compare https://eggrobin.github.io/unicode-annotations/charindex.html (old) and https://eggrobin.github.io/unicode-annotations/charindex-smol.html (with this change).
(Note that I will probably replace charindex.html with the -smol one after merging this.)

Take all the highly repetitive strings (both the actual property values and the HTML snippets), stick them in a giant string, and deflate that, replacing the strings with indices in the giant string throughout the data structures: it goes from 22 MiB to 1388 kiB (6.3%). Also don’t try to pretty-print a map with 66666 entries.

This brings the generated charindex.html from 42.3 MiB to 8.91 MiB (21% of its size).
The page gets compressed by the server, and the compressed size doesn’t change much (4477 kB vs. 3175 kB, says Chrome), so this doesn’t change download times very much.

However, this massively reduces the time spent parsing JS. When the page is loaded from disk cache, the time to DomContentLoaded goes from 2.10 s to 636 ms.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/text/tools/Indexer.java
+++ b/unicodetools/src/main/java/org/unicode/text/tools/Indexer.java
@@ -87,6 +87,27 @@ public class Indexer {
 
     private static int maxRSEntryCharacters = 0;
 
+    private static class StringIndexer {
+        public StringIndexer() {}
+
+        public int getStringIndex(String s) {
+            int result = stringIndices.getOrDefault(s, allTheStrings.length());
+            if (result == allTheStrings.length()) {
+                allTheStrings.append(s).append(RECORD_SEPARATOR);
+                stringIndices.put(s, result);
+            }
+            return result;
+        }
+
+        @Override
+        public String toString() {
+            return allTheStrings.toString();
+        }
+
+        private final HashMap<String, Integer> stringIndices = new HashMap<>();
+        private final StringBuilder allTheStrings = new StringBuilder();
+    }
+
     static {
         String baseRules =
                 "'<' > '&lt;' ;"
@@ -263,8 +284,7 @@ public class Indexer {
                 return left.getName().compareTo(right.getName());
             }
         }
-        final StringBuilder allTheStrings = new StringBuilder();
-        final HashMap<String, Integer> stringIndices = new HashMap<>();
+        final var allTheStrings = new StringIndexer();
         // Property to snippet based on property value (as an index in allTheStrings) to index
         // entry.
         Map<UnicodeProperty, Map<Integer, IndexEntry>> indexEntries =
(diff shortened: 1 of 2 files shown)


unicodetools/src/main/java/org/unicode/text/tools/Indexer.java
```diff
@@ -292,10 +300,14 @@ public int compare(UnicodeProperty left, UnicodeProperty right) {
                     } else if (prop == NAME) {
                         snippet = snippet.replace(Utility.hex(cp), "#");
                     }
-                    // Copy the snippet to a final variable for use in the λ.
-                    final String indexSnippet = snippet;
+                    final int snippetIndex =
+                            stringIndices.getOrDefault(snippet, allTheStrings.length());
+                    if (snippetIndex == allTheStrings.length()) {
+                        allTheStrings.append(snippet).append(RECORD_SEPARATOR);
```
markusicu, line 306: Please create & use a helper function that takes a string (without the separator) and returns the index. Internally, figure out whether to reuse or append.

markusicu, line 306: optional suggestion/idea:
- try not terminating each string with RS
- for deduplication, instead of a Map from string to start index, try to use allTheStrings.indexOf(s)
    - much slower (linear search), but lets you share substrings and overlapping strings
    - additionally, if there is no substring match, you could overlap the new string with the trailing 1..length-1  allTheStrings `char`s -- and only append the remainder
- keep not just the index, but index+length
- in the builder code, you could use a struct
- in JS, you could store an int like `(index << 8 | length)`
- this of course requires keeping to a maximum string length; that could be hardcoded (with the builder failing when a string is too long), or be determined at build time, telling the JS code the shift width
- JS would not need to look for the RS
- JS would however get larger index+length numbers

eggrobin: > keep not just the index, but index+length

That would ~double the data size; these are big numbers, interchanged as text; note that at this point most of the size is the indices, not the compressed strings. Much worse than the cost of the RSes (negligible after compression).

Reusing suffixes could work, but it is complex, I am not sure we have that many of those, and it would probably not help much with the size of allTheStringsCompressed (deflate already reuses substrings, see, e.g., https://bsky.app/profile/exurd.neocities.org/post/3lzdykla3tc2s). The only thing it could help with would be by making the indices smaller.

markusicu: Ok.
Would you mind making the other change -- using a helper function for a repeated pattern?

eggrobin: Done. And by systematically checking for a pre-existing string (I was doing that for the property values but not for the HTML), brought the size down to 8.76 MiB (from 8.91 MiB) mentioned above.

markusicu (approved):

PR #1338, opened 2026-04-24
remove UVERSION env var

### LINKED ISSUE #376: default value for UVERSION
We set UVERSION for running tools and tests, but it looks like the only place that uses it is in GenerateUnihanCollators.java.

That code already checks if it gets `version == null`, and then defaults to Settings.latestVersion, which is usually what we want.

Check that this works (do we also need to test for `version.isEmpty()`??) and then remove the UVERSION parameter from the docs and from the GitHub CI workflows.

markusicu:
Change the only class that looks at the UVERSION environment variable -- GenerateUnihanCollators -- to just use Settings.latestVersion, and then remove UVERSION from build scripts and instructions.

Fixes https://github.com/unicode-org/unicodetools/issues/376

### DIFF
--- a/.github/workflows/build-jsp.yml
+++ b/.github/workflows/build-jsp.yml
@@ -1,8 +1,6 @@
 name: Build JSP
 
 env:
-  CURRENT_UVERSION: 18.0.0 # FIX_FOR_NEW_VERSION
-  PREVIOUS_UVERSION: 17.0.0  # not used at present
   REGISTRY: ghcr.io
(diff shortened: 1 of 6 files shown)


josh-hadley (dismissed):

markusicu:
@josh-hadley thanks for the quick approval! However, spotless was unhappy, so I pushed a small fix. PTAL

nedley (approved):

eggrobin (approved):

PR #1340, opened 2026-04-24
update spotless & apply

markusicu:
My machine updated to Java 25, and spotless throws exceptions.

I updated spotless to the same version that ICU is using for the last couple of months, and the underlying google-java-format to its latest version.

Then I ran spotless, and it changed a few files. Mostly (maybe only), comments before a switch-case are indented only as much as the `case CONSTANT`, not as much as the preceding case body code.

This broke CI initially, because it was using the same Java 11 for compiling Unicode Tools code as well as for building and running spotless. @mihnita fixed this with PR #1342.

@mihnita FYI

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,7 +47,9 @@
         <!-- Need to use an updated surefire plugin here. -->
         <maven-surefire-plugin-version>3.0.0-M5</maven-surefire-plugin-version>
 
-        <spotless.version>2.43.0</spotless.version>
+        <spotless.version>3.1.0</spotless.version>
+        <!-- https://mvnrepository.com/artifact/com.google.googlejavaformat/google-java-format -->
+        <google-java-style.version>1.35.0</google-java-style.version>
     </properties>
 
     <modules>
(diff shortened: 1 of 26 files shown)


pom.xml
```diff
@@ -134,7 +134,7 @@
                             <googleJavaFormat>
                                 <!-- version of google-java-style, see
                                     https://mvnrepository.com/artifact/com.google.googlejavaformat/google-java-format -->
-                                <version>1.22.0</version>
+                                <version>1.35.0</version>
```
mihnita: Might also be nice to move this version to properties.

markusicu: done

mihnita (approved):

markusicu:
Taking this down to draft while begging Mihai to help fix the Java version conflict.

mihnita:
If you have a way to re-trigger the tests this should pass now.

markusicu:
It looks like this works now. I would love for one of the reviewers to please stamp this :-)

eggrobin (approved):

Issue #312, opened 2022-08-17
make all data file generators write into the repo source tree / only if real changes


markusicu:
Change data file generators so that they write files that are version-controlled into the repo source tree, rather than into the "Generated" side folder, but only if there are changes other than timestamps and similar. There is relevant code already in use by the emoji and UCD file generators, and by CLDR.

Note that the security data file generators already write into the repo source tree, but do so even for no-op changes, which is not good.

Benefits:
- We can see relevant diffs directly via `git diff` and similar.
- We need not even look at file diffs if there are no real changes.
- We need not manually copy files that do have real changes.

We should probably split this into sub-tasks.
- [ ] UCD files (MakeUnicodeFiles)
- [ ] UCA files
- [ ] emoji
- [ ] idna
- [ ] security

eggrobin:
(I just ticked two boxes here.)

srl295:
check out CLDR's TempPrintWriter  with the copyright-exclusion option.

eggrobin:
We know about it, but we use our own DiffingPrintWriter (which ultimately diffs the result of getLineWithoutFluff).

PR #1352, opened 2026-04-29
IdnaMappingTable and IdnaTestV2 in MakeUnicodeFiles

eggrobin:
Drive-by bugfix to link data generation which was using ICU properties to figure out whether to print a literal character.

Fixes the idna part of #312.

Still need to do IDNA2008.txt, this will come in a later PR.


unicodetools/src/main/java/org/unicode/tools/GenerateLinkData.java
```diff
@@ -48,11 +49,15 @@
  */
 public class GenerateLinkData {
 
-    private static final Transliterator FIX_ODD =
-            Transliterator.createFromRules(
-                    "any-html",
-                    ":: [[:C:][:Z:][:whitespace:][:Default_Ignorable_Code_Point:]] hex/unicode ; ",
-                    Transliterator.FORWARD);
+    private static final Transliterator FIX_ODD = getFixOdd();
```
markusicu, line 52: FYI: You could probably use a lambda to avoid a named function that is not used anywhere else.

markusicu (dismissed):

unicodetools/src/main/java/org/unicode/tools/GenerateLinkData.java
eggrobin: I tried, but failed to get something that was legal Java.

markusicu (approved):

PR #1343, opened 2026-04-27
add maven.config: no-transfer-progress, UTF-8

markusicu:
Copy ICU's .mvn/maven.config file:

a) suppress "Progress" output lines like in
```
Downloading from central: (URL)
Progress (1): 415 kB
Progress (2): 415 kB | 7.7/790 kB
Progress (2): 415 kB | 16/790 kB 
Progress (2): 415 kB | 24/790 kB
Progress (2): 415 kB | 40/790 kB
```

b) make Maven invocations default to reading/writing files using UTF-8, which should avoid having to specify this in each Maven target

Suggested by @mihnita.

- [x] Approver: Feel free to merge on my behalf
    - [x] rebase & merge one or more commits
    - [ ] squash & merge multiple commits into one

PR #1355, opened 2026-04-29
Remove comments from maven.config

roozbehp:
Maven 3.8.x does not officially support comments in the maven.config file. When Maven attempts to parse the file, it treats every line—including those intended as comments—as a command-line argument, leading to "Unrecognized option" or "Unable to parse" errors.


eggrobin:
See my reply on https://github.com/unicode-org/unicodetools/pull/1343#issuecomment-4348251896.

Apache Maven works fine (what would nonstandard Apache Maven be anyway?), you just need a newer one.

markusicu (changes requested):
Please upgrade to Maven 3.9 or later.

See the [Unicode Tools dev meeting minutes from yesterday](https://docs.google.com/document/d/1hnkTMB62_PynKZ_D-GMBSnYDyf2Xbz0n0NC0RGdPKt4/edit?tab=t.0#heading=h.2sdbnciydf3o).

markusicu:
See https://maven.apache.org/configure.html#:~:text=are%20treated%20as%20comments

roozbehp:
Well, my build environment is in WSL. The lastest version of Ubuntu it supports is 24.04 LTS (Noble Numbat) and the latest version of maven that has is 3.8.7. I would really need to mess up my system in order to get maven 3.9.

I would really appreciate it if for a few months we can keep the comments off of this file, until WSL starts to support Ubuntu 26.04 LTS (Resolute Raccoon), apparently scheduled for August 2026.

eggrobin:
> my build environment is in WSL

Oh, that might explain why I could not figure out how to build ICU4J on WSL last month. (I think at some point I eventually managed to put a mvn 3.9 somewhere, but it isn’t the default one. Then I ran into wrong JDK issues and gave up, since the reason why I wanted to build ICU4J on WSL had disappeared.)

(For what it’s worth, note that the unicodetools work fine natively on Windows, that’s what I use on all my machines, and I think that’s what John uses too.)

roozbehp:
OK, apparently WSL allows other Linux distributions too. I managed to setup Fedora Linux 44 (which has maven 3.9.11) on WSL and could setup and build unicodetools there. Closing this PR.

markusicu:
Thank you!
Otherwise we could have done this but would have wanted to remember to revert it in a few months.

PR #1364, opened 2026-04-30
Unihan 18 beta

markusicu:
data from @kenlunde https://github.com/unicode-org/unihan-tools/releases/tag/20260423-v18b


eggrobin:
We should add kJapaneseNewVariant and kJapaneseOldVariant to ExtraPropertyAliases etc., but I can do that in another PR if you want.

(I might also take a stab at USourceData, it might be useful to have those keyed on field 2.)

eggrobin (approved):

unicodetools/data/ucd/dev/Unihan/kDefinition.txt
```diff
@@ -18514,7 +18514,7 @@
 950B	point of spear, sharp point
 950C	zinc (element 30, Zn)
 950D	matte, a crude mixture of sulfides formed in smelting sulfide ores
-950E	caesium/cesium (element 55, Cs)
+950E	californium (element 98, Cf)
```
eggrobin, line 18517: The diff here makes me think of this pub trivia question for Unicadetti: Which General_Category value aliases are also chemical elements?

Issue #1365, opened 2026-04-30
make tests faster


markusicu:
- [x] In CI, two or more workflows build a cldr-unicodetools.tgz which takes over 2 minutes. Do we need all of these?
    - Elango thinks that we probably don't need the one in cli-build-instructions.yml
- [ ] Maven test in parallel, using multiple processes
    - https://github.com/unicode-org/unicodetools/pull/1368
    - but made tests flaky, and hid failures...
    - https://github.com/unicode-org/unicodetools/pull/1376
- [ ] TestTestUnicodeInvariants takes over 3 minutes.
    - [ ] Split it into multiple test classes for better granularity / more parallel testing.
        - tried that in https://github.com/unicode-org/unicodetools/pull/1369, some tests failing
        - dependency on an earlier test run? race condition? debug?
        - maybe try to just split out the security invariants?
    - [ ] Idea: Try to turn many of the test cases into a parameterized test. Read each per-proposal file as a separate test case. Maybe even turn the large file into multiple test cases?
- [ ] TestPropertyAccess takes 2.5 minutes.
- [ ] jsptest.TestProperties takes half a minute.
- [ ] In CI, the longest-running task is [Build JSPs/ Regenerate the property cache files](https://github.com/unicode-org/unicodetools/blob/0ff07d52fb7b345ac920c880fc3b4ee4cd203a24/.github/workflows/build-jsp.yml#L65)
    - Takes some 12 minutes!
    - Maybe try to profile the code and look for good optimization ROI. Might end up making all of the tools/tests faster.

### Notes

TestTestUnicodeInvariants.testAdditionComparisons() has this comment:
```
        // Ideally, this would be a @ParameterizedTest, and we would use the Dtest parameter of the
        // mvn command line to filter a specific test.
        // Unfortunately, this does not appear to be possible with JUnit 5 and Surefire, see
        // https://stackoverflow.com/questions/69198795/junit5-running-a-single-instance-of-parameterized-test,
        // so we have one big test and we filter by hand.
        final String rmgIssueFilter = CldrUtility.getProperty("RMG_ISSUE", null);
```

PR #1368, opened 2026-05-01
surefire forkCount: parallel testing

markusicu:
Trying to make tests run a bit faster by using the maven-surefire-plugin `forkCount` option to use multiple JVM processes in parallel. As before, subsequent test classes share JVMs.
- https://github.com/unicode-org/unicodetools/issues/1365
- https://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html#forked-test-execution
- CI runtime of running most tests in the out-of-source build goes down from about 8min to about 5min

Note: "println()" output from parallel unit tests gets intermingled, making the test log harder to read. When that's a problem, run just the failing unit test, or temporarily go back to forkCount=1.

One unit test failed here:
```
Error:  org.unicode.propstest.TestProperties.TestDefaults -- Time elapsed: 0.189 s <<< FAILURE!
org.opentest4j.AssertionFailedError: Null Link_Bracket for U+FFFF but property is String with default value type LITERAL. Add an @missing line to ExtraPropertyValueAliases.txt if needed. ==> expected: not <null>
```

It also failed locally for me when just running that one unit test alone. Apparently when it runs after _something else_ it passes.

I fixed it by adding an `@missing` value for Link_Bracket as the error message suggests.
- I did not try to add (generate) the `@missing` line in the data file itself.
- I did not try to figure out why this unit test failed when run alone, or which other previously-run unit test makes it pass.
- @eggrobin @macchiati FYI

### DIFF
--- a/.github/workflows/build-jsp.yml
+++ b/.github/workflows/build-jsp.yml
@@ -12,7 +12,8 @@ on:
     tags:
     - '*'
 jobs:
-  build:
+  build-jsp:
+    name: Build JSPs
     runs-on: ubuntu-latest
     permissions:
       contents: read
--- a/.github/workflows/cli-build-instructions.yml
+++ b/.github/workflows/cli-build-instructions.yml
@@ -192,10 +192,10 @@ jobs:
 
       # Only test once we know the UCD is internally consistent.
       # MakeUnicodeFiles is much faster than this anyway.
-      - name: Test invariants
+      - name: Run most tests
         run: |
           cd unicodetools/mine/src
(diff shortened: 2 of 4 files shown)


echeran (approved):
LGTM

eggrobin:
I think this made the tests flaky, possibly because they use the BIN cache. I had to do a rerun on https://github.com/unicode-org/unicodetools/pull/1370.

We should probably revert this until we figure out a correct way to do it, flakiness will waste way more time than this saves.

markusicu:
> I think this made the tests flaky, possibly because they use the BIN cache. I had to do a rerun on #1370.
> 
> We should probably revert this until we figure out a correct way to do it, flakiness will waste way more time than this saves.

Sadly, it looks like you are right. I am doing a second rerun now on https://github.com/unicode-org/unicodetools/pull/1374 :-(

PR #1369, opened 2026-05-01
split TestTestUnicodeInvariants into multiple classes

markusicu:
...  for better granularity / more parallel testing.
- https://github.com/unicode-org/unicodetools/issues/1365

Parallel testing (forkCount>1) runs test *classes* in parallel. Splitting a very long-running class lets the smaller, shorter-running classes finish faster.

TODO: some tests fail :-(


eggrobin:
How does this improve the granularity, when we already run individual functions from the various CI workflows?

markusicu:
> How does this improve the granularity, when we already run individual functions from the various CI workflows?

Parallel testing (forkCount>1) runs test *classes* in parallel. Splitting a very long-running class lets the smaller, shorter-running classes finish faster.

PR #1370, opened 2026-05-04
Add kJapanese(Old|New)Variant

eggrobin:
Follow-up on #1364.

Also parse UAX45 while we are at it, and make the JSPs work again after all these toolchain upgrades.


markusicu (approved):
rslgtm

PR #1376, opened 2026-05-05
mvn test serially again

markusicu:
Running tests in parallel currently makes tests flaky :-(
- https://github.com/unicode-org/unicodetools/pull/1368#issuecomment-4371865240
- https://github.com/unicode-org/unicodetools/issues/1365

Probably races among different tests reading/building/writing data files?

- [x] Approver: Feel free to merge on my behalf
    - [x] rebase & merge one or more commits

PR #1379, opened 2026-05-05
Regices for kJapaneseMeowVariant and the U-Source pseudoproperties

eggrobin:
Previously in #1376:
> _@markusicu_. @eggrobin Why would I get a [new test failure](https://github.com/unicode-org/unicodetools/actions/runs/25394130646/job/74476196367?pr=1376) _when all I do is turn off parallel testing_?
> _@eggrobin_. I was wondering why that one had not failed on me in https://github.com/unicode-org/unicodetools/pull/1370 (new properties now require entries in IndexPropertyRegex).
> Apparently the answer is that the parallel testing was hiding real failures.
> _@markusicu_. So... can I merge this one, and then you fix these issues?
> _@eggrobin_. Yes. But dinner will happen in between.

One beef Wellington later, here is the fix.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/props/UcdProperty.java
+++ b/unicodetools/src/main/java/org/unicode/props/UcdProperty.java
@@ -112,7 +112,11 @@ public enum UcdProperty {
     kJapaneseNewVariant(
             PropertyType.String, DerivedPropertyStatus.Provisional, "cjkJapaneseNewVariant"),
     kJapaneseOldVariant(
-            PropertyType.String, DerivedPropertyStatus.Provisional, "cjkJapaneseOldVariant"),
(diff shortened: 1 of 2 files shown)


markusicu:
Failure: GenerateEnums has changes that need to be committed.

And a Build JSPs failure that I don't understand.

eggrobin:
> Failure: GenerateEnums has changes that need to be committed.

Oops. As expected, I changed the cardinalities. Done.

> And a Build JSPs failure that I don't understand.

That looks like some infrastructure in shambles beyond our control…

markusicu (approved):

markusicu:
@jowilco FYI

Issue #194, opened 2022-01-25
Upgrade UnicodeProperties


macchiati:
Replace the use of old APIs in UCD.java and related files by using the API in the org.unicode.props package instead.

**Background**
V1: The code in the org.unicode.text.UCD and UCA packages, and related packages is very old. It was written when Java was very young, and not nearly as powerful as it later became. The performance was also a fraction of what is currently available on modern computers, which necessitated a lot of hacks for speed. It is ugly to maintain and extend; eg, adding a new property value requires changing some int constants in UCD_Types (the code predated enums) and some string arrays in UCD_Names.

`
UCD ucd = UCD.make("14.0.0");
byte intEnum = ucd.getBidiClass(codepoint);
String propValue1 = UCD.getBidiClassID_fromIndex(intEnum, UCD_Types.LONG);
String propValue1s = UCD.getBidiClassID_fromIndex(intEnum, UCD_Types.SHORT);
System.out.println("V1: <bidi class> = " + propValue1 + ", " + propValue1s);
`

V2: A layer using UnicodeProperty was built on top of this, which made it somewhat easier to write code without knowing as much about the underlying implementation.

`
ToolUnicodePropertySource propertySource = ToolUnicodePropertySource.make("14.0.0");
UnicodeProperty property2 = propertySource.getProperty("BidiClass");
List<String> propNames2 = property2.getNameAliases();
String propValue2 = property2.getValue(codepoint);
List<String> propValueAliases = property2.getValueAliases(propValue2);
System.out.println("V2: " + propNames2.get(1) + " = " + propValueAliases.get(1) + ", " + propValueAliases.get(0));
`

V3: Later, the tooling in org.unicode.props was developed. It uses a much more general parsing mechanism — as data-driven as possible — and more modern API (more type-safe, ...). It has a few more advantages; it only parses (and caches) the files it needs, and the caches are flushed if any of the data files are more recent. Etc.

`
IndexUnicodeProperties unicodeProperties = IndexUnicodeProperties.make(Age_Values.V14_0); // can also use VersionInfo
UnicodeProperty property3 = unicodeProperties.getProperty(UcdProperty.Bidi_Class);
UnicodeMap<Bidi_Class_Values> map = unicodeProperties.loadEnum(UcdProperty.Bidi_Class, UcdPropertyValues.Bidi_Class_Values.class);
Bidi_Class_Values propValue3e = map.get(codepoint); // get typesafe map
[...]

PR #1394, opened 2026-05-08
Normalizer: switch from UCD backend to IUP

markusicu:
The Normalizer loads data from one of two "sources":
- NormalizationDataStandard which reads from class UCD (old properties API)
- NormalizationDataIUP which reads from IndexUnicodeProperties ("new")

The code is equivalent. (I compared the two classes.) However, duplication means unnecessary maintenance, and we want to get away from the old properties APIs:
- https://github.com/unicode-org/unicodetools/issues/194

I also found that class Normalizer is mutable, and there are two call sites that modify it. This feels scary.

Here:
- Delete NormalizationDataStandard, use NormalizationDataIUP instead.
- Construct or fetch from enum NormalizationMode, not a UCD_Types integer constant.
- Lazy-init & cache each standard Normalizer for the latestVersion.
- Make Normalizer Freezable, and freeze the cached ones. (Moving to a Builder model would take more work which does not seem justified here.)
- Construct a new one when it is mutated, or when it is always made for an old version.
- "getOrMake" a Normalizer as much as possible, rather than always making a new one, to hopefully often get cached ones.
- Where possible, pass a Normalizer, or a NormalizationForm, instead of a UCD_Types integer.


eggrobin (dismissed):

unicodetools/src/main/java/org/unicode/text/UCA/UCA.java
```diff
@@ -266,7 +266,11 @@ public UCA(String sourceFile, String unicodeVersion) throws java.io.IOException
 
         // load the normalizer
         if (toD == null) {
-            toD = new Normalizer(UCD_Types.NFD, unicodeVersion);
+            // TODO: We should remove the unicodeVersion argument and
+            // not try to create a collator for an old Unicode version
+            // because we do not track changes to special weight values and algorithm edge cases.
```
eggrobin, line 271: I would actually like to be able to see changes to collation in the tools… (and there I mostly don’t care about the specials and edge cases).

I had started doing something about that, I should find where I put that code and look at it again…

eggrobin (dismissed):

eggrobin (approved):

unicodetools/src/main/java/org/unicode/text/UCA/UCA.java
markusicu: A simple thing we could do is to check https://github.com/unicode-org/cldr/blob/main/common/uca/FractionalUCA_blanked.txt into this repo, so that sort order diffs are visible.

eggrobin: I found the branch where I played around with that: https://github.com/eggrobin/unicodetools/tree/ducet-jsps (my plan was to generate collation foldings at various levels as properties, and maybe a nextUp/nextDown or one per level, so that the changes to equivalence classes and ordering can be understood and queried with the character.jsp history view and versioned UnicodeSet queries).

PR #1402, opened 2026-05-13
UCDXML 18 20260513

markusicu:
Generated all six files, but we carry only two in the repo.
I will regenerate again closer to the beta.


echeran (approved):
LGTM

josh-hadley (approved):

PR #1374, opened 2026-05-05
UCA 18 beta: UCD changes, U+FFFE, U+FFFF, Tibetan contractions

markusicu:
Collation files, should be about ready for Unicode 18 beta UCA.

DUCET files and DUCET-code changes from KenW for
- UCD changes from UTC-187
- upstreaming CLDR behavior for U+FFFE, U+FFFF, Tibetan contractions

plus Unicode Tools UCA code changes to adjust for the upstreamed features which are now no longer CLDR-specific tailorings.

PR #1413, opened 2026-05-18
tweak UCA conformance tests

### LINKED ISSUE #1101: Better testing for U+0CC2 and U+0DCF collation edge cases
Two characters in the root collation have the special property that they occur in the middle of a contraction without also occurring at the start of a contraction. Therefore, checking if U+0CC2 and U+0DCF may start a contraction isn't a sufficient check for whether they can contract the next character.

This special case is worthwhile to test for explicitly, since this special case may cause a bug when skipping over the identical prefix of strings to be compared in a collator.

For U+0CC2, I suggest manually injecting the following into the collation test suite:
0CC8 0CC6 0CC2 0CD6
is less than
0CC8 0CC6 0CC2 0CD5

Here the inital 0CC8 is any filler character just in case to make the interesting case not occur right at the start of the input.

A similar case can probably be constructed for U+0DCF.

markusicu:
a) Use the usual function for adding custom test strings: addStringX() not addStringY().
- addStringX() adds additional string variants, but also prints comments with the names of all of the input string characters.
- Just calling addStringY() worked for the test strings but omitted the trailing characters from the comments, which was confusing.
- Follow-up on https://github.com/unicode-org/unicodetools/pull/1374

b) Add a couple of interesting test cases.
- Fixes https://github.com/unicode-org/unicodetools/issues/1101


echeran (approved):
rslgtm

PR #1414, opened 2026-05-19
align /Public/version/... readmes

markusicu:
Fixes https://github.com/unicode-org/properties/issues/8

### DIFF
--- a/unicodetools/data/emoji/dev/ReadMe.txt
+++ b/unicodetools/data/emoji/dev/ReadMe.txt
@@ -3,7 +3,10 @@
 # Unicode and the Unicode Logo are registered trademarks of Unicode, Inc. in the U.S. and other countries.
 # For terms of use and license, see https://www.unicode.org/terms_of_use.html
 
-This directory contains PUB_STATUS data files for Unicode Emoji, Version EMOJI_VER
+This directory contains PUB_STATUS data files for version EMOJI_VER of
+UTS #51: Unicode Emoji
+
+See https://www.unicode.org/reports/tr51/ .
 
 PUBLIC_EMOJI/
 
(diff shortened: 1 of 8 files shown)


unicodetools/data/emoji/dev/ReadMe.txt
```diff
@@ -3,19 +3,20 @@
 # Unicode and the Unicode Logo are registered trademarks of Unicode, Inc. in the U.S. and other countries.
 # For terms of use and license, see https://www.unicode.org/terms_of_use.html
 
-This directory contains PUB_STATUS data files for Unicode Emoji, Version EMOJI_VER
+This directory contains PUB_STATUS data files for version EMOJI_VER of
+UTS #51: Unicode Emoji
+
+See https://www.unicode.org/reports/tr51/ .
```
nedley: I’m not a huge fan of the floating period, and it seems like some files didn’t bother with a period at all for trailing links.

markusicu: I am not a huge fan either, but I don't want to trip people up when their software doesn't do good linkification.

I prefer finishing the sentence with a period, but I could live without it.

eggrobin: Even if they do poor linkification, [https://www.unicode.org/reports/tr51/.](https://www.unicode.org/reports/tr51/.) goes to the right place…

asmusf: If "/." goes to the right place, then I'd prefer not to have an odd space. I'm also happy to live without any period here.

markusicu: I removed the spaces. PTAL

nedley (approved):

PR #1417, opened 2026-05-19
Unescaped final Soft characters in LinkFormattingTest.txt #527

macchiati:
There was a bad parameter passed in when processing the wiki URLs for use in the test data file. This exposed a bug in how that parameter was used. This corrects that, and gives the parameters more appropriate names (using an enum instead of a naked boolean).

Fixes https://github.com/unicode-org/properties/issues/527

NOTE: I could also regenerate the v17 file if needed. It has ~10 fixes also.

- [X] Approver: Feel free to merge on my behalf

### DIFF
--- a/unicodetools/data/linkification/dev/LinkFormattingTest.txt
+++ b/unicodetools/data/linkification/dev/LinkFormattingTest.txt
@@ -1,5 +1,5 @@
 # LinkFormattingTest.txt
-# Date: 2026-05-19, 16:47:55 GMT 
+# Date: 2026-05-19, 22:01:10 GMT 
 # © 2026 Unicode®, Inc.
 # Unicode and the Unicode Logo are registered trademarks of Unicode, Inc. in the U.S. and other countries.
 # For terms of use and license, see https://www.unicode.org/terms_of_use.html
@@ -193,7 +193,7 @@ https://fiu-vro.wikipedia.org/wiki/Maa_(hod'ot%C3%A4ht)
 https://fiu-vro.wikipedia.org/wiki/Maa_(hod'otäht)
 
 # {𝑺=https:// 𝑯=ty.wikipedia.org 𝑷=wiki 𝑷=’Afirita}
-https://ty.wikipedia.org/wiki/’Afirita
+https://ty.wikipedia.org/wiki/%E2%80%99Afirita
 https://ty.wikipedia.org/wiki/’Afirita
 
 # {𝑺=https:// 𝑯=ab.wikipedia.org 𝑷=wiki 𝑷=Вашингтон,_Џьорџь}
@@ -201,7 +201,7 @@ https://ab.wikipedia.org/wiki/%D0%92%D0%B0%D1%88%D0%B8%D0%BD%D0%B3%D1%82%D0%BE%D
 https://ab.wikipedia.org/wiki/Вашингтон,_Џьорџь
 
 # {𝑺=https:// 𝑯=mni.wikipedia.org 𝑷=wiki 𝑷=ꯅ꯭ꯌꯨ_ꯌꯣꯔ꯭ꯛ_ꯁꯤꯇꯤꯒꯤ_ꯌꯨ.ꯑꯦꯁ.}
-https://mni.wikipedia.org/wiki/ꯅ꯭ꯌꯨ_ꯌꯣꯔ꯭ꯛ_ꯁꯤꯇꯤꯒꯤ_ꯌꯨ.ꯑꯦꯁ.
+https://mni.wikipedia.org/wiki/%EA%AF%85%EA%AF%AD%EA%AF%8C%EA%AF%A8_%EA%AF%8C%EA%AF%A3%EA%AF%94%EA%AF%AD%EA%AF%9B_%EA%AF%81%EA%AF%A4%EA%AF%87%EA%AF%A4%EA%AF%92%EA%AF%A4_%EA%AF%8C%EA%AF%A8.%EA%AF%91%EA%AF%A6%EA%AF%81.
 https://mni.wikipedia.org/wiki/ꯅ꯭ꯌꯨ_ꯌꯣꯔ꯭ꯛ_ꯁꯤꯇꯤꯒꯤ_ꯌꯨ.ꯑꯦꯁ%2E
 
 # {𝑺=https:// 𝑯=azb.wikipedia.org 𝑷=wiki 𝑷=واشینقتن،_دی.سی.}
@@ -209,11 +209,11 @@ https://azb.wikipedia.org/wiki/%D9%88%D8%A7%D8%B4%DB%8C%D9%86%D9%82%D8%AA%D9%86%
 https://azb.wikipedia.org/wiki/واشینقتن،_دی.سی%2E
 
 # {𝑺=https:// 𝑯=mad.wikipedia.org 𝑷=wiki 𝑷=Tasè’}
-https://mad.wikipedia.org/wiki/Tas%C3%A8’
+https://mad.wikipedia.org/wiki/Tas%C3%A8%E2%80%99
 https://mad.wikipedia.org/wiki/Tasè%E2%80%99
 
 # {𝑺=https:// 𝑯=wuu.wikipedia.org 𝑷=wiki 𝑷=聖保羅（巴西）}
-https://wuu.wikipedia.org/wiki/聖保羅（巴西）
+https://wuu.wikipedia.org/wiki/%E8%81%96%E4%BF%9D%E7%BE%85%EF%BC%88%E5%B7%B4%E8%A5%BF%EF%BC%89
 https://wuu.wikipedia.org/wiki/聖保羅（巴西）
 
 # {𝑺=https:// 𝑯=vep.wikipedia.org 𝑷=wiki 𝑷=Brüssel'}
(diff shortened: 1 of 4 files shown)


unicodetools/src/main/java/org/unicode/utilities/LinkUtilities.java
```diff
@@ -580,7 +585,7 @@ public String minimalEscape(boolean atEndOfText, Counter<Integer> escapedCounter
                     }
                 }
                 // fix
-                if (atEndOfText || part != lastPart) {
+                if (endStatus == EndStatus.MEDIAL || part != lastPart) {
```
nedley, line 600: I don’t see any caller passing `EndStatus.MEDIAL`, what purpose does it serve?

unicodetools/data/linkification/dev/LinkFormattingTest.txt
```diff
@@ -202,31 +202,31 @@ https://ab.wikipedia.org/wiki/Вашингтон,_Џьорџь
 
 # {𝑺=https:// 𝑯=mni.wikipedia.org 𝑷=wiki 𝑷=ꯅ꯭ꯌꯨ_ꯌꯣꯔ꯭ꯛ_ꯁꯤꯇꯤꯒꯤ_ꯌꯨ.ꯑꯦꯁ.}
 https://mni.wikipedia.org/wiki/ꯅ꯭ꯌꯨ_ꯌꯣꯔ꯭ꯛ_ꯁꯤꯇꯤꯒꯤ_ꯌꯨ.ꯑꯦꯁ.
-https://mni.wikipedia.org/wiki/ꯅ꯭ꯌꯨ_ꯌꯣꯔ꯭ꯛ_ꯁꯤꯇꯤꯒꯤ_ꯌꯨ.ꯑꯦꯁ.
+https://mni.wikipedia.org/wiki/ꯅ꯭ꯌꯨ_ꯌꯣꯔ꯭ꯛ_ꯁꯤꯇꯤꯒꯤ_ꯌꯨ.ꯑꯦꯁ%2E
 
 # {𝑺=https:// 𝑯=azb.wikipedia.org 𝑷=wiki 𝑷=واشینقتن،_دی.سی.}
 https://azb.wikipedia.org/wiki/%D9%88%D8%A7%D8%B4%DB%8C%D9%86%D9%82%D8%AA%D9%86%D8%8C_%D8%AF%DB%8C.%D8%B3%DB%8C.
-https://azb.wikipedia.org/wiki/واشینقتن،_دی.سی.
+https://azb.wikipedia.org/wiki/واشینقتن،_دی.سی%2E
```
markusicu, line 209: percent-escaping parts of an RTL segment is pretty unreadable.
not a regression here (in fact, a fix), but should we create a PAG issue to discuss this?
as is, this does not work as advertised.

unicodetools/data/linkification/dev/LinkFormattingTest.txt
```diff
@@ -202,31 +202,31 @@ https://ab.wikipedia.org/wiki/Вашингтон,_Џьорџь
 
 # {𝑺=https:// 𝑯=mni.wikipedia.org 𝑷=wiki 𝑷=ꯅ꯭ꯌꯨ_ꯌꯣꯔ꯭ꯛ_ꯁꯤꯇꯤꯒꯤ_ꯌꯨ.ꯑꯦꯁ.}
 https://mni.wikipedia.org/wiki/ꯅ꯭ꯌꯨ_ꯌꯣꯔ꯭ꯛ_ꯁꯤꯇꯤꯒꯤ_ꯌꯨ.ꯑꯦꯁ.
```
markusicu: Not a regression here, but a bug:
- The file format documentation says "The first line of each pair is the source, fully escaped."
- But some of these first lines are not escaped at all.

markusicu:
> NOTE: I could also regenerate the v17 file if needed. It has ~10 fixes also.

I think we should treat published files as immutable, as we normally do.

unicodetools/src/main/java/org/unicode/utilities/LinkUtilities.java
```diff
@@ -580,7 +585,7 @@ public String minimalEscape(boolean atEndOfText, Counter<Integer> escapedCounter
                     }
                 }
                 // fix
```
markusicu: fix what?

unicodetools/src/main/java/org/unicode/utilities/LinkUtilities.java
markusicu, line 1: Please also see [my earlier question](https://github.com/unicode-org/properties/issues/527#issuecomment-3923552339):

Does the formatting code not test whether the formatted link works with the detection code?

unicodetools/data/linkification/dev/LinkFormattingTest.txt
macchiati: I saw that. The simplest option is to add a note that the URLs may reorder in display if they have RTL characters.

We could add a comment line for every such URL that displayed with Bidi HL4 (using bidi isolates). But I don't think we want to include them in the *test* data.

markusicu: The goal of the spec is to minimally escape URLs so that they are as readable as possible.
In and right after RTL segments, percent-escaping a character creates an unreadable jumble.
I think we should discuss this (not here).

unicodetools/src/main/java/org/unicode/utilities/LinkUtilities.java
macchiati: It doesn't. Let me look at adding that.

unicodetools/src/main/java/org/unicode/utilities/LinkUtilities.java
macchiati: spurious
```suggestion
```

unicodetools/src/main/java/org/unicode/utilities/LinkUtilities.java
macchiati: It is for testing. I didn't remove it entirely, because I wanted to make minimal changes to fix the problem.

macchiati:
> > NOTE: I could also regenerate the v17 file if needed. It has ~10 fixes also.
> 
> I think we should treat published files as immutable, as we normally do.

Of course. The question is whether we should make a 17.1 file available, or document the problem so that people don't stumble over it until they implement 18.

unicodetools/data/linkification/dev/LinkFormattingTest.txt
macchiati: I'll check that.

macchiati: It looks like that line is actually the "source", which can be fully escaped or not. The small fix would be to change the documentation to reflect that. 

The larger fix would be to fully escape it. I'll get a sense of whether that is simple or not. (That isn't quite as trivial as it sounds. We need to _mostly_ escape everything, but not syntax characters.  And the same character in the same part may be syntax or might not be. a=b%3Dc in a query, where b%3Dc == 'b=c')

unicodetools/data/linkification/dev/LinkFormattingTest.txt
macchiati: Ok, we can discuss elsewhere. I don't see any obvious choices: essentially any escaped character after a bidi character is going to be garbled. On the other hand, is 

https://azb.wikipedia.org/wiki/%D9%88%D8%A7%D8%B4%DB%8C%D9%86%D9%82%D8%AA%D9%86%D8%8C_%D8%AF%DB%8C.%D8%B3%DB%8C%E2

**more** readable than

https://azb.wikipedia.org/wiki/واشینقتن،_دی.سی%2E

The third option is to not escape final soft characters. But that has usability issues, and doesn't prevent other escapes from appearing within the Part.

unicodetools/src/main/java/org/unicode/utilities/LinkUtilities.java
macchiati: Added tests for that, both in the generation and in the unit tests.

unicodetools/src/main/java/org/unicode/tools/GenerateLinkData.java
```diff
@@ -526,6 +531,15 @@ public static void generateFormattingTestData(String copyrightYear) {
         }
     }
 
+    private static void ensureRoundTrip(UrlInternals sourceInternals, String resultingPath) {
+        UrlInternals targetInternals = UrlInternals.from(resultingPath);
+        if (!targetInternals.equals(sourceInternals)) {
+            throw new IllegalArgumentException("Fails roundtrip");
+        }
+        // for testing
+        String minimal = sourceInternals.minimalEscape(EndStatus.FINAL, null);
```
markusicu: please remove, unless there are useful/necessary side effects other than what someone might see in a debugger when stepping in

markusicu (commented):
output lgtm now.

looks like spotless is unhappy, and maybe some other CI check :-/

unicodetools/data/linkification/dev/LinkFormattingTest.txt
macchiati: I was able to fix this.

unicodetools/data/linkification/dev/LinkFormattingTest.txt
macchiati: I'll make this into an issue in the props repo.

markusicu: I already did, and I tagged you on it.

unicodetools/src/main/java/org/unicode/tools/GenerateLinkData.java
markusicu: this one is not done yet

macchiati:
Is that important enough to block the PR? I want to return to this code in
making a draft for ICU link apis, and will do some more thorough cleanupvas
part of that.

unicodetools/src/main/java/org/unicode/tools/GenerateLinkData.java
macchiati: Ok, removing

macchiati:
I think this should do it; I am also escaping any final Soft characters in the full-escaped string.

markusicu (approved):
lgtm tnx

macchiati:
BTW, if we want to be more explicit, we could make the following change in the header of the test file.

OLD

> \# The fully-escaped field percent-escapes all literal syntax characters and all characters above ASCII.

NEW

> \# The fully-escaped field percent-escapes all code points based on https://url.spec.whatwg.org/#percent-encoded-bytes.
> \# This means all literal syntax characters in each Part and all code points above ASCII.
> \# It also percent-escapes the last character, if it is Link_Term:Soft.

markusicu:
> BTW, if we want to be more explicit, we could make the following change in the header of the test file.

good idea

PR #1431, opened 2026-05-27
Generate a charindex.html suitable for its final location, generate a -draft version, and document it all

eggrobin:
See:
- tools meeting minutes, 2026-05-26;
- RMG minutes, 2026-05-27;
- https://www.unicode.org/charts/charindex.html;
- https://www.unicode.org/charts/charindex-draft.html.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/text/tools/Indexer.java
+++ b/unicodetools/src/main/java/org/unicode/text/tools/Indexer.java
@@ -368,8 +368,7 @@ public class Indexer {
                         // Override word breaking of ' and - in appropriate contexts so that
                         // radical/stroke indices are atomic.
                         // With ICU4J we could do that with custom segmentation rules, but we need
-                        // to
-                        // have the same segmentation in the JavaScript where we do not have that
+                        // to have the same segmentation in the JavaScript where we do not have that
                         // luxury, so poor man’s tailoring by segmenting a mangled string it is.
                         final String mangledForWordBreak =
                                 snippet.replaceAll("\\.-", ".0")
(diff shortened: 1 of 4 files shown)


eggrobin:
Note: ?w=1 is your friend to review the changes to Indexer.java: https://github.com/unicode-org/unicodetools/pull/1431/changes?w=1.

unicodetools/src/main/java/org/unicode/text/tools/Indexer.java
```diff
@@ -38,730 +39,721 @@
...
+
+        VersionedIndexer(
+                VersionInfo version, ReleasePhase phase, String chartsRoot, String filename) {
+            this.version = version;
+            this.phase = phase;
+            this.chartsRoot = chartsRoot;
+            this.filename = filename;
+            IUP = IndexUnicodeProperties.make(version);
+            NEW_CHARACTERS =
+                    IndexUnicodeProperties.make(Settings.LAST_VERSION_INFO)
```
markusicu: It seems wrong to keep the last version hardcoded. Now that the current `version` is variable, its relation to the LAST_VERSION may not make sense. I think you should pass in both the version and the previousVersion.

eggrobin: That makes sense.

(Of course, this has no effect, as we only called this with version=LAST_VERSION_INFO with a value of versionPhase that meant we did not look at NEW_CHARACTERS, but making sense only when called in certain ways does feel wrong.)

Done.

markusicu (approved):

PR #1432, opened 2026-05-27
charindex feedback from John, part 1: show prefix matches even if there are exact matches (but after the exact matches)

eggrobin:
See:
- https://www.unicode.org/charts/charindex-draft.html?q=du (100 results, was 22);
- https://www.unicode.org/charts/charindex-draft.html?q=dup (10 results, was 2);
- https://www.unicode.org/charts/charindex-draft.html?q=dupl (7, unchanged).


markusicu (approved):
rslgtm

PR #1433, opened 2026-05-28
charindex: Bold matching words, prefer matches in order with fewer intervening words, do not permute entries that have too much punctuation

eggrobin:
@jowilco had pointed out the infelicitous entry
> DU), Inuktitut (TAI), Athapascan (DI), Carrier (

And I have been annoyed by ⋝ being before ≥ when searching for `greater or equal`.

query: `du`
| before | after |
|---|---|
| image | image |

query: `greater or equal`
| before | after |
|---|---|
| image | image |

query: `plus minus`
| before | after |
|---|---|
| image | image |

query: `a grave`
| before | after |
|---|---|
| image | image |

query: `a dot`
| before | after |
|---|---|
| image | image |

query: `alpha perispomeni`
| before | after |
|---|---|
| image |  image |

query: `greater equal`
| before | after |
|---|---|
| image | image |


kemalaysegulonay-del (changes requested):
Değiştir

unicodetools/src/main/resources/org/unicode/text/tools/charindex.js
```diff
@@ -135,13 +138,54 @@ async function search(/**@type {string}*/ query) {
...
+      //   — increasing number of interspersed non query words;
+      // — the stuff after the segment that matches the query (the thagomizer),
+      //   in alphabetical order;
+      // — the segment that matches the query, in alphabetical order.
+      // — the stuff before the segment that matches the query (the head), in
+      //   alphabetical order;
       return [i,
-              snippet.substring(getPivot(i)) + ' \uFFFE ' +
-              snippet.substring(0, getPivot(i))];
+              String.fromCodePoint(0x10FFFE - orderedMatches)    + '\uFFFE' +
```
markusicu, line 184: someone seems to have had some useful idea for using U+FFFE in collation...

(_might_ be worth a comment)

markusicu (approved):
very nice improvements!

glancing at how you do it but not trying to understand it all

unicodetools/src/main/resources/org/unicode/text/tools/charindex.js
eggrobin: Well, it is pretty clearly documented in UTS10 now, and this is a textbook usage of FFFE for multi-level keys.

PR #1435, opened 2026-06-03
delete unused Main class dispatch options & related code

### LINKED ISSUE #1401: review/delete code called from UCA/Main & UCD/Main dispatch
We have two Main.java classes that can be called with various command-line arguments and dispatch to other classes. I suspect that some of these are obsolete. They don't show up as "unused" because they are called in the dispatch.

Review each dispatch item and see if we still need that code.

@macchiati could you please help review/evaluate?

- UCD/Main: https://github.com/unicode-org/unicodetools/blob/main/unicodetools/src/main/java/org/unicode/text/UCD/Main.java
- UCA/Main: https://github.com/unicode-org/unicodetools/blob/main/unicodetools/src/main/java/org/unicode/text/UCA/Main.java

markusicu:
Fixes https://github.com/unicode-org/unicodetools/issues/1401

See
- https://github.com/unicode-org/unicodetools/issues/1401

(repeat link to get a readable title)

Best reviewed one commit at a time.
I intend to rebase+merge these separate commits.


nedley (approved):

macchiati (approved):
no objections

eggrobin:
That is a Δ to celebrate: 
image

PR #1436, opened 2026-06-05
delete some unused UCD code

markusicu:
Towards
- https://github.com/unicode-org/unicodetools/issues/194

### DIFF
--- a/unicodetools/src/main/java/org/unicode/text/UCD/ToolUnicodePropertySource.java
+++ b/unicodetools/src/main/java/org/unicode/text/UCD/ToolUnicodePropertySource.java
@@ -24,7 +24,6 @@ import java.util.Locale;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
-import java.util.regex.Pattern;
 import org.unicode.props.IndexUnicodeProperties;
 import org.unicode.props.UcdProperty;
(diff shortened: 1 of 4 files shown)


unicodetools/src/main/java/org/unicode/text/UCD/ToolUnicodePropertySource.java
```diff
@@ -2161,7 +2160,7 @@ public List<String> _getValueAliases(String valueAlias, List<String> result) {
                                 throw new IllegalArgumentException("Internal Error: " + prop);
                         }
                     } catch (final ArrayIndexOutOfBoundsException e) {
-                        continue;
+                        // continue;
```
nedley, line 2163: So this was working around an error before? If so and it’s no longer being worked around, I would argue in favor of removing the try block entirely.

markusicu: I didn't take a good look at this. It was and is catching/ignoring an exception.

IntelliJ shows lots of warnings and suggestions, and I chose to not ignore this one where it complained that the "continue" was redundant -- the loop continues anyway. I commented it out to keep the intent visible. IntelliJ is microscopically happier...

echeran (approved):

PR #1448, opened 2026-06-30
Another LB test case

eggrobin:
This exercises an ICU bug discovered by @robertbastian.


unicodetools/src/main/java/org/unicode/text/UCD/GenerateBreakTest.java
```diff
@@ -1652,6 +1652,10 @@ public GenerateLineBreakTest(UCD ucd, Segmenter.Target target) {
                         // inside.
                         "Une nuit, —\u00A0c’est toujours la nuit dans le tombeau,\u00A0— Il s’éveilla.",
                         "Une nuit, –\u00A0c’est toujours la nuit dans le tombeau,\u00A0– Il s’éveilla.",
+                        // Regional indicator pairs with ZWJs inside the first pair and joining the
+                        // first pair to the second.  This was incorrectly segmented by ICU for man
```
robertbastian: 👨 

```suggestion
                        // first pair to the second.  This was incorrectly segmented by ICU for many
```

unicodetools/data/ucd/dev/auxiliary/LineBreakTest.txt
```diff
@@ -19362,9 +19362,10 @@
...
+× 1F1FF × 200D × 1F1FF × 200D × 1F1FF × 1F1FF ÷	#  × [2.0] REGIONAL INDICATOR SYMBOL LETTER Z (RI) × [9.0] ZERO WIDTH JOINER (ZWJ) × [8.1] REGIONAL INDICATOR SYMBOL LETTER Z (RI) × [9.0] ZERO WIDTH JOINER (ZWJ) × [8.1] REGIONAL INDICATOR SYMBOL LETTER Z (RI) × [30.11] REGIONAL INDICATOR SYMBOL LETTER Z (RI) ÷ [3.0]
```
robertbastian, line 19365: it doesn't matter, but `REGIONAL INDICATOR SYMBOL LETTER A` is probably the more canonical symbol

unicodetools/data/ucd/dev/auxiliary/LineBreakTest.html
```diff
@@ -4653,11 +4653,20 @@ <h3><a href='#samples' name='samples'>Sample Strings</a></h3>
 
 </font></td></tr>
 <tr><th style='text-align:right'><a href='#s295' name='s295'>295</a></th><td><font size='5'>
+<span title='2.0'><span>&nbsp;</span>&nbsp;</span><span title='U+1F1FF REGIONAL INDICATOR SYMBOL LETTER Z (RI)'>&#x1F1FF;</span><span title='9.0'><span>&nbsp;</span>&nbsp;</span>
```
robertbastian, line 4656: is this manually kept in sync with `LineBreakTest.txt`?

robertbastian (approved):
do we have any tests that line/sentence/word breaks are a subset of grapheme cluster breaks? do the specs imply that anywhere?

unicodetools/data/ucd/dev/auxiliary/LineBreakTest.txt
eggrobin: Both AA and ZZ are user-assigned in ISO 3166-1 so it does not matter much, but ZZ is assigned by CLDR as  Unknown or Invalid, whereas AA is not used, so ZZ seems nicer as a test case.

unicodetools/data/ucd/dev/auxiliary/LineBreakTest.html
eggrobin: No, both files are generated by MakeUnicodeFiles (and the update is enforced by CI), the only manual change here was in the Java.

robertbastian: Consider marking them as generated: https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

eggrobin:
> do we have any tests that line/sentence/word breaks are a subset of grapheme cluster breaks?

We do not, because 
> do the specs imply that anywhere?

I hope they do not, because that is not true. There is a very long standing battery of action items to do something about this, see https://github.com/unicode-org/properties/issues/19. @charlotte-buff and I have been looking at that issue recently. I think eventually this will involve moving LB30a before LB9, meaning that the correct behaviour on this test case will change. We are always a bit skittish when it comes to touching the non-tailorable part, but if ICU has been failing to implement this interaction between the rules for eight years, that rather weakens the argument that touching it is a problem for implementers.

unicodetools/data/ucd/dev/auxiliary/LineBreakTest.html
eggrobin: This HTML file is a bit noisy, but in general in this repository we want to review the generated files—often more so than the code.

robertbastian: and they will still show up in PRs, just collapsed by default, so you can review the non-generated code first

eggrobin: I am familiar with the way they show up (I have occasionally looked at ICU4X PRs). That’s not what I am looking for here, where changes to the data files means changes to release products that may have implications of perpetual stability (whereas changes to the code that have no effect on the data files are only a concern for the few people who actually work on this codebase).

markusicu: I agree. We need to pay attention to generated files.

markusicu (approved):

PR #1482, opened 2026-07-15
East_Asian_Width 3.1.1 ætiology

eggrobin:
This came up while writing the PAG review of https://github.com/unicode-org/sew/issues/653, see https://github.com/unicode-org/unicodetools/pull/1474.

### DIFF
--- a/unicodetools/src/main/resources/org/unicode/text/tools/reasons_manual.txt
+++ b/unicodetools/src/main/resources/org/unicode/text/tools/reasons_manual.txt
@@ -1,3 +1,10 @@
+# [87-M20] Motion: Update the East Asian data files changing the list in L2/01-189, 00AE to 273D to A, and all Jamo to W, post, and include in Unicode 3.1.1.
+# L2/01-189 Updates to EastAsianWidth.
+# First group (to ea=A).  U+00AF not changed.
+East_Asian_Width ; 3.1.1 ; [\u00AE\u014B\u02C4\u02DF\u2022\u2024\u203E\u2116\u2153\u215C\u215D\u21B8\u21B9\u21E7\u273D] ; L2/01-189, 87-M20
+# Second group (to ea=W).  U+00AF, U+FF78 not changed.
+East_Asian_Width ; 3.1.1 ; [[\u2024\u2022\u203E\u02C4\u273D\u02DF\u2116\u21E7\u21B8\u21B9][\u2116][\u2116\u2116][\u014B\u2153\u215C\u215D\u2116\u00AE]] ; L2/01-189, 87-M20
+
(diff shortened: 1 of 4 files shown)


unicodetools/src/main/resources/org/unicode/text/tools/reasons_manual.txt
```diff
@@ -1,3 +1,10 @@
+# [87-M20] Motion: Update the East Asian data files changing the list in L2/01-189, 00AE to 273D to A, and all Jamo to W, post, and include in Unicode 3.1.1.
+# L2/01-189 Updates to EastAsianWidth.
+# First group (to ea=A).  U+00AF not changed.
+East_Asian_Width ; 3.1.1 ; [\u00AE\u014B\u02C4\u02DF\u2022\u2024\u203E\u2116\u2153\u215C\u215D\u21B8\u21B9\u21E7\u273D] ; L2/01-189, 87-M20
```
robertbastian, line 4: I find `[®ŋ˄˟•․‾№⅓⅜⅝↸↹⇧✽]` more readable than whatever these escapes are. with `[®ŋ˄˟•․‾№⅓⅜⅝↸↹⇧✽]` I can see that it's the same set across the files

robertbastian (commented):
not sure what you want me to review. I don't have access to the linked issue

unicodetools/src/main/resources/org/unicode/text/tools/reasons_manual.txt
eggrobin: This is mostly a write-only input file, and these code points are from L2/01-189 (I copied them from there). The output file (reasons_auto) tries to be readable and uses [®ŋ˄˟•․‾№⅓⅜⅝↸↹⇧✽].

eggrobin:
> not sure what you want me to review

(Getting off a train right now, will give you more context when I’m back at a keyboard)

eggrobin:
@robertbastian: What is happening here is that I found out why 15 characters have the East_Asian_Width assignment they have, (based on proposal L2/01-189 and a decision UTC-87-M20 from 2001), and I am recording that in a machine-readable form so that next time we come across one of those we don’t have to wonder why their East_Asian_Width assignment is what it is and why it changed in 3.1.1.

Here’s the broader context around this tool—Peter’s digest of the minutes do not include anything before UTC 90, which is part of why this has to come from reasons_manual, and why the generated comment in reasons_auto ends up being `null`—:
https://github.com/unicode-org/unicodetools/blob/d758f82179cbdd283c16d0195bd482cbce99ac0d/unicodetools/src/main/java/org/unicode/text/tools/%C3%86tiologer.java#L40-L82

robertbastian (approved):

PR #1483, opened 2026-07-24
Support for constructing segmenters for older versions of the standard, with data for 17

eggrobin:
See the changes to Segmenter.java, and the rules in Segmenter-17.0.0.txt.

Data for older versions coming later (I would like a version selector in https://util.unicode.org/UnicodeJsps/breaks.jsp just like we have one in https://util.unicode.org/UnicodeJsps/bidic.jsp), but I really need to have a way to access the 17.0 rules so that I can produce the state machines for PRI-555 without working on a fork from March (and simultaneously on a more recent fork to produce the 18.0 state machines). Maybe I will tweak the syntax of the Segmenter* files a little bit before I add the older stuff, so that I can more easily produce them from the annexes.

The rest of the changes are due to the removal of some vestiges of separate SegmenterCldr rule files (removed in https://github.com/unicode-org/unicodetools/pull/947). It is now clear that CLDR tailorings will not be separate rule sets in the style of UAX-14 and UAX-29.


unicodetools/src/main/java/org/unicode/tools/Segmenter.java
```diff
@@ -66,27 +62,21 @@ public enum Target {
...
-    }
-
-    public static Builder make(UnicodeProperty.Factory propFactory, String type, Target target) {
+    public static Builder make(VersionInfo version, String type) {
         String sourceFileName =
-                target == Target.FOR_CLDR ? "SegmenterCldr.txt" : "SegmenterDefault.txt";
-        Builder b = new Builder(propFactory, target);
+                "Segmenter"
+                        + (version == Settings.LATEST_VERSION_INFO
+                                ? "Default"
```
robertbastian: so you're reappropriating the "Default" from "Default"/"Cldr" to mean current version? why not "SegmenterDefault-17.0.0.txt"?

eggrobin: > so you're reappropriating the "Default" from "Default"/"Cldr" to mean current version?

Yes.

> why not "SegmenterDefault-17.0.0.txt"?

Well, the Default is a bit redundant here now that CLDR doesn’t touch UAX-style rules, so I might as well drop it instead of having to remember in perpetuity that this is Default as opposed to a tailoring which we once thought could be expressed in terms of UAX-style rules.

However, the convention here for the latest version is dev, not Default, so I suppose I should call the latest one Sgementer-dev.txt.

robertbastian (approved):

PR #1491, opened 2026-08-05
Fix some error correction for USourceData 9..12.1

eggrobin:
I had failed to segment `U+2793DUTC-02138` in #1370.

### DIFF
--- a/unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
+++ b/unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
@@ -1798,11 +1798,11 @@ public class PropertyParsingInfo implements Comparable<PropertyParsingInfo> {
                                 && parts[1].equals("UK-2015")) {
                             keyField = "";
                         }
-                        // U+2793 is ➓.  Surely that is nonsense.
+                        // U+2793D is 𧤽, similar to UTC-02139 ⿰角間 (now encoded at 𰴦).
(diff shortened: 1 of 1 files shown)


unicodetools/src/main/java/org/unicode/props/PropertyParsingInfo.java
```diff
@@ -1798,11 +1798,11 @@ private static void parseFields(
                                 && parts[1].equals("UK-2015")) {
                             keyField = "";
                         }
-                        // U+2793 is ➓.  Surely that is nonsense.
+                        // U+2793D is 𧤽, similar to UTC-02139 ⿰角間 (now encoded at 𰴦).
```
markusicu, line 1801: Why does the smooshed value say UTC-02138 but the comment says UTC-02139?

eggrobin: The line in question is
```
UTC-02139;UK-2015;U+2793DUTC-02138;148.12;1145.081;⿰角間;UTCDoc L2/15-260 827
```

So `UTC-02139` (now U+30D26 𰴦, kIRG_UKSource=UK-02139) was marked as similar to the pre-existing character U+2793D. It was also flagged as similar to `UTC-02138` (now UK-02138, encoded as U+30D25 𰴥).

markusicu (approved):

PR #1498, opened 2026-08-06
comment out more Chisoi data

markusicu:
These lines might be harmless/ignored if the characters don't exist, but let's try to have clean data.

Will regenerate UCA & UCDXML later.


roozbehp (approved):
LGTM

PR #1504, opened 2026-08-17
Matched version number to Unicode 17, and removed U+2612 after discus…

jowilco:
…sion with David

TODO: Please describe your changes here.


davidcarlisle (approved):
Confirming that this does as advertised (renames to 17 and drops U+2612). 

The alternative of adding U+2610 and U+2611 to match U+2612 was also discussed in email, I have a mild preference for the version here, dropping all three, but don't mind either way.

Approving but I'll leave the merge to @asmusf

eggrobin:
This should probably be on top of a rename of revision-16 to revision-17 then.

eggrobin:
Also, this needs updating to match:
https://github.com/unicode-org/unicodetools/blob/240b4f2c91099efd743625e6198e15d04309c733/unicodetools/src/main/java/org/unicode/text/utility/Utility.java#L1600-L1631

Issue #1127, opened 2025-05-02
jsp outdated ExtraProperty(Value)Aliases


markusicu:
I just found that the jsp tree includes very outdated versions of ExtraProperty(Value)Aliases.txt.
Are these still used?
If so, is there a reason that they are out of date, other than us forgetting that they exist there separately?

```
meld unicodetools/src/main/resources/org/unicode/props/ExtraPropertyAliases.txt UnicodeJsps/src/main/resources/org/unicode/jsp/data/ExtraPropertyAliases.txt

meld unicodetools/src/main/resources/org/unicode/props/ExtraPropertyValueAliases.txt UnicodeJsps/src/main/resources/org/unicode/jsp/data/ExtraPropertyValueAliases.txt
```

@eggrobin @macchiati FYI

macchiati:
A lot of the jsp/ stuff dates from when they were completely separate
projects. I see no reason to keep the duplicates in jsp/ nowadays

Issue #716, opened 2024-02-22
Users of IndexUnicodeProperty should not put their hands in the raw map


eggrobin:
Some of that was uncovered in https://github.com/unicode-org/unicodetools/pull/681, but the 💩 is really approaching the 🪭 with https://github.com/unicode-org/unicodetools/pull/699.
For now I will probably fix what breaks, but I want to come back to it to deal with that properly, which means:
- IUP.load should be void for users (and the one that returns a UnicodeMap should be called loadRawMap).
- most of the IUP.getWhatever should delegate the resolution to the IndexUnicodeProperty (which needs beefing up), or be removed altogether (callers can just use the UnicodeProperty).

PR #1524, opened 2026-09-08
Collation folding pseudoproperties

eggrobin:
Versioned back to 2.1.9.
Also: total orders of code points, and tertiary weights (with tests that would have caught two recent PAG issues).

Drive-by changes:
- Decomposition_Type from the beginning rather than from 3.1.
- Deleted old copies of (Extra)Property(Value)Aliases that were lurking in the JSPs.
- Various bugfixes to resolve values not stored in the `UnicodeMap`s.

DRAFT (I expect the addition comparison tests to fail in potentially interesting ways with collation folding, and I should probably read this over again).


unicodetools/src/test/java/org/unicode/propstest/TestCollationProperties.java
```diff
@@ -0,0 +1,142 @@
...
+                Character.toString(0x1000C),
+                iup151.getProperty(UcdProperty.UCA_Next_Shifted).getValue("\uFFFF"));
+        // The variables are before FFFE in shifted order in this version; this is a defect in 18.0:
+        // https://github.com/unicode-org/properties/issues/617.
+        final var iup18 = IndexUnicodeProperties.make(VersionInfo.getInstance(18));
+        assertEquals(
+                "18.0 code point following FFFE in shifted order",
+                "ː",
+                iup18.getProperty(UcdProperty.UCA_Next_Shifted).getValue("\uFFFE"));
+        // Sholud be the same as below, U+0009, once that issue is fixed.
```
eggrobin, line 102: Shai-hulud!

eggrobin:
See it in staging:
* [℅](https://unicode-jsps-staging-o2ookmn2oq-uc.a.run.app/UnicodeJsps/character.jsp?a=%E2%84%85&B1=Show&history=full&showDevProperties=1);
* [ß](https://unicode-jsps-staging-o2ookmn2oq-uc.a.run.app/UnicodeJsps/character.jsp?a=%C3%9F&B1=Show&history=full&showDevProperties=1);
* [æ](https://unicode-jsps-staging-o2ookmn2oq-uc.a.run.app/UnicodeJsps/character.jsp?a=%C3%A6&B1=Show&history=full&showDevProperties=1)—I should tell the ætiologer that the 4.1 change to primary folding [comes from](https://github.com/unicode-org/unicodetools/blob/4c29e794a88ade2b005de2640a8ef62de7383243/c/uca/sifter/unidata.txt#L14750-L14754) UTC-101-M5—;
* [Shifted primary equivalents of 一](https://unicode-jsps-staging-o2ookmn2oq-uc.a.run.app/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AUCA_Fold_1_Shifted%3D%E4%B8%80%3A%5D&g=&i=).

roozbehp:
> * Deleted old copies of (Extra)Property(Value)Aliases that were lurking in the JSPs.

So this would fix https://github.com/unicode-org/unicodetools/issues/1127 too?

eggrobin:
Apparently.

roozbehp:
Pasting GPT review:

I reproduced three problems with the folding results, plus an API inconsistency and two smaller issues.

1. **A folding can change the character’s collation order.**

   In [the expansion-folding code](https://github.com/unicode-org/unicodetools/blob/6b8d5dbfa3be5ec2556194b881aa58f746af4b5c/unicodetools/src/main/java/org/unicode/text/UCA/CollationProperties.java#L197-L209), concatenating individually equivalent representatives does not necessarily produce an equivalent string: normalization can reorder the representatives.

   With the current development data, U+018D LATIN SMALL LETTER TURNED DELTA folds to `U+1DE6 U+1ABF` under `UCA_Fold_3_Shifted`, `UCA_Fold_4_Shifted`, and `UCA_Fold_3_Non_Ignorable`.

   Those combining characters have canonical combining classes 230 and 220, so normalization reverses them:

   | Input                                  | Primary weights |
   | -------------------------------------- | --------------- |
   | Original U+018D                        | `27C3 2793`     |
   | Result of folding, after normalization | `2793 27C3`     |

   This violates a fundamental requirement of collation folding.

   **Suggested fix:** verify the completed candidate against the source’s normalized sort key at the applicable strength. If concatenation changes the key, retain an equivalent representative of the whole expansion. Merely normalizing the candidate does not fix the problem.

2. **Algorithmic Hangul syllables are missing from the equivalence classes.**

   [The folding builder](https://github.com/unicode-org/unicodetools/blob/6b8d5dbfa3be5ec2556194b881aa58f746af4b5c/unicodetools/src/main/java/org/unicode/text/UCA/CollationProperties.java#L119-L129) enumerates explicit table entries, leaving algorithmically handled Hangul syllables with identity mappings.

   For example, with `UCA_Fold_1_Non_Ignorable`:

   | Character                        | Folded value |
   | -------------------------------- | ------------ |
   | 가 U+AC00 HANGUL SYLLABLE GA      | `AC00`       |
   | ㉮ U+326E CIRCLED HANGUL KIYEOK A | `1100 1161`  |

   These characters have equal primary collation weights. Nevertheless:

   ```java
   property.getSet("\u1100\u1161")
   ```

   includes U+326E and excludes U+AC00. This breaks the equivalence queries that the new properties are intended to support. Canonical equivalence between the two returned strings does not help because property-value matching compares the strings.

   I reproduced this at primary and secondary strengths, with both shifted and non-ignorable handling.

   **Suggested fix:** include algorithmically decomposed Hangul in the folding computation and add tests that equivalent inputs receive identical folded values.

3. **The advertised historical support produces additional incorrect folds.**

   The [new historical collator construction](https://github.com/unicode-org/unicodetools/blob/6b8d5dbfa3be5ec2556194b881aa58f746af4b5c/unicodetools/src/main/java/org/unicode/text/UCA/UCA.java#L1758-L1786) loads the old tables, but the folding calculation does not consistently reconcile their raw collation elements with normalization and implicit weights.

   Two examples using version 2.1.9:

   * U+3285 CIRCLED IDEOGRAPH SIX folds to U+F9D1 at primary strength. The source has primary weight `516D`; the target, after canonical decomposition and implicit-weight processing, has `FB40 D16D`.
   * U+01C4 LATIN CAPITAL LETTER DZ WITH CARON folds to `dž` at secondary strength. The source’s secondary weights are `0020 0041`; the target’s are `0020 0020 0041`.

   These differences occur using the **same versioned collator** for the source and result; they do not depend on comparing old results with modern DUCET.

   Checking every explicit folding mapping for 2.1.9 found five failures in each primary variant and fifty in each secondary variant.

   **Suggested fix:** resolve the historical normalization and implicit-weight inconsistencies, or narrow the supported versions until they are resolved. Tests using only 15.0 and 15.1 do not establish correctness back to 2.1.9.

4. **The new default leaks as the literal string `<next code point>`.**

   [`resolveValue()` handles the new default](https://github.com/unicode-org/unicodetools/blob/6b8d5dbfa3be5ec2556194b881aa58f746af4b5c/unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java#L873-L880), but [`_getUnicodeMap()` does not](https://github.com/unicode-org/unicodetools/blob/6b8d5dbfa3be5ec2556194b881aa58f746af4b5c/unicodetools/src/main/java/org/unicode/props/IndexUnicodeProperties.java#L772-L789). Consequently, for `UCA_Next_Non_Ignorable`:

   ```java
   property.getValue(0x100000)                 // U+100001
   property.getUnicodeMap().get(0x100000)      // "<next code point>"
   property.getAvailableValues()
           .contains("<next code point>")    // true
   ```

   I reproduced this with both normal and incremental property loading. The `getAvailableValues()` problem also affects callers that avoid the UnicodeMap API entirely.

   **Suggested fix:** support the new default consistently in value enumeration and resolved-map access, as well as individual lookups.

5. **The mixed-tertiary validation cannot detect anything.**

   [The loop](https://github.com/unicode-org/unicodetools/blob/6b8d5dbfa3be5ec2556194b881aa58f746af4b5c/unicodetools/src/main/java/org/unicode/text/UCA/CollationProperties.java#L276-L281) repeatedly checks `collationElements.at(0)` against the value previously obtained from that same element. Its exception is unreachable.

   However, changing `at(0)` to `at(i)` would introduce failures for valid data. For example, U+00C0 has tertiary weights `0008` and `0002`; implicit collation elements also have continuation weights that differ from the first element.

   **Suggested fix:** clarify whether the property means “the first collation element’s tertiary weight.” If so, document that and remove the misleading validation. Otherwise, define which elements should participate before implementing the check.

6. **The folding-value regexes reject legitimate expansions.**

   [All seven folding properties declare `$codePoint`](https://github.com/unicode-org/unicodetools/blob/6b8d5dbfa3be5ec2556194b881aa58f746af4b5c/unicodetools/src/main/resources/org/unicode/props/IndexPropertyRegex.txt#L128-L134), although their values can contain multiple code points. For example, the PR’s own ℅ → `co` example produces `0063 006F`, which fails the declared regex.

   These should use `$codePoints`. Computed properties bypass the ordinary parsing validation, which explains why this mismatch does not fail CI.

Two minor cleanups: `stringToElementsByType` is populated but never read, and `TestNextCodePoint()` repeats the same Unicode 18.0 shifted-FFFE assertion.

The most valuable additional tests would check two general properties across all seven folding modes: **folding preserves the normalized sort key**, and **collation-equivalent code points receive identical folds**. Include algorithmic Hangul and the earliest supported historical data.

eggrobin:
4 is WAI, see https://github.com/unicode-org/unicodetools/issues/716. 1 is out of scope, this is not an implementation of collation folding of strings.

eggrobin:
> 4 is WAI

Ah no, this is _getUnicodeMap, not _getRawUnicodeMap. (I’d like to get rid of _getUnicodeMap, but in the meantime, I need to deal with it…)

PR #1527, opened 2026-09-11
Arabic Astronomical Zero

charlotte-buff:
[[188-C11](https://www.unicode.org/cgi-bin/GetL2Ref.pl?188-C11)] Consensus: Provisionally assign 1 code point U+1EEE0 ARABIC ASTRONOMICAL ZERO in the Arabic Mathematical Alphabetic Symbols block, as described in [L2/26-172](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/26-172). [Ref: 2.7 in [L2/26-158](https://www.unicode.org/cgi-bin/GetMatchingDocs.pl?L2/26-158)]

* unicode-org/sew#880


roozbehp:
I agree that gc=Nl is a better general category for this than the gc=Lo we had proposed.

PR #1528, opened 2026-09-11
Remove a lot of unused code from GenerateConfusables.java

roozbehp:
This is divided into individual commits to make it more understandable.


eggrobin (approved):
A change that says +50 -532 looks good (I assume you checked that regenerating the security files has no effect after this change? I think they run in CI only as a smoke test.)

roozbehp:
> (I assume you checked that regenerating the security files has no effect after this change? I think they run in CI only as a smoke test.)

Yes. Absolutely no change to the security files beyond their dates.

PR #1530, opened 2026-09-14
Let USE_BETA drive choice of tr51 link

nedley:
The tr51 link was still to proposed.html due to an unnoticed `USE_PROPOSED` flag. And another flag, `Emoji.BETA_IS_OPEN`, could cause a beta header to be shown. Since each flag effectively opens a new configuration space that we don’t need, remove them both.


unicodetools/data/emoji/dev/internal/emoji-ordering-rules.txt
```diff
@@ -1,4 +1,4 @@
-<!-- Machine-readable version of the emoji ordering rules for v18.0β (corresponding to CLDR). -->
+<!-- Machine-readable version of the emoji ordering rules for v18.0 (corresponding to CLDR). -->
```
nedley, line 1: Opportunistic fix, file is not part of the release anyway

unicodetools/src/main/java/org/unicode/tools/emoji/Emoji.java
```diff
@@ -1034,7 +1023,7 @@ public static String show(String key) {
 
     public static final String TR51_HTML_BETA = "../../reports/tr51/proposed.html";
     public static final String TR51_HTML =
-            IS_BETA || USE_PROPOSED ? TR51_HTML_BETA : "https://unicode.org/reports/tr51/tr51.html";
+            IS_BETA ? TR51_HTML_BETA : "https://www.unicode.org/reports/tr51/#emoji_data";
```
nedley: Whoops! I think this has the fragment appended elsewhere, let me check…

roozbehp:
One minor cleanup: in ChartUtilities.getPointToOther(), the inner Emoji.IS_BETA conditional is now redundant. Whenever execution reaches it, IS_BETA must be true, so the "new beta version" branch is unreachable. Removing that branch would make the new behavior clearer.

roozbehp (approved):
LGTM

nedley:
> One minor cleanup: in ChartUtilities.getPointToOther(), the inner Emoji.IS_BETA conditional is now redundant. Whenever execution reaches it, IS_BETA must be true, so the "new beta version" branch is unreachable. Removing that branch would make the new behavior clearer.

ChartUtilities.getPointToOther() is also used in GenerateEmoji.java.

roozbehp:
> ChartUtilities.getPointToOther() is also used in GenerateEmoji.java.

Yes. My suggestion is only about the inner conditional inside that method:
* When IS_BETA is false, the outer conditional returns "".
* Otherwise, IS_BETA must be true, so the inner conditional always selects "current released version".

Consequently, the inner "new beta version" branch is unreachable regardless of which caller invokes the method. Removing that branch preserves the behavior for both callers.

nedley:
> Yes. My suggestion is only about the inner conditional inside that method:

I understand. My point was that changing the inner conditional would require a corresponding change in another file and poses no problem as the code stands.

nedley:
Oh, you mean the _inner_ inner conditional! I thought you were referring to the (hidden) conditional due to `skipVersion` in `writeHeader`. I won’t be tackling this right now.

roozbehp:
@nedley, would you like me to submit a pull request myself for the cleanup I suggested?

nedley:
> @nedley, would you like me to submit a pull request myself for the cleanup I suggested?

By all means, thanks!

roozbehp:
> > @nedley, would you like me to submit a pull request myself for the cleanup I suggested?
> 
> By all means, thanks!

https://github.com/unicode-org/unicodetools/pull/1531

PR #1532, opened 2026-09-15
Check input size in bidic.jsp

roozbehp:
Use codePointCount(), rather than length(), because the C array receives one entry per code point, while Java length() counts UTF-16 code units. The return prevents the native process from being launched after the error response.


roozbehp:
Please take a look. For the record, I have not run the code (I don't know how yet), I just proved it correct.

UnicodeJsps/src/main/webapp/bidic.jsp
```diff
@@ -133,6 +133,10 @@ function setUbaInput(str) {
     String ubaDetail = utfParams.getParameter("d", "2");
     boolean ubaShowVacuous = !"off".equals(utfParams.getParameter("y", "off"));
     String valInputCharSeq = utfParams.getParameter("s", "\u0645\u0627\u0631\u0652\u0643 \u2066\u0031\u2013\u0033%\u2069 mark (\u0366v.2)\u0368!");
+    if (valInputCharSeq.codePointCount(0, valInputCharSeq.length()) > 200) {
```
eggrobin, line 136: I think you could use valInputCharSeq.codePoints().count(), which seems nicer (but maybe @markusicu knows of Java reasons why this would be preferable?).

markusicu: .codePointCount() should be slightly faster because it creates no new objects, and because it need not/does not assemble code point integers; lower-level construct that only does what we want

eggrobin (approved):

PR #1538, opened 2026-09-15
Limit history in character.jsp to ASCII alphanumerics

roozbehp:
This rejects every character that could break out of the URL, JavaScript
string, or HTML attribute and automatically permits future alphanumeric
values without changing the JSP. It also protects the downstream links
generated by UnicodeJsp.showProperties, because only safe characters are
passed into Java.


UnicodeJsps/src/main/webapp/character.jsp
```diff
@@ -24,6 +24,9 @@ th           { text-align: left }
         UtfParameters utfParameters = new UtfParameters(queryString);
         String text = utfParameters.getParameter("a", "\u2615", "\u2615");
         String history = utfParameters.getParameter("history", "", "");
+        if (!history.equals("assigned") && !history.equals("full")) {
```
eggrobin: I would prefer escaping it below, or maybe restricting it to not contain ASCII non-alphanumerics; it isn’t implausible that I might want to add more values, and having to update it here as well as in the Java seems annoying.

roozbehp: Fixed. PTAL.

eggrobin (approved):

PR #1546, opened 2026-09-19
First pass at MathClass for Unicode 18

jowilco:
First pass at MathClass for Unicode 18


roozbehp (changes requested):
1. U+1DB1C FACIT SYMBOL should be R (it currently has N). The examples in [WG2/N5334R2](https://www.unicode.org/wg2/docs/n5334R2-L-2603.Proposal.to.encode.historical.mathematical.relations.pdf#page=53), pages 53–54, put it between a calculation and its result, functioning as an equality/result relation. That supports R.

2. Line 18 of addmathclass.xsl says `<mathchars unicode="18.0" tr25="16">`. It should say `tr25="17"`.

3. The HTML row for U+1CEDE says `<t rclass="changed">`. It should be `<tr class="changed">`.

4. The expression `(@gc|../@gc)` is producing two general categories in comments such as `<!--So Sm SQUARE ROOT OF SQUARE ROOT-->`. I think you should use `(@gc, ../@gc)[1]` instead.

5. U+1CEDD..U+1CEDF should be L. The existing square, cube, and fourth-root signs are L (as well as the Arabic-Indic cube and fourth roots at U+0606 and U+0607).

6. I think U+1D6A6 should be A. Assigning N to it would make it the only non-digit character with N in that block.

eggrobin:
> U+1DB1C FACIT SYMBOL should be R. The examples in [WG2/N5334R2](https://www.unicode.org/wg2/docs/n5334R2-L-2603.Proposal.to.encode.historical.mathematical.relations.pdf#page=53), pages 53–54, put it between a calculation and its result

Clearly semantically a relation. Of course a letter can get drafted as a relation, look no further than https://en.wikipedia.org/wiki/Binary_relation#Definition. But the spacing in these modern typesettings seems to treat it _typographically_ like a relation, so I agree with R.

> U+1CEDD..U+1CEDF should be L.

I am not convinced. All the evidence I see in [WG2 N5333R2 p. 18](https://www.unicode.org/wg2/docs/n5333R2-L-2533%20Proposal%20to%20encode%2012%20cossic%20characters.pdf#page=18) is together with √ as `\surd`, not as `\sqrt`.

In particular, with U+1CEDD ③ for cube root of cube root, the layout behaviour of √ as `\sqrt` would be counterproductive.

> I think U+1D6A6 should be A.

I agree.

jowilco:
@roozbehp - Just making sure that you saw Robin's comment:

>> U+1CEDD..U+1CEDF should be L.
>
>I am not convinced. All the evidence I see in [WG2 N5333R2 p. 18](https://www.unicode.org/wg2/docs/n5333R2-L-2533%20Proposal%20to%20encode%2012%20cossic%20characters.pdf#page=18) is together with √ as `\surd`, not as `\sqrt`.
>
>In particular, with U+1CEDD ③ for cube root of cube root, the layout behaviour of √ as `\sqrt` would be counterproductive.

I'll implement the changes once there is agreement.

Issue #453, opened 2023-04-28
invariant test: no new bpb with decomp


markusicu:
From discussion during UTC-175 of PAG report (L2/23-079) item [4.1 Dependency of the bidi algorithm on normalization](https://docs.google.com/document/d/1vgIrigEd5X9cEdJwiyBweklr1tVnEJQJN5fx14ZKH7M/edit#heading=h.gsvtuu2qt7ud):

We recorded
- [175-C] Consensus: The UTC will not add further characters to the Unicode Standard with both (a) canonical decompositions, and (b) Bidi_Paired_Bracket_Type ≠ None. Unicode will also not add further characters where the canonical decompositions *contain* characters whose Bidi_Paired_Bracket_Type ≠ None. See L2/23-079 item 4.1.
- [175-C] Consensus: The UTC establishes the above Consensus 175-C?? as a precedent according to TC procedures 10.5.2.

We should add an invariant test for this.
The test comment should note that this is a *precedent*.

@Manishearth @eggrobin @macchiati @roozbehp

roozbehp:
Pull request at https://github.com/unicode-org/unicodetools/pull/1547

PR #1547, opened 2026-09-20
Add invariant for UTC-175-C21

roozbehp:
See https://github.com/unicode-org/unicodetools/issues/453


unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
```diff
@@ -343,6 +343,19 @@ In \p{dt=canonical}, bc * (take 1) * dm  = bc
...
 
+# Precedent (UTC-175-C21; L2/23-079, item 4.1): The UTC will not add further
+# characters to the Unicode Standard with both (a) canonical decompositions,
+# and (b) Bidi_Paired_Bracket_Type ≠ None. Unicode will also not add further
+# characters where the canonical decompositions *contain* characters whose
+# Bidi_Paired_Bracket_Type ≠ None.
+# https://github.com/unicode-org/unicodetools/issues/453
+# The existing exceptions to both restrictions are U+2329 and U+232A.
+Let $decomposableBidiBrackets := [\u2329\u232A]
+[\p{dt=Canonical} & \P{Bidi_Paired_Bracket_Type=None}] ⊆ $decomposableBidiBrackets
```
eggrobin:
```suggestion
[\p{dt=Canonical} & \p{Bidi_Paired_Bracket_Type≠None}] ⊆ $decomposableBidiBrackets
```

unicodetools/src/main/resources/org/unicode/text/UCD/UnicodeInvariantTest.txt
```diff
@@ -343,6 +343,19 @@ In \p{dt=canonical}, bc * (take 1) * dm  = bc
...
+# Use full canonical decomposition so that indirect mappings are checked too.
```
eggrobin: I don’t think that’s a good idea; looking at every character means we would check the outcome of recursive decompositions as well, but this assertion does not see what happens in the middle.

Of course if something happens in the middle, it means that we have a new decomposable bidi bracket or that the decomposition of one of the existing bidi brackets changed, both of which will break other assertions in this file. But I’d rather check the statement more directly:
```
In [\p{dt=Canonical} - $decomposableBidiBrackets], (delete-adjacent-duplicates) * Bidi_Paired_Bracket_Type * Decomposition_Mapping = (constant None)
```

PR #1550, opened 2026-09-21
Delete unchanged generated files instead of renaming them

### LINKED ISSUE #457: Delete unchanged files instead of renaming to "ZZZ-UNCHANGED-..."
Per discussion in unicodetools meeting of 2 May 2023, to simplify merging in data-for-new PRs, identical files should be simply deleted rather than renamed `ZZZ-UNCHANGED-...`

See https://github.com/unicode-org/unicodetools/blob/b7ccbe65c3a1ec9aadf416fa17ad6dc5fdbf1a9d/unicodetools/src/main/java/org/unicode/text/utility/Utility.java#L1335-L1347

roozbehp:
Fixes https://github.com/unicode-org/unicodetools/issues/457


josh-hadley (approved):

eggrobin:
(I think Markus’s instructions are out of date here, since MakeUnicodeFiles does the copy by default since https://github.com/unicode-org/unicodetools/pull/1289; so this doesn’t simplify much for the user of the tools. But it’s probably better to not leave random ZZZ-garbage lying around in the Generated directory, so this is a good change.)

PR #1553, opened 2026-09-21
Faster Age generation

### LINKED ISSUE #1124: faster/simpler Age / DerivedAge.txt generation
We currently compute a character's Age by iterating over all Unicode versions from 1.1 to latest=dev and returning the first version where its code point is no longer unassigned.

This means that we have to load/parse data files for each Unicode version.

This should be unnecessary, because we already have a DerivedAge.txt file from at least the last version. We should be able to:
- parse the dev=latest file
- recent additions: take the set of `[[:age=NA:]&[:^gc=Cn:]]` and map them to the latest version
- recent removals: take the set of `[[:^age=NA:]&[:gc=Cn:]]` and map them to "NA"

We should move the old logic into a test, to verify that the simple logic yields the same Age property map as the brute force code.

Discussion: See this comment and replies to it:
https://github.com/unicode-org/unicodetools/pull/1116#issuecomment-2848036146

roozbehp:
Fixes https://github.com/unicode-org/unicodetools/issues/1124


roozbehp:
Reduces time to generate UCD by ~25%!

unicodetools/src/main/java/org/unicode/text/UCD/ToolUnicodePropertySource.java
```diff
@@ -2251,35 +2266,15 @@ protected String _getValue(String string) {
...
-                    // Private use and CJK compatibility ideographs, not overt in UnicodeData 1.1.5.
-                    return UCD_Names.LONG_AGE[i];
-                }
+            if (ageMap == null) {
+                ageMap =
+                        deriveAge(
+                                ucd.getVersionInfo(),
+                                ucd::isAllocated,
+                                IndexUnicodeProperties.make(ucd.getVersion())
+                                        .load(UcdProperty.Age));
```
eggrobin: I would like to avoid using the UnicodeMap API directly, see https://github.com/unicode-org/unicodetools/issues/716, but also because I think we may need to move away from a pure UnicodeMap to solve some of our performance issues. Since you are iterating over all characters and mutating a clone of the map, I don’t think you have much to gain from doing it here; use getProperty instead.

roozbehp: You mean like this? (see the second commit)

eggrobin (commented):
Yes, like this. And I have checked that the performance at this second commit is indistinguishable from at the first commit, and that going through the motions of adding a new character to UnicodeData.txt and turning the crank still gives it the right value in DerivedAge, so this looks good to me.

unicodetools/src/test/java/org/unicode/text/UCD/TestAge.java
```diff
@@ -0,0 +1,79 @@
...
+                ToolUnicodePropertySource.deriveAge(
+                        currentVersion,
+                        cp -> cp == added,
+                        new UnicodeProperty.UnicodeMapProperty().set(input));
+
+        assertEquals(currentAge, result.get(added));
+        assertEquals("Unassigned", result.get(removed));
+        // The input Age map must remain unchanged.
+        assertEquals("Unassigned", input.get(added));
+        assertEquals(currentAge, input.get(removed));
```
eggrobin, line 59: This seems like an odd thing to test (if a user of the UnicodeProperty API can use that to change the map that a specific subclass uses as a representation, something has gone horribly wrong; and if it can do that to a _frozen_ map, something has gone even more horribly wrong), but it probably makes the test clearer (I write expectations that serve more as notes than tests of requirements quite often).

(Excitingly, it looks like I managed to transmute a multiline comment into a single-line comment, making the suggestion here nonsensical, so the suggestion is in a separate comment below instead.)

eggrobin:
```suggestion
        // The input Age map remains unchanged.
        assertEquals("Unassigned", input.get(added));
        assertEquals(currentAge, input.get(removed));
```

roozbehp: Yes. The test started as something else, it made more sense in that earlier iteration. I was tempted to just drop the whole `testAdditionsAndRemovals()`, but thought it won't hurt since I already had it. If you think it's a waste of cycles, I can just delete it.

eggrobin: I think it’s fine so long as the comment makes sense.

eggrobin (approved):

roozbehp:
Thank you for the review! Should I land this or wait for https://github.com/unicode-org/unicodetools/pull/1555 to land first?

unicodetools/src/test/java/org/unicode/text/UCD/TestAge.java
eggrobin: > I was tempted to just drop the whole `testAdditionsAndRemovals()`

By the way, while the last two expectations are more documentation of the test than testing, this test is very important: it is the only thing that checks the « DerivedAge inconsistent with UnicodeData » paths, which we use when adding new characters (or removing them prior to release). As I mentioned, I also checked that by hand to make sure there was nothing unexpected, but I’d rather regressions be caught in CI by a unit test than by Charlotte or I when doing propertizing.

eggrobin:
No functional change, so no need to wait for #1555.

unicodetools/src/test/java/org/unicode/text/UCD/TestAge.java
roozbehp: > which we use when adding new characters (or removing them prior to release).

Did you notice the easter egg then?

eggrobin: I had not looked at the code points… 😭

eggrobin:
(I should avoid typoing out negations. The above comment, now corrected, was meant to say _no_ need to wait, go ahead and merge.)

PR #1555, opened 2026-09-21
Unicode 18.0→ 19.0

eggrobin:
The same procedure as https://github.com/unicode-org/unicodetools/pull/1204<span></span>—the same procedure as every year.


roozbehp:
Pasting GPT-6 Astra review:

The core rollover looks correct, but I found one functional omission, some missed documentation updates, and a bidi follow-up.

1. `LAST2_VERSION_INFO` needs to become 17.0.
Settings.java, line 53 still says `VersionInfo.getInstance(15, 1)`.
This affects behavior: `VersionedSymbolTable.forReview()` uses that constant for the `U-1:` qualifier. I verified that, after this PR, review-mode `U-1:` resolves to 15.1, whereas the previous version relative to the newly released 18.0 is 17.0. Development-mode `U-1:` correctly resolves to 18.0.
Change it to `VersionInfo.getInstance(17, 0)`. This value was already stale on `main`, but the rollover should update it. A regression test checking the relationship between the version constants would prevent this recurring.

2. Two generation documents still point to 18.0 output directories. These need the same treatment as the updated `docs/pipeline.md`:
    * `docs/idna.md`: the comparison and copy commands for both IDNA files.
    * `docs/ucdxml.md`: the listing, cleanup, comparison, and directory-change commands.

    The generators now produce 19.0 output. Following these instructions would either fail or operate on leftover 18.0 files.

3. The C bidi web tool still needs its 18.0 update—but there is an external prerequisite.
UBAVersion.java still ends at `V17_0`, and the Dockerfile still downloads the 17.0 implementation.
These should eventually advance together. However, the public 18.0 source and include directories currently contain no files. Changing the Docker version now would therefore break its build. I would track this as separate release follow-up work.

The main changes check out:

* All 235 archived files, across all seven data families, exactly match their corresponding `dev` files at the PR’s base commit.

* Targeted Java checks passed for the age tables, version lookup, 19.0 → `dev` routing, 18.0 → archived-data routing, and development-version qualifiers.

* Removing U+1CF5 from the Devanagari test expectation is correct: its Unicode 18.0 `Script_Extensions` value is `Beng`.

* The reported CI checks pass. I ran targeted checks locally, rather than rerunning the full Maven suite.

eggrobin:
> LAST2_VERSION_INFO needs to become 17.0. Settings.java, line 53 still says `VersionInfo.getInstance(15, 1)`.

Looks like I missed that one last time, too.

> A regression test checking the relationship between the version constants would prevent this recurring.

Or I could just move the things that need synchronous updating so they are together in the file.

> docs/idna.md

That will get updated in 2027 when Markus deals with that before α: https://github.com/unicode-org/unicodetools/commits/main/docs/idna.md

> docs/ucdxml.md

Same, but before β: https://github.com/unicode-org/unicodetools/commits/main/docs/ucdxml.md.

> bidirefc

I know, Ken knows, the BRS knows.

roozbehp (approved):
LGTM

unicodetools/src/main/java/org/unicode/text/utility/Settings.java
```diff
@@ -22,7 +22,7 @@ public class Settings {
     // Are they equivalent for our purposes?
 
     /** Used for the default version. */
-    public static final String latestVersion = "18.0.0";
+    public static final String latestVersion = "19.0.0";
```
markusicu: optional: why not add a FIX_FOR_NEW_VERSION comment to the relevant Settings fields?

unicodetools/src/main/resources/org/unicode/text/UCD/MakeUnicodeFiles.txt
```diff
@@ -231,6 +231,9 @@ Value:  V17_0
 Value:  V18_0
 # Newly assigned in Unicode 18.0.0 (September, 2026)
 
+Value:  V19_0
+# Newly assigned in Unicode 19.0.0 (September, 2026)
```
markusicu: optional: is there a way in this file's syntax to add an actual FIX_FOR_NEW_VERSION comment? (one that isn't copied into the output file)

pub/copy-alpha-to-draft.sh
```diff
@@ -12,8 +12,8 @@ UNITOOLS_DATA=$UNICODETOOLS/unicodetools/data
 
 # Adjust the following for each year and version as needed.
```
markusicu: optional: why not add a FIX_FOR_NEW_VERSION comment into each pub script?

unicodetools/src/main/resources/org/unicode/text/UCD/MakeUnicodeFiles.txt
```diff
@@ -232,7 +232,7 @@ Value:  V18_0
 # Newly assigned in Unicode 18.0.0 (September, 2026)
 
 Value:  V19_0
-# Newly assigned in Unicode 19.0.0 (September, 2026)
+# Newly assigned in Unicode 19.0.0 (September, 2027)
```
markusicu: optional: squash the 2027 update into the 19.0.0 update?

unicodetools/src/main/java/org/unicode/props/VersionToAge.java
```diff
@@ -41,6 +41,7 @@ public enum VersionToAge {
 
     ucd(
             ImmutableMap.<VersionInfo, Long>builder()
+                    .put(VersionInfo.getInstance(19, 0), getDate(2027, 9))
```
markusicu: optional: FIX_FOR_NEW_VERSION comments?

unicodetools/src/main/java/org/unicode/text/UCD/UCD_Names.java
```diff
@@ -663,6 +663,7 @@ public final class UCD_Names implements UCD_Types {
         "V16_0",
         "V17_0",
         "V18_0",
+        "V19_0",
```
markusicu: optional: squash into earlier UCD_Names commit?

unicodetools/data/emoji/18.0/internal/emoji-diff.txt
markusicu: optional: squash the "I am bad at copying" commit into the earlier copy?

unicodetools/data/emoji/dev/internal/emoji-proposals.txt
markusicu: optional: squash the GenerateEmojiData commits together?

UnicodeJsps/src/test/java/org/unicode/jsptest/TestMultivalued.java
```diff
@@ -23,7 +23,7 @@ public void TestScx1Script() {
         String unicodeSetString = "\\p{scx=deva}";
         UnicodeSet parsed = UnicodeSetUtilities.parseUnicodeSet(unicodeSetString);
 
-        UnicodeSet mustContain = new UnicodeSet("[ᳵ।]"); // one character B&D, other B&D&D&G&...
+        UnicodeSet mustContain = new UnicodeSet("[।]"); // B&D&D&G&...
```
markusicu: why did this pass before the version upgrade?

markusicu (approved):
optional reminders & commit cleanups...

UnicodeJsps/src/test/java/org/unicode/jsptest/TestMultivalued.java
eggrobin: We are in the JSPs, so this gets parsed with the VersionedSymbolTable.forReview, which has the latest released version as its implicit version:
https://github.com/unicode-org/unicodetools/blob/6215dee83eb27787c91c90a9dd79162c01347684/UnicodeJsps/src/main/java/org/unicode/jsp/UnicodeSetUtilities.java#L113

So this was using 17.0 data. We could change these tests to use `Udev:` so they break when we change the data and not when we do the rollover, but this is a pretty rare occurrence.

unicodetools/src/main/java/org/unicode/text/utility/Settings.java
eggrobin: Oops, I missed that there were comments other than the last. Might as well, in a subsequent PR.

unicodetools/src/main/resources/org/unicode/text/UCD/MakeUnicodeFiles.txt
eggrobin: Not sure, I’ll look.

pub/copy-alpha-to-draft.sh
eggrobin: Good question, I’ll add those…

unicodetools/src/main/resources/org/unicode/text/UCD/MakeUnicodeFiles.txt
eggrobin: squash-merged anyway.

unicodetools/src/main/java/org/unicode/text/UCD/UCD_Names.java
eggrobin: all squish-merged.

unicodetools/data/emoji/dev/internal/emoji-proposals.txt
eggrobin: https://youtu.be/OAGfx7eB3gA?t=155

PR #1557, opened 2026-09-22
Make GenerateUnihanCollators not use ICU Unicode properties

### LINKED ISSUE #1336: GenerateUnihanCollators should not use ICU Unicode properties
GenerateUnihanCollators relies on ICU Unicode properties. This means that when we generate data for an upcoming Unicode version, and especially when the Unicode Tools --> ICU dependency is outdated, we get properties data for an old version of Unicode.

Examples:
```java
    private static final UnicodeSet NOT_NFC = new UnicodeSet("[:nfc_qc=no:]").freeze();
    private static final UnicodeSet NOT_NFD = new UnicodeSet("[:nfd_qc=no:]").freeze();
    private static final UnicodeSet NOT_NFKD = new UnicodeSet("[:nfkd_qc=no:]").freeze();
    private static final UnicodeSet UNIHAN_LATEST =
            new UnicodeSet("[[:ideographic:][:script=han:]]").removeAll(NOT_NFC).freeze();
```

We should use Unicode Tools APIs to get properties.

@macchiati FYI

roozbehp:
Fixes https://github.com/unicode-org/unicodetools/issues/1336


eggrobin:
It might be more readable to still use UnicodeSet syntax, but use the constructor with a SymbolTable, like here:
https://github.com/unicode-org/unicodetools/blob/6215dee83eb27787c91c90a9dd79162c01347684/unicodetools/src/main/java/org/unicode/text/UCD/GenerateBreakTest.java#L384-L387


I would like to find a way to
https://github.com/unicode-org/unicodetools/blob/6215dee83eb27787c91c90a9dd79162c01347684/unicodetools/src/main/java/org/unicode/text/UCD/GenerateConfusables.java#L127

systematically so the ordinary constructor does the right thing, but for static initialization this seems extremely brittle.

macchiati:
I agree with "It might be more readable to still use UnicodeSet syntax, but
use the constructor with a SymbolTable, like here:"

roozbehp:
> It might be more readable to still use UnicodeSet syntax, but use the constructor with a SymbolTable, [...]

Done. PTAL.

